PR #2, opened 2016-03-14
New ACSREJ algorithm and bug fixes

pllim:
**You can merge this once approval is obtained from ACS Team.**

See:
https://trac.stsci.edu/ssb/stsci_python/ticket/598
https://trac.stsci.edu/ssb/stsci_python/ticket/976
https://trac.stsci.edu/ssb/stsci_python/ticket/1223
https://trac.stsci.edu/ssb/stsci_python/ticket/1224

**Instructions to install this branch for testing**

1. Create a directory to store the source codes and `cd` into it. (If you have done it before for another PR, just `cd` into it your existing directory.)

2. `git clone https://github.com/pllim/hstcal.git ./` (this makes a copy of my HSTCAL fork on your disk, skip this if you already cloned it before for another PR)

3. `git fetch origin lim-calacs-ticket976`

4. `git checkout origin/lim-calacs-ticket976 -b lim-calacs-ticket976` (if this does not work, your `git` version is too old, so you will need to update `git` first or install a newer version using `conda install git`)

5. `git status` (if this shows that you are now in `lim-calacs-ticket976` branch, then proceed)

6. `./waf configure --debug --prefix=$CONDA_PREFIX` (this only works if you use `conda env`)

7. `./waf build` (make sure there is no error)

8. `./waf install`

9. Now `cd` into your working directory with data files.

10. ~~`calacs.e --version` (if it is not "9.1.1 (03-Apr-2017)", you installed the wrong version)~~
     `calacs.e --gitinfo` check the commit against that of this branch as a sanity check. 

11. `calacs.e your_file_name -v` (`-v` gives verbose output in trailer file and screen, which is useful for debugging)

**Instructions to grab updates from existing branch checkout**

1. `cd` into source checkout directory.

2. `git checkout lim-calacs-ticket976` to switch to this branch, if not already on it (you can check with `git status`)

3. `git fetch origin lim-calacs-ticket976` (picks up updates)

4. `git rebase origin/lim-calacs-ticket976` (applies the updates locally)

5. Continue from Step 8 above.

### DIFF
--- a/pkg/acs/calacs/Dates
+++ b/pkg/acs/calacs/Dates
@@ -1,3 +1,6 @@
+ TBD        :  CALACS 9.1.1 ACSREJ calculations are now done entirely in
+                            electrons. Threshold formula now uses ERR arrays to
+                            account for post-flash noise, if any. And bug fixes.
  01-Jun-2017:  CALACS 9.2.0 New CTE algorithm implemented.
  21-Feb-2017:  CALACS 9.1.0 New SINKCORR step added in ACSCCD for WFC.
  22-Nov-2016:  CALACS 9.0.0 BLEVCORR now uses new OSCNTAB that correctly
--- a/pkg/acs/calacs/History
+++ b/pkg/acs/calacs/History
@@ -1,4 +1,5 @@
-### 01-Jun-2017 - JAN -- Version 9.2.0
+### TBD - PLL  -- Version 9.1.1
+    ACSREJ calculations now in electrons. Threshold uses ERR arrays. Bug fixes.### 01-Jun-2017 - JAN -- Version 9.2.0
     New CTE algorithm implemented. Use of either new (gen2) & old (gen1) 
     dynamically detected from CTE_NAME in PCTETAB, however --ctegen 1 is explicitly
     required to override the gen2 default.
--- a/pkg/acs/calacs/Updates
+++ b/pkg/acs/calacs/Updates
@@ -29,6 +29,9 @@ Update for version 9.2.0 - 01-Jun-2017 (JAN) : git diff f0c177019cb4c159c801e022
     pkg/acs/calacs/lib/wscript
     wscript
 
+Update for version 9.1.1 - TBD (PLL)
+    acsrej/*
+
 Update for version 9.1.0 - 21-Feb-17 (PLL)
     acsccd/doccd.c
     acsccd/dosink.c
--- a/pkg/acs/calacs/acsrej/acsrej_init.c
+++ b/pkg/acs/calacs/acsrej/acsrej_init.c
@@ -11,12 +11,12 @@
 
 # define    OK          (short)0
 
-/*  acsrej_init -- get the initial average pixel values
+
+/* acsrej_init -- get the initial average pixel values
 
   Description:
   ------------
   Get the initial average according to the specified scheme
-  This function
 
   Date          Author          Description
   ----          ------          -----------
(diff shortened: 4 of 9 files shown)


jamienoss:
@pllim What's the status for this? Is it a 2017.2a (hstcal-1.3.0) or 2017.3 (hstcal-1.4.0) addition (if at all)?

pllim:
No, just leave it alone and open for now. I'll provide update when available.

pkg/acs/calacs/acsrej/acsrej_init.c
```diff
@@ -196,25 +185,26 @@ int acsrej_init (IODescPtr ipsci[], IODescPtr ipdq[], clpar *par, int nimgs,
...
-                       raw0 is DN/s - is /s supposed to be there?
+                  https://trac.stsci.edu/ssb/stsci_python/ticket/1223#comment:13
+                       ERR = (READNSE^2 + VAL + (SCALE * VAL)^2) / EXPTIME^2
+                       nse is e^2
+                       raw0 is e/s - is /s supposed to be there?
                        exp2n is s^2
                     */
                     Pix(sg->err.data, i, j) =
-                        (nse[0] + raw0 / gn[0] + SQ(scale * raw0)) / exp2n;
+                        (nse[0] + raw0 + SQ(scale * raw0)) / exp2n;
```
mdlpstsci: From dimensional analysis alone, I do not see how the computation of Pix(sg->err.data,i,j) can be correct here, line 245, and line 314.  I will wait to see if this is clarified in later updates.

pllim: Hmm, GitHub is saying that you commented on outdated code. Were you looking at an intermediate commit that has since been superseded by a subsequent one? That is, is this change still present in the final commit of the PR?

mdlpstsci:
Hi,

I am looking at the intermediate changes, so this well may be cleared up later.  I am also looking at the final version.  I debated about making any comments – guess I should have waited until the final version.  Even if this code is still present in the final, this is not something you did.  It does not look like the very original code is correct to me.  I admit I have been looking at so much code today that I am weary and probably should not comment at all.  Please ignore the comments on the intermediate code, and I will refrain from making any comments until the final.

Best,
M

Michele D. De La Pena
Space Telescope Science Institute
mdelapena@stsci.edu
667.218.6501

pllim:
@mdlpstsci , no worries. I don't quite recall now but I don't think the commits were separated in such a way that the code would work properly in between. So I wouldn't recommend looking at each commit and think too hard about it; instead look at the combined diff when in doubt. Hope this helps.

jamienoss:
@mdlpstsci since you've discovered that the individual commits are not all linearly independent, I would just squash the whole thing. This has to be done anyhow before we merge. As @pllim sated, you can just review the code as if it were squashed already. Just click on the "Files Changed" tab instead of the "Commits" tab.

jamienoss:
@pllim was there a reason for these commits or some of them not to be squashed? Was their intention to aid review or...? Do you feel that the current instance adequately reflects this intent?

jamienoss:
@pllim 

> I would prefer not to squash if possible.

If you don't want them to be squashed they will have to be adequately self-contained, which some see not to be. You have changes in later commits that really belong to former ones. As they stand, not all 12 will be merged. What I'm asking is if you can clearly indicate to which ones you feel confident are self-contained. Otherwise, they'll all be squashed.

> Having them separate might come in handy to track down bugs in the future.

If they were self-contained, this would be true. Since they are not, they can actually do the opposite and hinder bug hunting by leading you down the wrong hole.

> 12 commits aren't that bad.

Any commit that isn't directly and independently associated with an issue is bad. 12 of them is 12xbad.

pllim:
I'll defer to @mdlpstsci as I am not the one who has to fix the bugs.

jamienoss:
@pllim 
> I'll defer to @mdlpstsci as I am not the one who has to fix the bugs.

Does this mean that you are unable to help out and indicate which commits can stay?
Are you also stating that you are not in a position to rectify the commits such that they are self-contained?

pllim:
Yes. I don't clearly remember what I did back then. Either squash them all into one or leave them alone, that is my recommendation.

jamienoss:
> Also this needs another rebase. FYI.

Yeah, I'm going to hold off and wait this time until further reviewed.

pkg/acs/calacs/acsrej/acsrej_sky.c
```diff
@@ -88,52 +108,54 @@ void acsrej_sky (char *sky, IODescPtr ipsci[], IODescPtr ipdq[],
...
+       DELTA is difference between mean of data and MIN
+       MAX is mean plus DELTA
+       This insures that the mean falls in the center of the range
+       between min and max. */
     if (npt == 0) npt = 1;
     min = (data_min < MINVAL) ? MINVAL : data_min;
     mean = (sum > 0.) ? (int) ( (sum / (float)npt) + 1) : 1;
     max = 2 * mean - min;
-    
+
```
mdlpstsci: This comment is on "old" code which seems to pre-date Pey Lian's updates to the the ACSREJ algorithm.  While the comments in the code on finding the min, mean, and max values are true, this technique ignores values which are actually larger than the pseudo-max.  This is one way to eliminate outliers, but the mean itself has been computed based upon the outliers.  A kappa-sigma clipping is traditionally used to eliminate outliers and find a more robust mean.

mdlpstsci: Note to self: The determination of the mean is probably good enough as the histogram is being used as a filtering mechanism from which a mode value will be determined to characterize an entire input image.

pllim: Note on note to self: You should note down all these stuff in preparation to write an ACS Instrument Science Report on the new ACSREJ algorithm. :smile:

pkg/acs/calacs/acsrej/acsrej_init.c
```diff
@@ -27,266 +27,156 @@
...
                         val = 0.;
                     }
 
-                    if ( (n == 0) || (val < Pix(sg->sci.data,i,j)) ) {
+                    if ( (n == 0) || (val < Pix(sg->sci.data, i, j)) ) {
+                        /* If this pixel is bad in the first image,
+                           then the min is automatically set to 0. As a
+                           result, only negative val is going to be stored and
+                           valid positive min is ignored.
+                           SLIGHTLY BUGGY HERE??? */
```
mdlpstsci: Note to self: Agreed this algorithm is buggy, and it pre-dates any changes made by PLL.   Algorithm requires image with exposure time > 0 AND good pixel value in order to set a viable minimum.  Re-think algorithm.  Determine default method: median or minimum.

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -20,9 +20,10 @@
 # define    ABS(x)      ((x>0.) ? x : -x)
```
mdlpstsci: Note to self: Why is ABS defined rather than use the C library function fabs().

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -332,13 +390,13 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr ipdq[],
 
     /* Assumption: ALL images have the same noise/gain values */
     for (k = 0; k < NAMPS; k++) {
-        gain2[k] = 0.;
+        gain2[k] = 0.;  /* e/DN, copied from gain below */
         noise2[k] = 0.;
```
mdlpstsci: Note to self: If there were time, many variable names need to be clarified.  In particular, gain2 is not a squared value (it seems), yet is named as if it were.  Readability of code is reduced.  

Note this comment is based upon the original code and not the changes made for the new ACSREJ algorithm.

pkg/acs/calacs/acsrej/acsrej_sky.c
nmiles2718: In response to @pllim comment, I was just about to send you an email @mdlpstsci and tell you that I do plan on writing an ISR on the new ACSREJ algorithm because I have been doing a lot of verification of the data products produced. I think it would be a great idea to get together and go over the code in detail, specifically, any changes that were made to the actual thresholding process for determining if a pixel was affected by a cosmic ray. At the present moment, I can say definitively that the ERR arrays are finally correctly handled and match what is expected an error propagation standpoint. I have some nice code to generate region files showing what pixels each implementation finds in common, as well as, what the pixels each finds separately from the other. In order to know if this is expected, I need to understand what code changes where made to the thresholding procedure. I plan on presenting all of this at the next team meeting, so if we could meet sometime before (11/15) that would be best, but meeting after that date is fine.

nmiles2718:
I noticed today that the final data product header value for the **CRSPLIT** keyword is set 1. Is this intended @pllim or @mdlpstsci? If I understand correctly, that keyword should contain the number of images that were combined to create the final data product. The behavior is the same for the current version of acsrej and so I might be misunderstanding what this keyword actually represents...

pkg/acs/calacs/acsrej/acsrej_sky.c
mdlpstsci: Understood.  I need to stop being distracted and finish my review.  I should make an outline for both of us regarding the new algorithm.

pllim:
@nmiles2718 , I am not sure either. As far as I can "grep", that keyword is not actually used or updated anywhere within ACSREJ. Some questions below might (or might not) shed some light:

* What is its value in the *input* images?
* Does its value change in final data product when `acsrej.e` is run on ASN files instead of a list of individual files?

nmiles2718:
In the single images the crsplit value is always 1. I tried that thinking that it was probably because acsrej didnt view the files as an association of CR split exposures. So I created an association file with the list of input images, ensuring that MEMTPYE was EXP-CRJ (since by default is is EXP-DTH). The result was the same.

I went and checked the superdark files, since they are made in a similar way. The crsplit value is correct for the basedark, the CRSPLIT value is the total number for the given anneal cycle.  Since I am basically creating the basedark by combining an entire anneals worth of darks I would expect the result to be the same. The major difference between the way the reference file pipeline combines the images to create the basedark vs the way I combine the files to create a basedark is that the pipeline uses **CALACS** with **CRCORR** set to **PERFORM**, as opposed to what I do which is by passing a list of files or an association file to acstools.acsrej.acsrej. I will test this by creating a combined image by passing CALACS an association file, it might be that this is done after acsrej finishes and hence outside **ACSREJ** task but within **CALACS**

mdlpstsci:
I do not find any instances of an explicit update to the CRSPLIT keyword in all of CALACS.  I suppose there could be some clever code with a variable mapping to the CRSPLIT keyword (doubtful), and I would have to be more clever to find this without more familiarity with the code.


Michele D. De La Pena
Space Telescope Science Institute
mdelapena@stsci.edu
667.218.6501

pllim:
@mdlpstsci , most keywords are mapped to some `struct` in `*.h` but a simple grep did not reveal its location to me as well. This needs some digging...

nmiles2718:
Unfortunately, I have to delete/re-install all of my conda environments because it is now **only** finding the new version of acsrej. Once I have this fixed, I will run the test of **CALACS** processing vs **ACSREJ** only.

In any case, I wanted to test the new version on a set of bias frames to see how it handled the errors arrays. In theory, it should have remained the same in the sense that the only contributing noise term is  the read noise. However, it appears that there is a poisson term being included with the bias frames which should not be (see plot).  
error_vs_signal_in_bias

nmiles2718:
**CRSPLIT Issue**: I just did some digging into the ACS Reference Pipeline and the reason that value is updated with the correct number of files for the ACS reference files is because it has been explicitly set in each input file prior to any processing. I tested this by updating the **CRSPLIT** keyword in the headers of all the FLT files with the number of FLT files in the list passed to **ACSREJ** (which was 48) and the result was that the CRSPLIT value in the output of acsrej was equal to 48.

pllim:
> In theory, it should have remained the same in the sense that the only contributing noise term is the read noise. However, it appears that there is a poisson term being included with the bias frames which should not be (see plot).

That is weird, because in both "old" and "new" algorithms, there is a special logic for when EXPTIME is zero. In that case, ERR would just be a fill value, instead of being calculated from ACSREJ. Hopefully @mdlpstsci can get into the bottom of this... (Sorry...)

> CRSPLIT... because it has been explicitly set in each input file prior to any processing

So, are you saying that this keyword was manually adjusted outside of HSTCAL? Maybe it is only for book-keeping purpose? @stsci-hack , do you remember the history behind this keyword?

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -403,304 +463,318 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr ipdq[],
...
+                                    /* Readnoise only for BIAS exposures.
+                                       nse is e^2 */
                                     sumvar[i] += nse[0];
                                 }
                             }
-                        }
+                        } /* End of loop over first amp used for line */
 
+                        /* SECOND AMP */
                         for (i = ampx; i < dim_x; i++) {
```
mdlpstsci: Just a note: These two "for loops" can be collapsed into a single loop where the nse[0|1] variable  just re-populates a temporary variable once (i > ampx).

nmiles2718:
**ERR ARRAYS Update**: I finally got everything set up again and @pllim suggested I try processing the bias frames using an association with **CALACS** to see if the error arrays were still improperly calculated. So, I took the same sets of raws (in **different** directories) created an association file for each, set the proper CALACS flags (**DQICORR=PERFORM**, **BLEVCOR=PERFORM**, **CRCORR=PERFORM**, everything else was set to **OMIT**), and processed one set using the current hstcal and the other with this version including the **ACSREJ** updates. The result was correct and as expected, CR's were flagged differently, but the error arrays were handled correctly in the sense that they only included a term from the read-noise for each amplifier.

Naturally, I wanted to double check my previous work running **ACSREJ** using  **acstools**. So I made another directory, copied the raw bias files over, processed them from scratch with the same flags as above, then passed the list of files to the python wrapper of ACSREJ in **acstools**. This resulted in incorrect ERR arrays in the final combined bias file as discussed above.

pllim:
@nmiles2718 , I suspect you still will get the same incorrect value if you run `acsrej.e` directly in the second case? Can you please try that? acstools is just a wrapper.

nmiles2718:
hey @pllim, you are correct executing ```acsrej.e `find . -path '*flt.fits' -print0 | tr '\0' ,` test_output_1.fits -v``` results in incorrect values populating the ERR arrays. Note, I am running from within a test environment when I call the above command and so it **is** grabbing the version of acsrej we are testing which can be seen in the initial lines of the output shown below.

```
(acsrej_test) nmiles@plhstins2:/user/nmiles/test_acsrej_ticket976/updated_acsrej/bias/WFC_ABCD_2.0_2017-06-26_using_executable$acsrej.e `find . -path '*flt.fits' -print0 | tr '\0' ,` test_output_1.fits -v
git tag: UNKNOWN
git branch: lim-calacs-ticket976
HEAD @: 331d591e8de487be2071897ace10291f390eaa59
ERROR:    Can't open file ./jd4yfklcq_flt.tra
ERROR:    Can't open file ./jd4yfmloq_flt.tra
ERROR:    Can't open file ./jd4yfotzq_flt.tra
ERROR:    Can't open file ./jd4za0bnq_flt.tra
ERROR:    Can't open file ./jd4yfqv2q_flt.tra
ERROR:    Can't open file ./jd4za2cbq_flt.tra
ERROR:    Can't open file ./jd4za4jmq_flt.tra
ERROR:    Can't open file ./jd4za6laq_flt.tra
ERROR:    Can't open file ./jd4za8q3q_flt.tra
ERROR:    Can't open file ./jd4zaaqlq_flt.tra
ERROR:    Can't open file ./jd4zacb1q_flt.tra
ERROR:    Can't open file ./jd4zaebqq_flt.tra
ERROR:    Can't open file ./jd4zagjuq_flt.tra
ERROR:    Can't open file ./jd4zaikaq_flt.tra
ERROR:    Can't open file ./jd4zakseq_flt.tra
ERROR:    Can't open file ./jd4zamstq_flt.tra
ERROR:    Can't open file ./jd4zaoagq_flt.tra
ERROR:    Can't open file ./jd4zaqb3q_flt.tra
ERROR:    Can't open file ./jd4zasjgq_flt.tra
ERROR:    Can't open file ./jd4zaujyq_flt.tra
ERROR:    Can't open file ./jd4zawryq_flt.tra
ERROR:    Can't open file ./jd4zayseq_flt.tra
ERROR:    Can't open file ./jd4zb0bpq_flt.tra
ERROR:    Can't open file ./jd4zb2c9q_flt.tra


CALBEG*** ACSREJ -- Version 9.2.0 (01-Jun-2017) ***
Begin    06-Nov-2017 17:33:55 EST


CRCORR   PERFORM
```

pllim:
For the record, when `acsrej.e` is invoked directly, it by-passed some pre-processing from upper-level `calacs.e`. Instead, it goes into the API through `mainrej.c`. So, the trick here is to figure out what skipped pre-processing step was instrumental in introducing this bug and how to apply it to standalone version.

The workaround for now is to use `calacs.e` when you try to perform ACSREJ in biases.

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -403,304 +463,318 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr ipdq[],
...
+                                val = (dum > 0.) ? dum : 0.;  /* e */
 
                                 if (newbias == 0) {
-                                    sumvar[i] += nse[0] + val/gn[0];
+                                    /* Variance term like DHB without SCALENSE
+                  https://trac.stsci.edu/ssb/stsci_python/ticket/1223#comment:13
+                                nse is e^2
+                                val is e (with sky)
+                                    */
+                                    sumvar[i] += nse[0] + val;
```
mdlpstsci: Note to self: This computation is wrong and has been for some time.  After tracing each factor in the equation to make sure of the units, the comment is correct.  nse is electrons2, but val is on electrons.

mdlpstsci: Need to double-check this as the comments are confusing.

nagrogin:
Approved for HSTDP 2018.1 by ACS Team Lead

jamienoss:
Original work exists [here](https://github.com/spacetelescope/hstcal/tree/lim-calacs-ticket976.prerebase) and the most recent pre-squash and final pre-rebase exists [here](https://github.com/spacetelescope/hstcal/tree/lim-calacs-ticket976-pre-squash-and-rebase). This branch includes the original 12 commits that might be needed for future bug tracking etc. These branches are protected.

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -879,10 +948,72 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr ipdq[],
...
+   ave    i: It is the median or minimum image, depending
+             on "initgues" provided by the user, in e/s, used for
+             comparison during CR rejection.
+   buferr i: ERR (electrons) of the image at the given line.
+   shadcorr i: SHADCORR values. All ones if no correction.
+   thresh_n o: Scrolling buffer of threshold for CR rejection.
+               Not sure why this needs to be scrolling but keeping it for now.
+   spthresh_n o: Scrolling buffer of threshold for SPILL marking.
+*/
+static void calc_thresholds(int line, int dim_x, int width,
```
jamienoss: Any `const`?

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -879,10 +948,72 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr ipdq[],
...
+                            float **thresh_n, float **spthresh_n) {
+    int i;
+    float dum, dum_nosky, pixsky;
+
+    /* Calculate the threshold for each pixel.
+       Compute the threshold based on the pixel values, ERR, and
+       shading correction. */
+    for (i = 0; i < dim_x; i++) {
+        /* Compute ERR^2 for threshold calculations. */
+        dum = SQ(buferr[i]);  /* e^2 */
```
jamienoss: `buferr` should be checked before accessed. Same with other passed in pointers.

pkg/acs/calacs/Dates
```diff
@@ -1,3 +1,6 @@
+ TBD        :  CALACS 9.1.1 ACSREJ calculations are now done entirely in
```
jamienoss: Needs to be updated

pkg/acs/calacs/History
```diff
@@ -1,4 +1,5 @@
-### 01-Jun-2017 - JAN -- Version 9.2.0
+### TBD - PLL  -- Version 9.1.1
```
jamienoss: Needs to be updated

pkg/acs/calacs/Updates
```diff
@@ -29,6 +29,9 @@ Update for version 9.2.0 - 01-Jun-2017 (JAN) : git diff f0c177019cb4c159c801e022
     pkg/acs/calacs/lib/wscript
     wscript
 
+Update for version 9.1.1 - TBD (PLL)
```
jamienoss: Needs to be updated

jamienoss:
@mdlpstsci @pllim sorry I fat fingered the review approval. Most of the above comments are all minor/trivial. There are a few mem access issues that may segfault. I'll defer to @mdlpstsci for final review approval on this.

jamienoss:
@mdlpstsci @pllim 
there's also the issue of the `#if false` block, whether that should remain?

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -831,27 +898,28 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr ipdq[],
...
 
     /* Use this marker to allow easier clean-up after an error condition.
        An error condition will have already set status to something else
-       that should be passed on...
-
-       Currently commented because not used.
+       that should be passed on... CURRENTLY NOT USED.
     */
-    /*cleanup: ;*/
+#if false
```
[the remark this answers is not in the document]
pllim: I really don't remember why I did this... :grimacing:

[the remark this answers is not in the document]
pllim: Looks like myself commented it out in 08acdf837f043b9. I think this change should go.

pkg/acs/calacs/acsrej/acsrej_init.c
```diff
@@ -27,266 +27,156 @@
...
-    int         ampx, ampy, detector, chip;
-    int         k, p;
-    short       dqpat;
-    float       exp2n, expn;
-    int         non_zero;
-    Hdr         dqhdr;
-    int         readnoise_only;
+    float     val, raw, dumf;
+    int       i, j, n, p, dum;
+    float     *buf;
```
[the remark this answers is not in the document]
mdlpstsci: Done.

pkg/acs/calacs/acsrej/acsrej_init.c
```diff
@@ -27,266 +27,156 @@
...
+    short     *bufdq;
+    int       *npts, *ipts;
+    short     dqpat;
+    Hdr       dqhdr;
 
-    void        ipiksrt (float [], int, int[]);
-    void        get_nsegn (int, int, int, int, float *, float*, float *, float *);
+    void      ipiksrt (float [], int, int[]);
```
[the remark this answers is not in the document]
mdlpstsci: Delayed until Issue#292 addressed.

pkg/acs/calacs/acsrej/acsrej_init.c
```diff
@@ -27,266 +27,156 @@
...
-                    /*
-                        Use sorted index array to match proper exptimes to
-                        selected pixels for use in ERR array calculation.
-                    */
-                    if ((dum/2)*2 == dum) {
-                        Pix(sg->sci.data,i,j) = (PIX(work,dum/2-1,i,nimgs) + PIX(work,dum/2,i,nimgs))/2.;
-                        expn = (exp2[ipts[dum/2-1]] + exp2[ipts[dum/2]]) / 2.;
+                    ipiksrt (&PIX(work, 0, i, nimgs), dum, ipts);
+                     /* Even number of input images for this pixel */
+                    if ((dum / 2) * 2 == dum) {
```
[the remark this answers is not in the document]
mdlpstsci: Delayed until Issue#292 addressed.

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -117,18 +121,73 @@ Code Outline:
...
+  15-Sep-2015   P.L. Lim    Added doc. Cleaned up codes.
+  21-Sep-2015   P.L. Lim    Fixed scrolling buffers [#1224]. Removed unnecessary
+                            processing by skipping images with zero exposure
+                            time. Removed duplicate buffer re-initializations.
+                            Cleaned up codes.
+  12-Jan-2016   P.L. Lim    Calculations now entirely in electrons.
+  14-Jan-2016   P.L. Lim    Replace threshold formula with ERR. Cleaned up
+                            threshold calculation function.
 */
+int acsrej_loop (IODescPtr ipsci[], IODescPtr iperr[], IODescPtr ipdq[],
```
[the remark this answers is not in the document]
mdlpstsci: Delayed until Issue#292 addressed.

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -154,7 +213,7 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr ipdq[],
 
     float   *sum;
     float   *sumvar;
-    float   *buf;
+    float   *buf, *buferr;
```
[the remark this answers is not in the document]
mdlpstsci: Done.

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -250,6 +307,7 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr ipdq[],
     sumvar = calloc (dim_x, sizeof(float));
 
     buf = calloc (dim_x, sizeof(float));
+    buferr = calloc (dim_x, sizeof(float));
```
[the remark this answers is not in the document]
mdlpstsci: Delayed until Issue#292 addressed.

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -403,304 +463,318 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr ipdq[],
...
+                                     efac[k], shadf_x, shadcorr);
+                        calc_thresholds(jndx, dim_x, ii, efac[k], exp2[k],
+                                        sig2, scale, ave, buferr, shadcorr,
+                                        thresh[k], spthresh[k]);
+                    } /* End of loop over each row in scrolling buffers */
                 } /* End loop over images */
             } /* End if...else line > 0 */
 
-            for (n = 0; n < nimgs; n++) {
+#if false
```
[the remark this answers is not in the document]
mdlpstsci: Removed after consultation with @pllim.

pkg/acs/calacs/acsrej/acsrej_loop.c
mdlpstsci: Will address this in Issue#293.

pkg/acs/calacs/acsrej/acsrej_loop.c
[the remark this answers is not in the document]
mdlpstsci: Done.

jamienoss:
Travis is experiencing issues with heir Mac builds. They then cancelled _everyones_ Mac builds whilst they investigate.

[Travis status](https://www.traviscistatus.com/incidents/mc9x2wmpnvhg)

Update
To help with the current situation, we are taking the hard decision of canceling Mac build jobs older than 2018-01-18 00:00 UTC. We apologize for this disruption and are thanking you for your understanding.
Posted about 5 hours ago. Jan 18, 2018 - 14:44 UTC

Investigating
We are sorry to inform you that you'll likely experience delays with your Mac builds (both private and open source) until further notice. Our team is currently addressing this. We are sorry for the inconvenience in the meantime and we are thanking you for your patience.
Posted about 5 hours ago. Jan 18, 2018 - 14:38 UTC

Further more, they seem to be having issues with checking out the correct commits. Travis is still trying to build b51a7270ba1fa5edc8b272356d156599b83b9bcd not 7dd0ad35c81c735dc902fc774a27fa00f68380d3

The configure log confirms this:

```
Setting top to                           : /home/travis/build/spacetelescope/hstcal 
Setting out to                           : /home/travis/build/spacetelescope/hstcal/build.Linux-4.14.12-041412-generic-x86_64-with-debian-jessie-sid 
Building app                             : HSTCAL 
Version                                  : 1.3.0-37-gb51a727 
git HEAD commit                          : b51a7270ba1fa5edc8b272356d156599b83b9bcd 
git branch                               : HEAD 
```

jamienoss:
I just built this with clang, it builds. Since Travis is not allowing Mac builds currently we should bypass the Travis build requirement and merge. @mdlpstsci if you're cool with this and merging as is, approve and I will merge it.

pllim:
**UPDATE FROM 2020:** Those Trac links are dead now. Thanks to @jhunkeler , I recovered the most important bit (at least on a personal level) where I wrote an essay about the scrolling buffer.

## Scrolling Buffer

To understand the bug, one has to understand a "scrolling buffer" in `acsrej_loop.c`. Scrolling buffer is basically an internal subarray of the image. It has the same columns as the image, but only stores a fixed number of rows that is a lot smaller than the image height. The current line (row) being processed is ''always'' at the center of the buffer.

```
                  <---+ "Upstream"
                      |
+---------------+     | <-- Current line being processed
|               |     |
|               | <---+ "Downstream"
|    Image      |
|    array      |
|               |         | Increasing line number
|               |         |
|               |         v
+---------------+
```

When the first line is processed, row 0 is inserted into center of the buffer. There is no real data "upstream" of the center, and they are ignored. All data "downstream" is also copied into the buffer.

For each subsequent line, the top-most "upstream" row in the buffer is discarded. All other rows in the buffer are moved up, such that `line` is now `line - 1`. A new row, the one after bottom-most "downstream", is copied into the buffer.

The last line is like the first line, but reversed. That is, there is no real data "downstream" of the center.

In the code, `[width]` of a scrolling buffer is equivalent to `[line]` of the image.

## The Bug

There are 4 scrolling buffers ''per image'': 

1. `mask` stores flagging information such as "HIT", "SPILL", or "EXCLUDE".
2. `pic` stores science data.
3. `thresh` stores thresholds for CR rejection ("HIT").
4. `spthresh` stores thresholds for marking surrounding pixels of a CR ("SPILL").

In reality, the scrolling functionality is only useful for `mask` and `spthresh`, because they need to access neighboring pixels. I am not sure why `pic` and `thresh` need to be "scrolling", but that is another story.

Now, `mask` and `pic` values are correct. ''However'', `thresh` and `spthresh` actually contain the same values as `pic` when the scrolling is performed. While `pic` goes through the proper unit conversion etc to ensure it has correct values, no such thing is done to ensure that the threshold buffers actually contain threshold values.

Farther down in the code, when `thresh` and `spthresh` is actually properly recalculated, the calculations are only done to `[width]` (the center of the buffer, that is also the current line being processed). Due to the scrolling nature of this buffer, all its rows "upstream" are calculated in previous line iterations (not to be confused with `iter`, which is the actual rejection iteration), and thus are correct. ''But'', its rows "downstream" are not yet recalculated, so they basically have unusable data. Although both `thresh` and `spthresh` are affected, only `spthresh` produces the bug because `thresh` does not really use the "upstream" and "downstream" rows.

Below is a snapshot of `spthresh` for a pixel that has detected CR and its neighboring pixels:
```
n = 0     (first image)
line = 2  (third row)
i = 1005  (IRAF X coord is 1006, which makes it AMP C)
nse[amp_index] = 4.439827  (squared readnoise in DN^2)
gn[amp_index]  = 2.017000  (gain in e/DN)
exp2n = 0.257962  (squared EXPTIME for that image in s^2)

spthresh[n][width][i] = sig2 * (nse[amp_index] + (val / gn[amp_index])) / exp2n

iter = 1
sig2[iter] = 30.250000
val = 398.859711  (central pix value from ave image in DN, with sky, clipped at zero)
spthres[n][width][i] = 23709.703125  ("SPILL" threshold for central pix)

                        +2.371E+04                          (spthres, iter=1)
            +2.371E+04  +2.371E+04  +2.371E+04              (spthres, iter=1)
+2.133E+04  +2.371E+04  +2.371E+04  +2.371E+04  +2.371E+04  (spthres, iter=1)
            +6.597E+02  +6.668E+02  +7.242E+02              (signal in electrons)  WRONG!!!
                        -1.984E+02                          (signal in DN/s, no sky)  WRONG!!!

iter = 2
sig2[iter] = 20.250000
val = 398.859711  (central pix value from ave image in DN, with sky, clipped at zero)
spthres = 15871.784180  ("SPILL" threshold for central pix)

                        +1.291E+04                          (spthres, iter=2)
            +1.435E+04  +1.587E+04  +1.587E+04              (spthres, iter=2)
+1.401E+04  +1.587E+04  +1.587E+04  +1.587E+04  +1.587E+04  (spthres, iter=2)
            +6.597E+02  +6.668E+02  +7.242E+02              (signal in electrons)  WRONG!!!
                        -1.984E+02                          (signal in DN/s, no sky)  WRONG!!!
```

The data is from my first WFC test image, `jc4823bpq_flt.fits` (the other one in the set is `jc4823brq_flt.fits`), that has the following properties:
```
READNSEA=        4.5700002E+00 / calibrated read noise for amplifier A
READNSEB=        3.9100001E+00 / calibrated read noise for amplifier B
READNSEC=        4.2500000E+00 / calibrated read noise for amplifier C
READNSED=        4.0400000E+00 / calibrated read noise for amplifier D

ATODGNA =        2.0200000E+00 / calibrated gain for amplifier A
ATODGNB =        1.8860000E+00 / calibrated gain for amplifier B
ATODGNC =        2.0170000E+00 / calibrated gain for amplifier C
ATODGND =        2.0109999E+00 / calibrated gain for amplifier D

EXPTIME = 0.5079

crthresh = 0.5555
sigma[iter] = [6.5, 5.5, 4.5]
sig2[iter] = [42.25, 30.25, 20.25]
```

## Proposed Solution

`thresh` and `spthresh` buffers needs to be recalculated properly when the scrolling is done, not after. Given the complexity of their calculations, it is easier to move that piece of the code out into a private function (`static void calc_thresholds(...)`) within `acsrej_loop.c` and then call that function (at several different places) where the entire scrolling buffers are manipulated (i.e., around the same places where `pic` is recalculated).

I also do not think `pic` and `thresh` need to be scrolling at all, but there are pros and cons to keeping them as-is:

* Pros: It is already implemented, so if we actually need their neighboring pixels info in the future, we can just use them as-is. If we remove them now, extra effort is necessary to re-implement.
* Cons: They currently take up unnecessary resources (computing time and memory).

PR #5, opened 2016-05-20
CTE sub-array 

brechmos-stsci:
The CTE sub-array changes were requested by Megan/Warren.  The code essentially reads in the sub-array data, creates AB and CD SingleGroups and puts the sub-array data into the appropriate AB/CD locations based on offset and size of the sub-array data. 

The code was validated in three ways:

1) The FLC and FLT results of the full-array data were compared from this code and from the original code and they were the same.  So this code did not affect the full-array data results.

2) A subset of the full-array data was cropped from chip 1, run through the new code and then the FLC and FLT were compared to cropped versions of the full-array data FLC / FLT from 1).

3) A subset of the full-array data was cropped from chip 2, run through the new code and then the FLC and FLT were compared to cropped versions of the full-array data FLC / FLT from 1).

An email will be sent to Megan that contains the Validation data. She will be the one (or will assign someone) to check and validate the code.

(I tried to attach the zip file of the validation code here but it would not attach)

### DIFF
--- a/pkg/wfc3/calwf3/wf3ccd/doccd.c
+++ b/pkg/wfc3/calwf3/wf3ccd/doccd.c
@@ -47,6 +47,9 @@
    This is part of the UVIS 2 update. SINK pixels are not currently 
    detected for subarrays or binned data.
     
+    C Jones May 2016
+    Made changes to the input of the sink-detect so that it will accept
+    sub-array data.
  */
 
 # include <string.h>
@@ -114,7 +117,9 @@ int DoCCD (WF3Info *wf3, int extver) {
     int GetCCDTab (WF3Info *, int, int);
     int GetKeyBool (Hdr *, char *, int, Bool, Bool *);
     int SinkDetect (WF3Info *, SingleGroup *);
-    
+    int FindLine (SingleGroup *, SingleGroupLine *, int *, int *,int *,
+            int *, int *);     
+
     /*========================Start Code here =======================*/	
     initSingleGroup (&x);
     if (wf3->printtime)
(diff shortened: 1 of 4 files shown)


pllim:
You can also dump the validation codes in your own GitHub account and link it here, if you wish. I think it is easier to look at than a ZIP file.

sosey:
@brechmos-stsci  I'm putting this here for completeness. Overall, good job! Delving into hstcal to make a science processing change with no fits or hst experience isn't straightforward. These are just my comments on this pull request, I've already included all the necessary changes in https://github.com/spacetelescope/hstcal/pull/10 since we didn't have time to work together on them. 

- remove the local loop var declarations, it's not wrong, I'm guessing you  didn't catch this  because you developed using clang on your mac. The current osx soft-links gcc to clang, which by default uses c99 styles. The gcc version we are using on the linux build is older and doesn't. Since we aren't building with c99, and it doesn't make a big timing difference, I'd prefer to keep the code uniform and move the loop declaration up top to avoid the errors.

- change the loop vars to match the rest of the code, in name and looping order. Since there are a lot of array operations with the CTE routines I'd like to keep the references uniform so that they are easy to follow. Because we have a non-square array, where  columns ~ 4 * rows, I initialized the large rotated arrays as RAZ_COLUMN (8412) by RAZ_ROWS (2070) so that the rows would be stored close to each other in memory. The CTE algorithm is correcting the pixels by going up and down columns in the reordered image, meaning we are going over rows the fastest. My hope was to make it as fast as possible and avoid page faults by keeping as much of the data currently in use in the cache.
- for the sink pixel routine, don't get the subarray location by asking for the relative location with the bias file, there may not always be a bias file associated with the image being processed so it's better to do the direct calculation
- move creation of the flag array for the CTE  to after the basic checks so it doesn't build the array when it's not necessary
- the CTE code has no need to carry around the ERR array, so the hardset there can be removed
- remove the #if 0 #endif statements as they are no longer needed
- don't use the nimsets==2 qualifier, use subarray == True
- for clarity, don't use double negative comparisons
- with subarrays implemented, you can do the CTE bias subtraction higher up, the bias routine should take care of cutting the correct portion out of the CTE bias image for the calibration. If it doesn't, it should be altered, don't make it force a full frame.
- **my oversight**, I forgot that not all the subarrays take physical overscan pixels - which are needed for the CTE correction. I've added a check for this.
- In general, I think it's nicer to make a subroutine to do the subarray manipulations. This should be callable by both the sink pixel and CTE arrays, and should go into the library. This keeps the main control-flow code clean, but also makes updates and bugs easier to do since you only have to update the routine which places the subarray into a full frame array.
- the cte correction isn't getting applied  to subarrays.  -- it's because there was no coded added to the CTE routines to actually deal with the difference between the subarrays and full arrays (ignoring places where there is no data by using the mask). This might have been a miscommunication on my end, so I've implemented the CTE internal algorithm.

PR #6, opened 2016-07-06
CALACS BLEVCORR for new subarrays

pllim:
Potential fix for one of the issues documented in https://trac.stsci.edu/ssb/stsci_python/ticket/1126 . This is needed for pipeline to be able to process new 2K subarrays added by FSW change done in May 2016.

@stsci-hack , what do you think of this fix? It ran successfully on the test image. Log attached:
[jd2j15j0q.tra.txt](https://github.com/spacetelescope/hstcal/files/350949/jd2j15j0q.tra.txt)

NOTE: Also need to make sure this pass in nightly test after merging...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+.lock-wafbuild
+.waf*
+bin.*
+build.*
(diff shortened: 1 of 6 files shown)


stsci-hack:
The changes look quite elegant and reasonable.  I see no problems with merging this request.

pllim:
Thanks, Warren! I'm going to merge this and check the regression tests tomorrow (if they are back online). Are they, @jhunkeler?

pllim:
Regression tests on 2016-07-07 using the new version confirmed that the fix does not affect existing data.

PR #16, opened 2016-09-22
**DO NOT MERGE** New FLSHCORR correction factor

### LINKED ISSUE #12: Add correction factor to FLSHCORR step in CALACS
Details from Andrea and Norman below. Targeted for OPUS delivery in Spring 2017.

**Norman**

In the FLSHCORR stage, there is a point at which the flash frame is input -- before subtraction from the flashed image to be corrected.

At that point, before the subtraction, we need to multiply each quadrant of the reference image by a separate correction factor that is dependent on the FLASHDUR keyword in the image to be corrected. Then the subtraction can occur, as usual (including the scaling of the tweaked reference image by the FLASHDUR).

If only a portion of that image is read in, because we are dealing with a subarray, then only apply the correction factor suitable to that subarray.

...

Think of this request as a correction to the flash reference file, rather than as a correction to flashed imagery.

If FLSHCORR is being applied equivalently to old and new subarray data --- which it should --- then this correction must also be applied to old and new subarray data.

If there is already some logic in place to handle flash-subtraction on jumbo/displaced subarrays (spanning the CCD mid-column line) [PL: No such thing is being done], then preserve that logic but adjust each quadrant of the (full-frame) flash reference image before excerpting to FLASHDUR-scale and subtract.

**Andrea**

The equations for AMPS A, B and D are of the form:

``` python
f(x) = 1 - ( a * sinh(b * x - 1) /  exp( c * x - 1) )
```

For AMP C:

``` python
f(x) = 1 + ( a / (x ** b + c) )
```

where `x` in the equations is the value of the FLASHDUR keyword in the header.

The coefficients are:

AMP A:

``` python
a = -0.27143973
b = 0.89499787
c = 1.64930466
```

AMP B:

``` python
a = -0.38018589
b =  1.63426073
c = 2.19362160
```

AMP C:

``` python
a = 0.04802005
b = 2.12968265
c = 0.01922919
```

AMP D:

``` python
a = -0.32215275
b = 1.19033580
c = 1.59900158
```

screenshot

pllim:
**DO NOT MERGE. Pending approval from ACS Team.**

Fixes #12 and bumps CALACS version.

Equivalent implementation in ACSTOOLS is spacetelescope/acstools#19.

### DIFF
--- a/pkg/acs/calacs/Dates
+++ b/pkg/acs/calacs/Dates
@@ -1,3 +1,4 @@
+ 22-Sep-2016:  CALACS 8.3.5 FLSHCORR now applies extra correction factor.
  07-Jul-2016:  CALACS 8.3.4 BLEVCORR modified to process new 2K subarrays.
  27-Jul-2015:  CALACS 8.3.3 ACSREJ can now process input list of any length.
  02-Mar-2015:  CALACS 8.3.2 ACSREJ can now process longer input list.
--- a/pkg/acs/calacs/History
+++ b/pkg/acs/calacs/History
@@ -1,3 +1,7 @@
+### 22-Sep-2016 - PLL -- Version 8.3.5
+    FLSHCORR now applies extra correction factor to post-flash reference image
+    prior to correction. The factor is calculated based on AMP and FLASHDUR.
+
 ### 07-Jul-2016 - PLL -- Version 8.3.4
     BLEVCORR modified to correctly process new subarrays added by FSW change in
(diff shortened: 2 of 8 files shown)


pllim:
Andrea rerun his analysis using #31 and concluded, "I can confirm that there is no longer need of a post-flash correction fix. The only minuscule thing (that does not affect the flash analysis, but maybe it is worth checking) is that amps A are a little lower than the other amps."

PR #22, opened 2016-10-21
BLEVCORR to process new pol/ramp

### LINKED ISSUE #19: BLEVCORR cannot handle polarizer/ramp using new subarray format
In the original new subarray change request (#6), only the new apertures were expected to use the new subarray format. Now, polarizers and ramp exposures also use the new subarray format. Therefore, BLEVCORR failed to remove the virtual overscan properly for them. Note that these polarizer/ramp apertures are not new, so we need to be careful to make any change fully backward compatible. ACS Team is deciding on a course of action.

pllim:
BLEVCORR subtracts virtual overscan from polarizer and ramp using new readout format. Fixes #19. Supersedes #21.


pllim:
Regression tests passed, indicating that no existing data in the tests are affected. Merging.

PR #24, opened 2016-11-14
HCD-8 Uninitialized variable 'i' in calacs prinfo.c

jamienoss:
https://jira.stsci.edu/browse/HCD-8

### DIFF
--- a/pkg/acs/calacs/lib/prinfo.c
+++ b/pkg/acs/calacs/lib/prinfo.c
@@ -257,7 +257,7 @@ char *GetTime (void) {
 
 /* Convert a string to upper case and pad with blanks. */
 
-void UpperAll (char *instr, char *outstr, int maxch) {
+void UpperAll (const char *instr, char *outstr, int maxch) {
 
 /* arguments:
 char *instr   i: input string
@@ -292,7 +292,7 @@ int maxch     i: allocated length of outstr (including EOS)
    to lower case, and pad with blanks.
 */
 
-void Upper1 (char *instr, char *outstr, int maxch) {
+void Upper1 (const char *instr, char *outstr, int maxch) {
 
 /* arguments:
 char *instr   i: input string
@@ -300,7 +300,7 @@ char *outstr  o: string with only the first letter capitalized
 int maxch     i: allocated length of outstr (including EOS)
 */
 
-	int i;
+	int i = 0;
 
 	if (PAD_SIZE >= maxch) {
 	    strcpy (outstr, "*****");


jamienoss:
@pllim please review.

pllim:
@jamienoss , thanks for the ping. However, in the future, can we just keep bug reports on GitHub for now until there is a clear integration between JIRA and GitHub issues? I find it a pain to have to login to yet another system to see the actual bug report.

pkg/acs/calacs/lib/prinfo.c
```diff
@@ -292,15 +292,15 @@ int maxch     i: allocated length of outstr (including EOS)
...
+void Upper1 (const char *instr, char *outstr, int maxch) {
 
 /* arguments:
 char *instr   i: input string
 char *outstr  o: string with only the first letter capitalized
 int maxch     i: allocated length of outstr (including EOS)
 */
 
-	int i;
+	int i = 1;
```
pllim: So it looks like `i` can only be un-initialized if `instr[0] == '\0'`. So, isn't it safer to just return `outstr[0] = '\0'` in that case instead of initializing `i = 1`, which does not help in that case?

jamienoss: What about the padding? My change fixes the bug only, without changing current implementation.

pllim: When you set `'\0'` as first element, it means a null string, right? There is no need to pad a null string, as there is nothing to pad. Or did I misunderstand the code? We can always consult @stsci-hack as he wrote the original CALACS.

jamienoss: I'm with you it would be much better to do
```
if (!instr || instr == '\0')
{
    outstr = '\0';
    return;
}
```

but the code looks like it pads and then even adds another null char at the end of the pad.

pllim: IMHO the original code just never accounted for null string case. If I have my way, I would initialize `i = 0` *and* add your logic above. Your current fix only gets rid of the compilation warning but does not really address the logic for null string.

philhodge:
It seems to me that i should be initialized to 0, and that's all you 
need to change.  The stis version appears to have the same bug.

The padding with blanks out to PAD_SIZE is for alignment in the printed 
messages.

pllim:
> The padding with blanks out to PAD_SIZE is for alignment in the printed messages.

Oh, I get it now! In that case, yes, I agree with @philhodge .

jamienoss:
@philhodge - If the only change is to initialize this to '0' then when a null terminated string is passed in and is padded for the print alignment you'll just get a blank string - is this intended or should there be an actual message (not from this function) to state that the string is 'empty', for whatever (don't have eclipse open right now to look) it is called by? Either way, it doesn't feel like this should be the responsibility of a function named 'Upper1'? I had questioned whether the padding loop should be contained within the outer IF condition but wasn't really sure why the padding was there at all, so I wanted to fix the possible mem corruption rather than change implementation.

philhodge:
I agree completely that i has to be assigned a value before being used as an index in outstr.  There should never be a case where Upper1 is called with a null string as input, but "there should never be a case" is no argument for not fixing it.

As an example of the use of Upper1, you can look at the PrGrpBegin function.  Upper1 converts the `label` argument in PrGrpBegin to upper case for the first letter and lower case for the rest of the string.  I don't know how (or if) PrGrpBegin is used in calacs, but in calstis `label` will be something like "Imset" or "Order".  When Upper1 is called, its argument `instr` will be the label followed by (usually) an integer value, e.g. "Imset 37".  PrGrpBegin constructs the complete line to be written, which will be that label and number, followed by "Begin", and also by the date and time.  The padding is just so "Begin" starts at the same column number in each message; it isn't necessary.  PrGrpBegin will not print a blank message under any circumstances.  If someone wrote a function to call Upper1 with a null string as input, well, you would get a string of eight blanks as output.  But in prinfo.c, Upper1 is used for the first word of a message, not the whole message.

pllim:
>  it doesn't feel like this should be the responsibility of a function named 'Upper1'

Let's not open that can o' worms. We don't need to rewrite the whole HSTCAL.

jamienoss:
>>  it doesn't feel like this should be the responsibility of a function named 'Upper1'
>
>Let's not open that can o' worms. We don't need to rewrite the whole HSTCAL. 

Exactly, which is why I only fixed the possible corruption and then ran - TBH I hesitated to do even that in case I opened the closet door and was engulfed by an avalanche!

pllim:
So in conclusion based on the discussions so far, it is fine to just initialize but to 0 instead of 1. Once you make that change, it is good to merge. Thanks!

jamienoss:
Reverted and moved to https://github.com/spacetelescope/hstcal/pull/29

PR #33, opened 2016-11-23
Initial commit of Travis-CI configuration

jhunkeler:
🎉  (or at least I hope so)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,59 @@
+language: generic
+
+matrix:
+  fast_finish: true
+
+  include:
+  - os: linux
+    sudo: false
+    python: 2.7
+    compiler: gcc
+
+  - os: linux
+    sudo: false
+    python: 3.5
+    compiler: gcc
+
+  - os: osx
+    python: 2.7
+    env:
+      - CFLAGS="-m64"
+      - LDFLAGS="-m64"
+
+  - os: osx
+    python: 3.5
+    env:
+      - CFLAGS="-m64"
+      - LDFLAGS="-m64"
+
+  - os: osx
+    python: 2.7
+    compiler: clang
+    env:
+      - CC="clang"
+      - CFLAGS="-arch x86_64"
+      - LDFLAGS="-arch x86_64"
+
+  - os: osx
+    python: 3.5
+    compiler: clang
+    env:
+      - CC="clang"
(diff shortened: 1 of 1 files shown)


jhunkeler:
This commit's results:
https://travis-ci.org/jhunkeler/hstcal/builds/178436258

sosey:
basic viability, nice :8ball:

pllim:
Should we include a small 512 subarray test for each CALxxx? Those should be fast to process.

jhunkeler:
As long as it doesn't take longer than the time limits:
https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts

sosey:
I'm don't think it's worth sitting on travis for a data test. Though the science array is small, many of the steps take full frame reference files, which also have to be transferred, and then the test is still just an aliveness check. I think there is also still a concurrency limit for the number of test running per organization. We have the nightly data regression for actual validation which needs to be looked at with each change anyways. I think the compile validation is decent for the pull requests right now to make sure we aren't outright  breaking the nightly dev build. When/If we move regression systems over to something besides Pandokia we'll have more local resources to throw at a full test for each change. In the mean time, coders should continue to run their own data tests during development and not request a pull until they see results they are expecting. I.E., you should minimally make sure the full package compiles, runs, and produces reasonable results, worthwhile to check on the build machines too and preferably against the test suite, before making the pull request to the hstcal master repo. All intermediate developing  should be done using your personal repository.

pllim:
Good point on the ref files.

However, I can't dig out any specific info on concurrent limit per org. Travis website says "fair use" but implies the limit is shared between all open source repos (?). I was just discussing this with Justin last week. So if you know if this limit is explained anywhere, please let me know. Thanks.

jhunkeler:
"Fair Use" to me implies the open source plan is load balanced aggressively, so if we start to abuse cputime, it will start halting/canceling/queuing/deferring our jobs. I think each virtual image (i.e. job) is limited to 4 threads and has a maximum run time of 120 minutes before being dropped.

jhunkeler:
@sosey Do you mind if I merge this? Or do you need me to implement anything else?

sosey:
@jhunkeler oh, sorry, I thought you had. Yes, merge away

jhunkeler:
WILCO :)

PR #34, opened 2016-11-28
Remove non-existent --yes option from install target

jhunkeler:
Fixes:

```bash
$ make install
[...]
waf: error: no such option: --yes
make: *** [install] Error 2
```

### DIFF
--- a/configure
+++ b/configure
@@ -88,7 +88,7 @@ all-progress:
 	$WAF -p build
 
 install:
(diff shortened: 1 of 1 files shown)


pllim:
Maybe merge #33 first so Travis tests something more meaningful?

jhunkeler:
That's the plan ;)

PR #37, opened 2016-11-29
Address -Wsometimes-uninitialized warnings

### LINKED ISSUE #36: Address -Wsometimes-uninitilized warnings

jamienoss:
This PR doesn't ultimately 'fix' the issue, though it does temporarily plug the holes preventing these bugs from impacting users results.

Fixes #36 

### DIFF
--- a/pkg/stis/calstis/cs1/dodqi.c
+++ b/pkg/stis/calstis/cs1/dodqi.c
@@ -10,6 +10,7 @@
 
 # include <stdio.h>
 # include <string.h>
+#include <assert.h>
 
 # include "c_iraf.h"
 # include "hstio.h"
@@ -652,6 +653,8 @@ int doppmin, doppmax  i: offsets for Doppler shift
 	    xlength = 2;
 	    ylength = tabrow->length * 2;
 	}
+	else
+	    assert(0);
 
 	nx = ydq->nx;
 	ny = ydq->ny;
--- a/pkg/stis/calstis/cs7/evaldisp7.c
+++ b/pkg/stis/calstis/cs7/evaldisp7.c
@@ -1,4 +1,5 @@
 # include <stdio.h>
+#include <assert.h>
 
 # include "stis.h"
 # include "calstis7.h"
@@ -97,6 +98,8 @@ double GetWavelength (DispRelation *disp, int sporder, double ix,
 	    wl_test = prismDisp (disp->coeff, ix);
 
 	}
+	else
+	    assert(0);
 
 	return (wl_test);
 }
--- a/pkg/stis/calstis/cs7/scalewcs.c
+++ b/pkg/stis/calstis/cs7/scalewcs.c
@@ -1,4 +1,5 @@
 # include <stdio.h>
+#include <assert.h>
 
 # include "stis.h"
 # include "calstis7.h"
@@ -69,6 +70,8 @@ CoordInfo *coord_o  io: size and coordinate info for output
 	    plate_scale = coord_o->cdelt[1];	/* CDELT2 */
 	else if (sts->dispaxis == 2)
 	    plate_scale = coord_o->cdelt[0];	/* CDELT1 */
+	else
+	    assert(0);
 
 	/* scale the plate scale to image pixels */
 	sts->plate_scale[0] = plate_scale / sts->ltm[0];
--- a/tables/c_tbtopn.c
+++ b/tables/c_tbtopn.c
@@ -1,6 +1,7 @@
 # include <stdlib.h>
 # include <string.h>
 # include <fitsio.h>
+#include <assert.h>
 # include "ctables.h"
 
 IRAFPointer c_tbtopn (char *tablename, int iomode, IRAFPointer template) {
@@ -90,6 +91,8 @@ function value          o: table descriptor
                         return NULL;
                     }
                 }
+                else
+                    assert(0);
                 tbl_descr->hdunum = hdunum;
                 tbl_descr->hdutype = hdutype;
 
--- a/pkg/stis/calstis/cs1/findover.c
+++ b/pkg/stis/calstis/cs1/findover.c
@@ -3,6 +3,7 @@
 # include "calstis1.h"
 # include "stiserr.h"
 # include "stissizes.h"	/* detector sizes, overscan sizes */
+#include <assert.h>
 
 # define FULL_FRAME_READOUT  1
 # define SUBARRAY_READOUT    2
@@ -160,6 +161,8 @@ int vx[2], vy[2]      o: range of pixel numbers for virtual overscan region
 	    biassect[0] = (nx - A_bsect[1] - 1);
 	    biassect[1] = (nx - A_bsect[0] - 1);
 	}
+	else
+	    assert(0);
 
 	/* Sanity check on image size. */
 	if (readout == FULL_FRAME_READOUT) {
--- a/pkg/stis/calstis/lib/prinfo.c
+++ b/pkg/stis/calstis/lib/prinfo.c
@@ -322,7 +322,7 @@ char *outstr  o: string with only the first letter capitalized
 int maxch     i: allocated length of outstr (including EOS)
 */
 
-	int i;
+	int i = 0;
 
 	if (PAD_SIZE >= maxch) {
 	    strcpy (outstr, "*****");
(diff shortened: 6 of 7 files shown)


pkg/stis/calstis/cs1/dodqi.c
```diff
@@ -652,6 +653,8 @@ int doppmin, doppmax  i: offsets for Doppler shift
 	    xlength = 2;
 	    ylength = tabrow->length * 2;
 	}
+	else
+	    assert(0);
```
pllim: I don't know about this. The standard practice in HSTCAL is to set `status` to the appropriate error code and then return it (which will result in HSTCAL terminating with some error code), and also to write the error message to trailer file.

jamienoss: Asserts are used not to inform the user but to inform the developer (though probably via the user) when they have miscomputed the boundaries to their logic. E.g. in the cases where the developer is pretty sure that some logic will always be true but adds an assert just to be safe. When the assert is triggered at runtime the process exits gracefully (think of this as more of a nice by-product rather than the main feature) and prints the asserted expression, the culprit file and line number (the main feature).

```
#include <assert.h>

int main(void)
{
  int myLogicIsFaultless = 0;
  int whatEver = 1;
  assert(myLogicIsFaultless == whatEver);
  return 0;
}
```

Produces
```
Assertion failed: (myLogicIsFaultless == whatEver), function main, file asserttest.c, line 7.
Abort trap: 6
```

sosey: assert(0) isn't a graceful exit, I believe it calls abort() (as printed above) which doesn't do cleanup. abort core dumps are not a great way to go. I think we  should minimally set the appropriate return flag and report the problem instead with a useful error statement, and add any cleanup code which is necessary for termination.

pllim: We also have to consider that most of the pipeline processing happen as a blackbox on the DP side. When something fails, the first thing they tend to look at are the trailer files. Just an abort message to screen is not going to help them.

jamienoss: I agree to all (as stated by opening caveat) that this is not 'the' fix but simply a temporary prevention of buggy results.

@sosey yep, I should have been more explicit and expanded my example, and also remembered that C has no exception handling and thus there's no way for the try-catch block that I assumed existed within all of our main() funcs to actually exist.

tables/c_tbtopn.c
```diff
@@ -90,6 +91,8 @@ function value          o: table descriptor
                         return NULL;
                     }
                 }
+                else
```
philhodge: Please remove the changes to tables/c_tbtopn.c.  It is not an error for `hdunum` to be different from 1.

philhodge:
If you're going to add 'else assert(0)', you should have a test case 
that demonstrates that the 'else' path could actually be taken.  If it 
can't, the 'else' implies that there is another case that isn't taken 
into consideration.  It also implies that that case would be an error, 
which is not necessarily true.  An 'if then else' does not have to be an 
exhaustive list.

When you change calstis code, the STIS team will need to test those 
changes.  How are you going to suggest that they test the changes?  It 
looks to me as if none of the 'assert(0)' changes to the calstis code 
could be considered useful.

The changes to c_tbtopn.c should be removed.  There was a single if 
statement regarding hdunum, and you added 'else assert(0)'. That's just 
wrong.

philhodge:
Thanks for pointing this out; I should have seen this before.

If hdunum is not 1, then `tbl_descr->hdutype = hdutype` would assign 
garbage, as you said.  However, the call to tbSaveInfo immediately 
following does actually get the hdutype by calling `fits_get_hdu_type`, 
and then it executes `tbl_descr->hdutype = hdutype` again, but with the 
correct value of hdutype.

pkg/stis/calstis/cs7/scalewcs.c
```diff
@@ -69,6 +70,8 @@ CoordInfo *coord_o  io: size and coordinate info for output
 	    plate_scale = coord_o->cdelt[1];	/* CDELT2 */
 	else if (sts->dispaxis == 2)
 	    plate_scale = coord_o->cdelt[0];	/* CDELT1 */
+	else
+	    assert(0);
```
jamienoss: ../pkg/stis/calstis/cs7/scalewcs.c:70:11: warning: variable 'plate_scale' is used uninitialized whenever 'if' condition is false [-Wsometimes-uninitialized]
        else if (sts->dispaxis == 2)
                 ^~~~~~~~~~~~~~~~~~
[247/615] Compiling pkg/stis/calstis/cs11/gettimes11.c
../pkg/stis/calstis/cs7/scalewcs.c:74:24: note: uninitialized use occurs here
        sts->plate_scale[0] = plate_scale / sts->ltm[0];
                              ^~~~~~~~~~~
../pkg/stis/calstis/cs7/scalewcs.c:70:7: note: remove the 'if' if its condition is always true
        else if (sts->dispaxis == 2)
             ^~~~~~~~~~~~~~~~~~~~~~~
../pkg/stis/calstis/cs7/scalewcs.c:40:20: note: initialize the variable 'plate_scale' to silence this warning
        double plate_scale;     /* before scaling according to binning */
                          ^
                           = 0.0

pkg/stis/calstis/cs7/evaldisp7.c
```diff
@@ -97,6 +98,8 @@ double GetWavelength (DispRelation *disp, int sporder, double ix,
 	    wl_test = prismDisp (disp->coeff, ix);
 
 	}
+	else
+	    assert(0);
```
jamienoss:
```
../pkg/stis/calstis/cs7/evaldisp7.c:95:13: warning: variable 'wl_test' is used uninitialized whenever 'if' condition is false [-Wsometimes-uninitialized]
        } else if (disp_type == PRISM_DISP) {
                   ^~~~~~~~~~~~~~~~~~~~~~~
../pkg/stis/calstis/cs7/evaldisp7.c:101:10: note: uninitialized use occurs here
        return (wl_test);
                ^~~~~~~
../pkg/stis/calstis/cs7/evaldisp7.c:95:9: note: remove the 'if' if its condition is always true
        } else if (disp_type == PRISM_DISP) {
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../pkg/stis/calstis/cs7/evaldisp7.c:73:33: note: initialize the variable 'wl_test' to silence this warning
        double wl_high, wl_low, wl_test, x_test;
                                       ^
                                        = 0.0
```

tables/c_tbtopn.c
```diff
@@ -90,6 +91,8 @@ function value          o: table descriptor
...
+                    assert(0);
```
jamienoss:
```
../tables/c_tbtopn.c:83:21: warning: variable 'hdutype' is used uninitialized whenever 'if' condition is false [-Wsometimes-uninitialized]
                if (hdunum == 1) {
                    ^~~~~~~~~~~
../tables/c_tbtopn.c:94:38: note: uninitialized use occurs here
                tbl_descr->hdutype = hdutype;
                                     ^~~~~~~
../tables/c_tbtopn.c:83:17: note: remove the 'if' if its condition is always true
                if (hdunum == 1) {
                ^~~~~~~~~~~~~~~~~
../tables/c_tbtopn.c:24:20: note: initialize the variable 'hdutype' to silence this warning
        int hdutype;            /* type of HDU */
                   ^
                    = 0
1 warning generated.
```

philhodge:
`sts->dispaxis` is gotten from a header keyword in getgrpinfo7.c, and 
the value is tested at that point:

     if (sts->dispaxis < 1 || sts->dispaxis > 2) {
         printf ("ERROR    Dispaxis = %d is invalid.\n", sts->dispaxis);
         return (GENERIC_ERROR_CODE);
     }

philhodge:
The only calls to GetWavelength are in do2dx.c, and in that file, 
`disp_type` is set to either PRISM_DISP or GRATING_DISP before calling 
GetWavelength.

philhodge:
`tabrow->axis` is gotten by function ReadBpixTab in dodqi.c, and there's 
a check on the value:

         if (tabrow->axis !=1 && tabrow->axis != 2) {
             printf ("ERROR    Axis = %d in BPIXTAB, but it must be 1 or 
2.\n",
                         tabrow->axis);
             c_tbtclo (tabinfo->tp);
             return (TABLE_ERROR);
         }

philhodge:
`sts->ccdamp` is gotten in file getkeyinfo1.c, and there's a test on its 
value:

     if (strchr ("ABCD", sts->ccdamp[0]) == NULL) {
         printf ("ERROR    CCDAMP = `%s' is invalid.\n", sts->ccdamp);
         return (GENERIC_ERROR_CODE);
     }

There is also a test that the value is no more than one character.

PR #41, opened 2016-11-30
Have gcc supersede clang when finding C compiler

### LINKED ISSUE #42: Have GNU supersede clang?
Might it be an idea to have the configure script default to gnu compilers rather than clang?

jamienoss:
Fixes #42 

### DIFF
--- a/wscript
+++ b/wscript
@@ -32,6 +32,11 @@ SUBDIRS = [
 # Have 'waf dist' create tar.gz files, rather than tar.bz2 files
 Scripting.g_gz = 'gz'
 
+# Have gcc supersede clang
+from waflib.Tools.compiler_c import c_compiler
+c_compiler['darwin'] = ['gcc', 'clang']
+c_compiler['default'] = ['gcc', 'clang']
+
 option_parser = None
 def options(opt):
     # We want to store


jamienoss:
@jhunkeler please take a look and play around, and see what you think. Cheers

sosey:
gcc silently links to the clang on current installs of MacOS, I'm not sure this gets around that issue, have you tested it?

Also, there are no plans to translate hstcal to c++, I think putting the future proofing statements in would be misleading to others looking at the code.

jamienoss:
@sosey Correct, in fact on 10.11 it's even a renamed copy rather than a link. I had tested this and even looked at the waf source - it runs the preprocessor and then searches through the output to determine if it's actually gcc rather than an imposter. @jhunkeler knows more about waf than I do, hence my message to have him look at it to extend any possible further testing etc.

jamienoss:
@sosey done :)
@jhunkeler does this change look ok to you?

jhunkeler:
For the record `/usr/bin/gcc` on OS X is just a link back to `clang`, however I like this because it will automagically discover our real gcc compiler while building under conda instead of exploding.

👍

sosey:
@jhunkeler  yah, I'm down with that. Perhaps we should make a note for repo users that if they are building outside of our conda environment they may experience clang related turbulence?

PR #43, opened 2016-12-07
Platform check

jhunkeler:
Removes the need to maintain a list of valid operating system versions, and their cutesy names over time. Instead, now we check whether the host operating system (OS X) is greater than or equal to 10.5.0. If not, the configuration step will fail out to the shell (i.e. exit 1). The probability of external users compiling `hstcal` on 10.4.x "Tiger" is rather low anyway at this point.

Due to subtle ambiguities we have experienced recently in the Conda world I added a visual dump for each compiler's `FLAGS` variable. This allows us to immediately determine whether C/FC/LDFLAGS were set correctly at compile-time or not.

The configuration order has changed slightly as well. CFITSIO is now configured after the global CFLAGS are defined. This allows the hidden CFITSIO `./configure` call to assimilate the FLAGS it needs to match the parent program's linkage. After moving it to the bottom I stopped receiving linker errors (I used to have to set `CFLAGS=-m64` and `LDFLAGS=-m64` manually as a workaround).

### Previous Order

- C compiler check
- Fortran compiler check
- Check OSX version
- Determine sizeof `int`
- Setup OpenMP
- Configure CFITSIO
- Setup FLAGS

### New Order

- C compiler check
- Fortran compiler check
- Check OSX version
- Setup OpenMP
- Determine sizeof `int`
- Setup FLAGS
- Dump FLAGS
- Configure CFITSIO

### DIFF
--- a/wscript
+++ b/wscript
@@ -69,6 +69,43 @@ def _setup_openmp(conf):
     else:
         conf.end_msg("OpenMP found.", 'GREEN')
 
+def _check_mac_osx_version(version):
+    '''
+     Native     Encoded
+    -------    --------
+     10.5.0 == 0x0A0500
+      ^ ^ ^       ^ ^ ^
+      | | |       | | |
+      major       major
+        | |         | |
+        minor       minor
+          |           |
+          patch       patch
+
+    version -> 24-bit encoded version
+    '''
+
+    assert isinstance(version, int)
+    floor_version = version
+    osx_version = 0
+
+    floor_version_major = (floor_version >> 8 & 0xff)
+    floor_version_minor = (floor_version >> 8 & 0xff)
+    floor_version_patch = (floor_version & 0xff)
+
+    s = platform.popen("/usr/bin/sw_vers -productVersion").read()
+    osx_version_major, osx_version_minor, osx_version_patch = \
+        tuple(int(x) for x in s.strip().decode().split('.'))
+
+    osx_version = (osx_version << 8 | osx_version_major & 0xff)
+    osx_version = (osx_version << 8 | osx_version_minor & 0xff)
+    osx_version = (osx_version << 8 | osx_version_patch & 0xff)
+
+    if osx_version < floor_version:
+        return False
+
+    return True
+
 def _determine_mac_osx_fortran_flags(conf):
     # On Mac OS-X, we need to know the specific version in order to
     # send some compile flags to the Fortran compiler.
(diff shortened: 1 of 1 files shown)


wscript
```diff
@@ -69,52 +70,59 @@ def _setup_openmp(conf):
...
+          patch       patch
+
+    version -> 24-bit encoded version
+    '''
+
+    assert isinstance(version, int)
+    floor_version = version
+    osx_version = 0
+
+    floor_version_major = (floor_version >> 8 & 0xff)
```
pllim: Fancy but unreadable. Can anyone besides Joe debug or expand this in the future? If the answer is no, then I recommend something simpler. This script is only run once at build time, so performance is not an issue.

wscript
```diff
@@ -126,6 +134,48 @@ def _determine_sizeof_int(conf):
...
+        try:
+            flags = os.environ[flag].split()
+        except KeyError:
+            return None
+
+        for arg in flags:
+            for arch in arch_modes:
+                if arg == arch:
+                    return arch
+
```
pllim: Nitpicking: Remove extra blank line.

wscript
```diff
@@ -126,6 +134,48 @@ def _determine_sizeof_int(conf):
...
+        conf.end_msg('x86')
+
+    conf.start_msg('User-defined CFLAGS')
+    conf.end_msg(get_forced_arch('CFLAGS'))
+
+    conf.start_msg('User-defined LDFLAGS')
+    conf.end_msg(get_forced_arch('LDFLAGS'))
+
+    if get_forced_arch('CFLAGS') is None:
+        if ('CFLAGS' not in conf.env) or \
```
pllim: Nitpicking: `\` is discouraged now. You can use extra parenthesis to do the same kind of breaking the logic into multiple lines.

wscript
```diff
@@ -167,16 +217,19 @@ def configure(conf):
...
 
+    _setup_openmp(conf)
+
     _determine_sizeof_int(conf)
 
     # The configuration related to cfitsio is stored in
     # cfitsio/wscript
     conf.recurse('cfitsio')
 
+
```
pllim: Nitpicking: Extra blank line is unnecessary.

wscript
jamienoss: I agree, it looks quite unnecessary.

wscript
```diff
@@ -126,6 +134,48 @@ def _determine_sizeof_int(conf):
...
+        for arg in flags:
+            for arch in arch_modes:
+                if arg == arch:
+                    return arch
+
+
+    conf.start_msg('Checking architecture')
+    if platform.architecture()[0] == '64bit':
+        arch_mode = '-m64'
+        conf.end_msg('x86_64')
```
jamienoss: See comment below.

wscript
pllim: I, for one, would not want to touch it ever again. But if you already approved it, then fine.

pllim: At least explain what the logic is trying to do in the docstring?

sosey: just to reiterate - me approving of the change doesn't mean we can't discuss it's value, I believe in everyone working together as a team and no one person dominating or mandating the conversation. If everyone thinks this section is better kept as human readable as possible we can revert to the old way.  I agree that another line of documentation would be good to add if we keep the current function.

jamienoss: My vote would be to revert to human readable (sorry dude). Build infrastructure is a dark art already, we don't need to restrict it further by upgrading access from nija to jedi.

philhodge:
Is this right?

+ floor_version_major = (floor_version >> 8 & 0xff)
+ floor_version_minor = (floor_version >> 8 & 0xff)
+ floor_version_patch = (floor_version & 0xff)

In the first line don't you want to shift by 16 bits instead of 8?

jhunkeler:
Guys, I'm working out all of the kinks here. Sorry for the apparent sloppiness I was just "jotting down" the things I wanted to do last night without checking everything I was working on.

I'm fleshing out all of the problems.

sosey:
:8ball: update -   no apologies necessary, iterations are expected :)

jhunkeler:
Just to explain a little what you saw before:
I had originally planned to allow `_check_mac_osx_version()` to accept a string `"x.y.z"`, then convert it to an integer and perform the comparison. I dropped the idea but didn't clean up the code.

Are bitwise operations in Python frowned upon or something? Maybe I've been staring into the x86 assembly abyss for far too long. ;)

philhodge:
On 12/07/2016 02:37 PM, Joseph Hunkeler wrote:
> Are bitwise operations in Python frowned upon or something?

They're part of the language.  As far as I'm concerned, feel free to use 
them.  'a >> b' is a lot easier to understand than some of the other 
language features.

pllim:
@jhunkeler , thanks for the expanded docstring!

PR #64, opened 2017-02-24
CALACS to have new SINKCORR in ACSCCD (APPROVED)

### LINKED ISSUE #44: New SINKCORR for CALACS
CALACS wants to do SINK pixel flagging like CALWF3 but might use a slightly different scheme of flagging.

~~Currently not sure if this should be done by Jamie as part of #13 (new PCTECORR) or in parallel by me (in case Jamie does not have time). I think Megan did both at the same time.~~

**Implementation Details**

Following details are extracted mostly from links below, and some from private communications:

* New reference files with suffix `*_snk.fits` with DQ values 1024 for affected pixels. This file will be populated by CRDS into **SNKCFILE** FITS header keyword in the `PRIMARY` header of RAW image.
* New flag named **SINKCORR** in  the `PRIMARY` header of RAW image, stating whether to apply `SNKCFILE` or not.
* Should only be applied if `BIASCORR=COMPLETE`.
* Format: The file is a two-dimensional FITS MEF but does not contain ERR or DQ image extensions. There are 2 SCI image extensions, Chip 2 in `EXT=1` (CD) and Chip 1 in `EXT=2` (AB), in RAW fullframe unbinned size of 2068 x 4144. *Storage wise, wouldn't it be more efficient to store this info as a binary table? Ah, well.*
* Test SNKCFILE was provided by Jenna Ryon.

Questions:

- [x] Why are there no DQ extensions in reference image? It is much easier for INS to calculate however they want to populate the DQ and just deliver the DQ values to populate. It is inefficient for CALACS to recalculate the same thing over and over again based on SNKCFILE's SCI data for every input image. The ISR talks about flagging based on background level but how is CALACS supposed to estimate background blindly? **Need science image pixel value to calculate (not global background). Exact flags will be different for each images.**
- [x] Will *all* WFC images have these new keywords? That is, full reprocessing of static archive. It is mentioned that there "will not be a SNKCFILE before January 2015," but does that mean no keywords or they are set to `OMIT` and `N/A`? **Yes.**
- [x] Where exactly in CALACS is this correction step going? **After BLEVCORR, before PCTECORR, which puts it as the last step in ACSCCD.**
- [x] If it is set to `PERFORM` but `BIASCORR!=COMPLETE`, skip silently, error, or warning? **Just warning.**
- [x] Extract subarray region for subarrays? It is implied but I want to make sure. **Yes.**
- [x] Which pieces of CALWF3 do you want to copy over? [wf3ccd/sink.c](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/wf3ccd/sink.c)

Todo:

- [x] Need to implement a new step for this.
- [x] Need Jenna's Python flagging script for reference.
- [x] Need test images and SNKCFILE for regression test. Maybe a fullframe and a subarray. That includes science input, reference file, and expected output.

**Update from Norman (2017-02-14)**

This has nothing to do with the new CTE code. The intent is to clone the CALWF3 sinkcorr stage into CALACS, with minor modification due to our change in reference file conventions.

We will generate a new sinkcorr reference file every anneal cycle (but only in the flashed-darks era), with appropriately indexed USEAFTER. We also change the head-pixel labeling convention, to reflect the depth of the sink rather than its birthday. We may use a different DQ bit for WFC sinkpix than is used by CALWF3: Jenna and/or David can confirm.

Jenna should be able to walk you through the modifications to the CALWF3 sinkcorr code (perhaps
with a little assistance from David)?  She should be able to hand you a prototype WFC sinkcorr reference file immediately.

**Related Links**

* [Sink Pixels in ACS/WFC (draft)](https://stsci.app.box.com/s/2226ipgtsivsx3qifh3ewh0ka0zmotec) by Ryon & Grogin 2017, Section 6
* [Problem Report 86721](https://www.ess.stsci.edu/prsystem/servlet/prbrowse/pr.86721) (New Sink Pixel Flagging Reference File)
* [Problem Report 86722](https://www.ess.stsci.edu/prsystem/servlet/prbrowse/pr.86722) (New Keyword for ACS/WFC Sink Pixel Flagging)
* [New Reference File Type for ACS/WFC Sink Pixel Flagging](https://jira.stsci.edu/browse/CRDS-92)

pllim:
**Approved by Jenna Ryon from ACS Team.**

Fixes #44 

**Instructions to install this branch for testing**

1. Create a directory to store the source codes and `cd` into it.

2. `git clone https://github.com/pllim/hstcal.git ./` (this makes a copy of my HSTCAL fork on your disk)

3. `git fetch origin new-sinkcorr`

4. `git checkout origin/new-sinkcorr -b new-sinkcorr` (if this does not work, your `git` version is too old, so you will need to update `git` first or install a newer version using `conda install git`)

5. `git status` (if this shows that you are now in `new-sinkcorr` branch, then proceed)

6. `./waf configure --debug --prefix=$CONDA_PREFIX` (this only works if you use `conda env`)

7. `./waf build` (make sure there is no error)

8. `./waf install`

9. Now `cd` into your working directory with data files.

10. `calacs.e --version` (only proceed if this says version 9.1.0 and is dated Feb 21, 2017)

11. `calacs.e your_file_name -v` (`-v` gives verbose output in trailer file and screen, which is useful for debugging)

**Instructions to grab updates from existing branch checkout**

1. `cd` into source checkout directory.

2. `git checkout new-sinkcorr` to switch to this branch, if not already on it (you can check with `git status`)

3. `git fetch origin new-sinkcorr` (picks up updates)

4. `git rebase origin/new-sinkcorr` (applies the updates locally)

5. Continue from Step 8 above.

**Todo after merge**

- [ ] Add WFC test data (fullframe and subarray) from Jenna Ryon to regression test suite.
- [x] All the other existing input RAW files for WFC and HRC need to have `SINKCORR = OMIT` in their primary headers.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .waf*
 bin.*
 build.*
+Makefile
--- a/pkg/acs/calacs/Dates
+++ b/pkg/acs/calacs/Dates
@@ -1,3 +1,4 @@
+ 21-Feb-2017:  CALACS 9.1.0 New SINKCORR step added in ACSCCD for WFC.
  22-Nov-2016:  CALACS 9.0.0 BLEVCORR now uses new OSCNTAB that correctly
                             process all subarrays.
  21-Oct-2016:  CALACS 8.3.5 BLEVCORR can now process polarizer and ramp
--- a/pkg/acs/calacs/History
+++ b/pkg/acs/calacs/History
@@ -1,3 +1,6 @@
+### 21-Feb-2017 - PLL -- Version 9.1.0
+    New SINKCORR step is added to ACSCCD for WFC.
+
 ### 22-Nov-2016 - PLL -- Version 9.0.0
     BLEVCORR modified to correctly process all subarrays by using new OSCNTAB
     that correctly define overscans for all WFC and HRC apertures.
--- a/pkg/acs/calacs/Updates
+++ b/pkg/acs/calacs/Updates
@@ -1,3 +1,10 @@
+Update for version 9.1.0 - 21-Feb-17 (PLL)
+    acsccd/doccd.c
+    acsccd/dosink.c
+    acsccd/getacsflag.c
+    acsccd/getccdsw.c
+    include/acsinfo.h
+
 Update for version 9.0.0 - 22-Nov-16 (PLL)
     acsccd/findover.c
     acsccd/doblev.c
--- a/pkg/acs/calacs/acsccd/doccd.c
+++ b/pkg/acs/calacs/acsccd/doccd.c
@@ -12,6 +12,7 @@
  10-Dec-2001 WJH: Added update of SIZAXIS keywords to reflect trimmed size.
  11-Dec-2012 PLL: Moved FLSHCORR stuff to ACS2D.
  12-Aug-2012 PLL: Separated PCTECORR from ACSCCD.
+ 21-Feb-2017 PLL: Added SINKCORR.
 
 ** This code is a trimmed down version of CALSTIS1 do2d.c.
 */
@@ -29,6 +30,7 @@ static void dqiMsg (ACSInfo *, int);
 /*static void AtoDMsg (ACSInfo *, int);*/ /* Not used */
 static void BiasMsg (ACSInfo *, int);
 static void BlevMsg (ACSInfo *, int);
+static void SinkMsg (ACSInfo *, int);
 
 
 int DoCCD (ACSInfo *acs_info) {
@@ -66,6 +68,7 @@ int DoCCD (ACSInfo *acs_info) {
     int dqiHistory (ACSInfo *, Hdr *);
     int doNoise (ACSInfo *, SingleGroup *, int *);
     int noiseHistory (Hdr *);
+    int SinkDetect (ACSInfo *, SingleGroup *);
     int GetACSGrp (ACSInfo *, Hdr *);
     int OmitStep (int);
     int PutKeyFlt (Hdr *, char *, float, char *);
@@ -420,6 +423,23 @@ int DoCCD (ACSInfo *acs_info) {
     }
     /************************************************************************/
 
+    /************************************************************************/
+    /* Flag sink pixels.                                                    */
+    SinkMsg (&acs[0], 1);
+
+    if (acs->sinkcorr == PERFORM) {
+        for (i = 0; i < acs_info->nimsets; i++) {
+            if (SinkDetect(&acs[i], &x[i]))
+                return (status);
+        }
+
+        PrSwitch("sinkcorr", COMPLETE);
+
+        if (acs_info->printtime)
+            TimeStamp("SINKCORR complete", acs_info->rootname);
+    }
+    /************************************************************************/
+
     /* Write this imset to the output file.  The
        CAL_VER and FILENAME keywords will be updated, and the primary
        header will be written.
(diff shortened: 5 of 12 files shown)


pllim:
@jryon, it turns out we do still need the untrimmed SNKCFILE reference files *but* with the following addition to their headers. Please update your files accordingly as follows:

```python
from astropy.io import fits

for filename in sinkfiles:
    with fits.open(filename, mode='update') as pf:
        pf[1].header['LTV1'] = 24.0
        pf[1].header['LTV2'] = 0.0
        pf[1].header['RAW_LTV1'] = 0.0
        pf[1].header['RAW_LTV2'] = 0.0
        pf[1].header['LTM1_1'] = 1.0
        pf[1].header['LTM2_2'] = 1.0
        pf[2].header['LTV1'] = 24.0
        pf[2].header['LTV2'] = 20.0
        pf[2].header['RAW_LTV1'] = 24.0
        pf[2].header['RAW_LTV2'] = 20.0
        pf[2].header['LTM1_1'] = 1.0
        pf[2].header['LTM2_2'] = 1.0
```

Also, using this code (see original issue post for installation instructions for testing), I find that flagged tails are shorter compared to your "truth" images. This is probably expected given that you flagged yours using final calibrated images with dark subtracted, etc, while CALACS flagged them right after BLEVCORR at the very end of ACSCCD step. So, the "background" is higher in the way that CALACS does it. Would be nice if you also re-flag your "truth" images independently as follows:

1. Run only `acsccd.e` on your RAW.
2. Use your Python script to flag `*blv_tmp.fits`.
3. Pass the flagged `*blv_tmp.fits` (don't change the filename) to `acs2d.e` to get the FLT. We can skip CTE correction here, as it is irrelevant to SINKCORR.

Thank you.

pllim:
@sosey , if you are so inclined, perhaps you want to have a look to see if this makes sense? Thanks.

jhunkeler:
@pllim `./waf configure --debug --prefix=/path/to/python` is worded dangerously, even with the note next to it. People might not read the note and nuke their python binary :)

It might be safer to tell the user to do this:
```
./waf configure --debug --prefix=/path/to/install/hstcal
./waf build
./waf install

#bash
export PATH=/path/to/install/hstcal/bin:$PATH

#csh
setenv PATH /path/to/install/hstcal/bin:$PATH

calacs.e --version
[...]
```

pllim:
This is really only intended for @jryon . Anyone else who wants to install this, not my problem. :stuck_out_tongue:

jhunkeler:
I feel like @jryon might be sad if her python _binary_ is replaced with a python _directory_ that contains hstcal 😀

pllim:
@jhunkeler , we can debate about this offline. I prefer to attach the installation to a conda env, so when testing is done, you can just nuke it.

pllim:
@jryon, I found a bug in my code and fixed it. There is now no difference between our outputs for the two test files. Please follow instructions above (with or without Joe's addendum), test this with more stuff, and let me know when to merge. Thank you for your patience!

jryon:
@pllim I've completed my testing and found no problems. You can merge. I tested this on a range of different filters, background levels, subarrays, and observation dates, and it works how it should. Thanks for your hard work!

pllim:
Thanks, @jryon !

pllim:
Just a rebase, no need to rerun Travis.

Issue #55, opened 2017-01-26
WFC3 CTE - incomplete conditional for subarrays in 2nd bias subtraction code


jamienoss:
[This](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/wf3cte/wf3cte.c#L630-L632) ``if`` hasn't realized that it's C and not python.

```
if(Pix(raz->dq.data,i+k*subcol,j))
     Pix(raz->sci.data,i+k*subcol,j) -= bias_pre[k];
     Pix(raz->sci.data,i+k*subcol,j) *= gain;
```

Missing braces.

I have not yet looked to see how this affects anything. This is only an issue for subarrays.

PR #69, opened 2017-04-02
wfc3 cte specific logic update for issue #55

sosey:
This updates the logic to be specific to the if loop it's contained in. The fix results in no data changes because the previous logic applied the gain correction to all pixels per the outside for loop.


jamienoss:
Fixes #55

PR #71, opened 2017-04-04
[EXP] Proof of concept of a new way to do regression test

### LINKED ISSUE #339: Retire Python 2
Python 2 will not be maintained past Jan 1, 2020 (see https://pythonclock.org/). Please remove all Python 2 compatibility and move this package to Python 3 only. 

For conda recipe (including `astroconda-contrib`), please include the following to prevent packaging it for Python 2 (https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html?preprocessing-selectors#skipping-builds): 

``` 
build: 
  skip: true  # [py2k] 
``` 

Please close this issue if it is irrelevant to your repository. This is an automated issue. *If this is opened in error, please let pllim know!*

pllim:
Proof of concept on how HSTCAL regression test is going to run on CI.

Supersede spacetelescope/acstools#11

Related: spacetelescope/braindump#11

Fix #339 

TODO:
- [x] Add option to use CRDS from central storage instead of always FTP.
- [x] Add option to run if input images are already on disk. Since they currently only exist on certain machines, which no human likes to use, I need some concrete use cases to figure out something logical. Maybe grab from some central storage? (Yes, that should work.)
- [x] Add tests to test the test helpers/utilities.
- [x] Make a base class for tests to avoid code redundancy.
- [x] Add more ACS tests (but not all).
- [x]  Add WFC3 and STIS tests (maybe one each) to just make the point? Sure.
- [x] Fix the only remote failure possibly due to CGI bug.
- [x] Try again when Jenkins `XBuild + junit` failure is fixed upstream.
- [x] Drop Python 2.
- [x] Move all Linux build tests from Travis to Jenkins.
- [x] Use `ci-watson` from `spacetelescope/ci_watson`, not PR branch. Needs spacetelescope/ci_watson#3 to be merged first.


pllim:
Took over 30 minutes to run? Yikes...

tests/acs/test_wfc_asn1.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Run test in own dir so we can keep results separate from
+        other tests.
+        """
+        p = tmpdir.mkdir('wfc_asn1')
+        self.prevdir = os.getcwd()
+        os.chdir(p)
+
+    def teardown_class(self):
+        os.chdir(self.prevdir)
```
pllim: I don't understand why this one gives error:
```
________ ERROR at teardown of TestFullframePreSM4.test_fullframe_presm4 ________

self = <class 'tests.acs.test_wfc_asn1.TestFullframePreSM4'>

    def teardown_class(self):
        """Change back to test directory."""
>       os.chdir(self.prevdir)
E       AttributeError: type object 'TestFullframePreSM4' has no attribute 'prevdir'
```

Am I going crazy?

pllim: Nevermind. I figured it out.

jhunkeler:
Oh weird. I saw this yesterday but it didn't show the comment on the code so I thought your comment was "empty". 

I noticed some other issues too. I'll send potential revisions today

pllim:
Hold your horses, I am still working on this. Don't push anything here directly.

jhunkeler:
I didn't mean in-place... that would be rude. 👍 

```diff
diff --git a/tests/acs/test_wfc_asn1.py b/tests/acs/test_wfc_asn1.py
index daccdd3..5aceda5 100644
--- a/tests/acs/test_wfc_asn1.py
+++ b/tests/acs/test_wfc_asn1.py
@@ -19,7 +19,7 @@ class TestFullframePreSM4(object):
         Run test in own dir so we can keep results separate from
         other tests.
         """
-        p = tmpdir.mkdir('wfc_asn1')
+        p = str(tmpdir.mkdir('wfc_asn1'))
         self.prevdir = os.getcwd()
         os.chdir(p)

`tmpdir.mkdir` was returning a `LocalPath` object causing `os.mkdir` to throw a `TypeError`. If I cast it to `str` then it works.

```diff
diff --git a/tests/helpers.py b/tests/helpers.py
index 2b8bec8..9b58bf9 100644
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1,6 +1,6 @@
 """HSTCAL regression test helpers."""
+from distutils.spawn import find_executable
 import os
-import shutil
 import subprocess
 import sys

@@ -10,7 +10,6 @@ import requests
 __all__ = ['remote_data', 'use_calacs', 'use_calwf3', 'use_calstis',
            'HAS_CALXXX', 'download_file_cgi']

-PREFIX = '/tmp/hstcal'  # As set in .travis.yml
 HAS_CALXXX = {}   # Set by set_exe_marker()


@@ -29,26 +28,11 @@ def set_exe_marker(instrument):
         raise ValueError('{} is not supported'.format(instrument))

     HAS_CALXXX[instrument] = False
-    travis_path = os.path.join(PREFIX, calxxx)
+    cal = find_executable(calxxx)

     # This is for Travis CI installation check.
-    if os.path.isfile(travis_path):
-        HAS_CALXXX[instrument] = travis_path
-
-    # This is for local installation check for Python 3.3 or later.
-    elif sys.version_info >= (3, 3):
-        if os.path.isfile(shutil.which(calxxx)):
-            HAS_CALXXX[instrument] = calxxx
-
-    # This is for local installation check before Python 3.3.
-    # NOTE: This will fail on Windows!
-    else:
-        try:
-            subprocess.check_call(['which', calxxx])
-        except Exception:
-            pass
-        else:
-            HAS_CALXXX[instrument] = calxxx
+    if os.path.isfile(cal):
+        HAS_CALXXX[instrument] = cal
```

I think instead of defining an arbitrary prefix we should probably just `which` the `PATH` directly. If the executable isn't there then the tests just fail out. I don't know if the travis tests are putting the resulting binaries into `PATH` -- if not, then we definitely should.

`distutils.spawn.find_executable` works the same on all platforms and should be available under 2.7 and 3.x. Also there's a caveat with `shutil.which`. If it doesn't find anything it returns `None`, so the later call to `os.path.isfile` would ultimately throw a `NoneType` exception.

pllim:
While you are at in, I am currently battling with `requests` refuses to recognize FTP connection. Any idea?

jhunkeler:
Sure, do you have a sample URL I can test against?

pllim:
```python
url = "ftp://ftp.stsci.edu/cdbs/jref/o151506fj_ccd.fits"
```
`requests.get(url)` gives invalid scheme error.

jhunkeler:
`requests` doesn't support FTP, so we could install requests_ftp (via pip) and use that. Or just switch to `urllib.urlretrieve` or `.urlopen` for FTP files.

sosey:
side info: url may need to be secure - I also ran into an issue with requests a couple weeks ago that the conda included package wasn't compiled against the SSL module is needed and I had to switch to curl.

pllim:
Yes, `requests` does not support FTP. And `requests-ftp` is of questionable quality and no longer being maintained (I just uninstalled it).

pllim:
@jhunkeler -- I don't know whether to :joy: or :sob: that with a single comment, you cut my function down by 3X. Also, no need for `str()`, `tmpdir` has `strpath` attribute.

tests/helpers.py
```diff
@@ -0,0 +1,174 @@
...
+
+    s = refname.split(sep)
+    filename = s[1]
+
+    # CRDS file for given name never changes, so no need to re-download.
+    if os.path.exists(filename):
+        return
+
+    url = 'ftp://ftp.stsci.edu/cdbs/{}/{}'.format(s[0], filename)
+    _download_file(url, filename, timeout=timeout)
```
pllim: Note: This currently does not work. Investigating.

pllim:
@jhunkeler , `distutils` is giving `NoneType` error on Travis CI, just as you suspected. How do we make Travis CI successfully find the executables, because it is not finding it with default config.

jhunkeler:
The FTP download (from home) worked perfectly fine. Maybe there's something up on-site that's causing the download issue?

```
IMPHTTAB jref$w3m17171j_imp.fits
PHOTCORR COMPLETE
Imset 2  End 21:49:44 EDT


End      06-Apr-2017 21:49:44 EDT


*** ACS2D complete ***
Finished CCD processing...
CALACS: Building DTH products
Warning    No DTH product name specified. No product created.
CALACS: Finished processing product 


End      06-Apr-2017 21:49:44 EDT


*** CALACS complete ***
CALACS completion for j6lq01010_asn.fits
```
So you can see that the first set finished no problem after a few minutes, but it failed out looking for `jref` anyway. I guess in `remote-data` mode we should ignore `jref` completely, or set it to to the path where pytest stored the file(s) from CRDS.

```python
.E

==================================== ERRORS ====================================
________ ERROR at teardown of TestFullframePreSM4.test_fullframe_presm4 ________

self = <class 'tests.acs.test_wfc_asn1.TestFullframePreSM4'>

    def teardown_class(self):
        """Change back to test directory."""
        os.chdir(self.prevdir)
>       os.environ['jref'] = self.prevjref

tests/acs/test_wfc_asn1.py:44: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../miniconda3/envs/hstcal/lib/python3.5/os.py:730: in __setitem__
    value = self.encodevalue(value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

value = None

    def encode(value):
        if not isinstance(value, str):
>           raise TypeError("str expected, not %s" % type(value).__name__)
E           TypeError: str expected, not NoneType

../../miniconda3/envs/hstcal/lib/python3.5/os.py:798: TypeError
===================== 1 passed, 1 error in 343.58 seconds ======================
```

As far as putting the binaries in the `PATH`, we should just change Travis config from `--prefix=/tmp/hstcal` to `--prefix=/tmp/miniconda3`.

I tested that idea locally and that's how I got the tests to kick off (and fail differently above).

pllim:
The teardown error is  because you don't have a `jref` set already on your side. I'll add logic to handle that. Ignoring `jref` completely would defeat the purpose of the test because that is how HSTCAL works for everyone else, so I disagree. I'll also try your prefix. Thanks!

pllim:
@jhunkeler , in regards to any performance improvement on paid version, I am not seeing it. The test here was put in queue, I guess because https://travis-ci.org/spacetelescope/specviz/builds/219700559 was running (which I then cancelled because of something unrelated here). We need some clarifications from the vendor.

pllim:
So there are actually 2 passes this time!
https://travis-ci.org/spacetelescope/hstcal/jobs/219706661
https://travis-ci.org/spacetelescope/hstcal/jobs/219706662

But then it fails (and I canceled the rest):
https://travis-ci.org/spacetelescope/hstcal/jobs/219706665

The failure is due to some missing file but I can't tell which one for sure (Travis does not give the full CALACS log, unlike when I run it locally). Basically CALACS failed at some point and did not product the output files. Maybe FTP site starts to block repeated access as a security measure?

But this raises another question: Do we really need to run CALACS for all the `CFLAGS` variants? They were only added to make sure HSTCAL compiles, right? In that case, we can turn off "remote data" for those. What do you think?

jhunkeler:
Add the `-s` argument to pytest to enable stdout/stderr streams. That's what I've been doing locally to make sure things are working.

```yaml
script:
  - pytest -s tests --remote-data
```

The `CFLAGS` and `LDFLAGS` make sure the program can build under different situations, but this primarily effects OS X. HSTCAL can't be compiled without messing with those settings. I got a little push back when I tried to add `xxFLAGS` auto-detection for cross-platform builds directly into the `wscript` -- so I abandoned the prospect.

jhunkeler:
Bah...

```
----------------------------- Captured stderr call -----------------------------
dyld: Library not loaded: @rpath/libgomp.1.dylib
  Referenced from: /tmp/miniconda3/bin/calacs.e
  Reason: image not found
```

conda-build fixes the RPATHs using _magic_ so a basic build using conda's `gcc` isn't good enough anymore. This used to work in the past... I'll see if I can force it with `LDFLAGS="-m64 -Wl,-rpath=/tmp/miniconda3/lib` or something.

If this doesn't work we might just want to strip gcc from the conda deps and rely on the non-openmp build using clang on OS X. We'll see.

pllim:
HSTCAL does not depend on Python version, so instead of testing Linux build twice with remote data for no reason, I now changed it to test it once on Linux and once on Mac.

@jhunkeler , feel free to open a PR against my PR branch (does that make sense?) with your fix for `gcc`.

jhunkeler:
Unless we change the build matrix to compensate, won't it always kick off `pytest` anyway and fail out the job? Or did I miss something?

pllim:
@jhunkeler , it depends. When remote data is turned off, this CALACS test will be skipped. So as long as HSTCAL is compiled successfully, it will pass, because there is no real test to run.

jhunkeler:
Aha yeah I just noticed that in the log output. Nice!

pllim:
Test passed! More work to be done but this is promising. Also, it look 5 minutes on Linux build but 25 on Mac OSX; I wonder why.

jhunkeler:
Travis' OSX images take _forever_ to do _anything_.  Sometimes PRs that finishes on a Linux container in seconds can end up taking hours simply because its waiting for OSX to boot up.

pllim:
So, you don't think using clang build for remote data would be faster? You mentioned something about forcing rpath would use GCC without openmp, which would result in CTE correction taking longer, or did I misunderstand​?

jhunkeler:
I think I'll run a few PoC benchmarks so we can see the difference in CPU and IO speed between Linux and OSX jobs. I've had some pretty bad (personal) experiences with virtualized OSX...

Anyway let's see if switching the remote-data flag over to a clang job improves the situation. It could be the openmp implementation just doesn't work very efficiently under OSX.

jhunkeler:
https://travis-ci.org/jhunkeler/shebeshebe

There's a major performance difference between Linux and OSX. The test script is very basic and isn't 100% accurate (the values will fluctuate based on system load) however it's a good indication of what we are working with behind the scenes. 

OS X is as worthless as I suspected.

pllim:
Wow, it times out on Mac OSX `clang`. The log is empty, so it didn't even get to run the test? I wonder why...

jhunkeler:
I restarted that single job just to make sure. I noticed earlier it said "error" and had no log whatsoever while it was "running".

jhunkeler:
I guess this answers the question of whether we're using the paid travis instances or not.

```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received

The build has been terminated
```

pllim:
I guess switching back to `gcc` is better than `clang` then. Will do that and squash the commits when I get the chance. Thanks for checking!

jhunkeler:
Come to think of it.... Hmm... The gcc+openmp job we ran the other day took about 2000 seconds to complete. That's much longer than 10 minutes. I guess that implies `pytest -s` circumvents the idle restriction, because it's blasting useless data to stdout the whole time?

Travis doesn't make sense sometimes.

pllim:
Yes, I think you are right about "-s". Mac OSX "gcc" build also times out when I took out the "-s" option. This is unfortunate, because "-s" is only useful when there is something really wrong. Otherwise, a test summary would be better because all the screen output drowns out actual `fitsdiff` reports.

pllim:
I would not use "travis_wait" for now, since "-s" works better with less guess work. Also does not look like `clang` build is any faster, so switching back to `gcc`.

pllim:
Fun fact: Time to build the Mac version (without remote data) is longer than the *entire* test on Linux. Maybe I just should disable remote data testing on Mac on Travis. People can still test it locally on a Mac if they are so inclined.

pllim:
Sample tests for ACS (all 3 detectors), WFC3/IR, and STIS FUV-MAMA all passed. :muscle: :muscle: :muscle:

stsci-hack:
The helpers module should eventually be made into its own package that could be called by any other package that wants to implement this interface for testing.  Otherwise, I like the current implementation and support having this code merged into HSTCAL.

pllim:
> The helpers module should eventually be made into its own package

That has been a long standing discussion on the Astropy side. For instance, "--remote-data" was copied from Astropy test helper. When Astropy helper is its own separate package, you request is naturally addressed. There is even a bid for a summer student to do this at [OpenAstronomy GSoC](http://openastronomy.org/gsoc/gsoc2017/) this year.

pllim:
@sosey 👍 or 👎 on merge?

pllim:
@jhunkeler @rendinam -- https://travis-ci.org/spacetelescope/hstcal/jobs/274737453 shows green, so theoretically this would pass in Jenkins now if things are set up right there. Feel free to take this updated feature branch for a spin. FYI.

rendinam:
Protected branch status checks should be available to provide validation against Travis, Jenkins, and/or other available status values.
https://help.github.com/articles/enabling-required-status-checks/

pllim:
Ideas are still floating around between @jhunkeler and I. I'll just write them down here so I don't forget:

* Leave the "build only" matrix on Travis (those don't require any data access).
* Mark most of the data intensive tests as "slow" (`--remote-data` and `--slow`); These will only run nightly on dev (and stable?) as a special Jenkins pipeline so they don't over-burden the test machine. Developers still can run these manually locally on their own branches as needed.
* Leave, say 10-15, data intensive tests as "normal" (`--remote-data`); These get run as part of Jenkins CI for every commit push to this repository.

jamienoss:
@pllim 
>Leave the "build only" matrix on Travis (those don't require any data access).

I would vote strongly against this. Travis is sooooo slow that even these are a workflow burden.

pllim:
> Travis is sooooo slow

Only the Mac builds are slow but you shut down #257 and Jenkins don't have Mac builds right now, so I don't know what is the solution here.

pllim:
Note: `[ci skip]` skips Travis build, as intended, but has no effect on Jenkins.

jamienoss:
@pllim is there somewhere that outlines and describes the test and running frameworks 
 that this PR implements?

pllim:
This is still very experimental, so no comprehensive documentation yet. The `Jenkinsfile` should give a rough idea but that is about it right now.

jamienoss:
@pllim 
> This is still very experimental, so no comprehensive documentation yet. The Jenkinsfile should give a rough idea but that is about it right now.

Is this expected to be merged in it's "experimental" state or properly architected once you've done with the (un-merged) POC? Is there really nowhere where you've outlined the design before implementing it? In addition, are you saying that it will be completely re-written once properly designed, based on architected frameworks, to facilitate the final version?

pllim:
No, please don't merge this yet. My question back on Apr 18 is no longer valid. When it is ready, I'll remove `[EXP]`.

jamienoss:
@pllim 
> No, please don't merge this yet. My question back on Apr 18 is no longer valid. When it is ready, I'll remove [EXP].

I'm not sure if this is directed at me. It would not be my responsibility to merge this, especially in its current state or without explicitly asking so, which I was not doing BTW. My [comment above](https://github.com/spacetelescope/hstcal/pull/71#issuecomment-345817457) was questioning your intent and plan regarding this, which remains largely unanswered.

pllim:
Okay. I can answer your questions offline some time today.

Jenkinsfile
```diff
@@ -0,0 +1,68 @@
...
+                        if (name == 'release') {
+                            sh 'conda install -q -y pytest requests astropy'
+                            sh 'pip install -q pytest-remotedata'
+                        }
+                        withEnv(runtime) {
+                            sh "${test_cmd}"
+                        }
+                    }
+                }
+                catch (e) {
```
jhunkeler: I know this was based on what was in my own Jenkinsfile, but I just wanted to make sure you know:
`catch (e) {` _only_ records test information in the event of a failure. If you replace with this with `finally {` the test results will be recorded no matter what happens.

pllim: The report as-is is good enough for me. And in previous attempt with failure, I found that info reported as sufficient for tracking where the failure came from as well (at least for that case). Thank you!

```
+ pytest tests --basetemp=tests_output --junitxml results.xml --remote-data
============================= test session starts ==============================
platform linux -- Python 3.6.1, pytest-3.2.1, py-1.4.34, pluggy-0.4.0

rootdir: ..., inifile:
plugins: remotedata-0.2.0
collected 15 items

tests/acs/test_hrc.py .
tests/acs/test_sbc.py s.
tests/acs/test_wfc.py .
tests/stis/test_lev1_fuvspec.py x
tests/utils/test_helpers.py ........s
tests/wfc3/test_ir_01.py .

 generated xml file: .../results.xml 

============== 12 passed, 2 skipped, 1 xfailed in 239.93 seconds ===============
```

jhunkeler: Sounds good. To take full advantage of the analyzer's pass/fail/diff/newtest/oldtest graphing, drill downs, and "weather report" functionality you'd probably need to keep a steady stream of results. Otherwise you'd only be diffing failures over time.

pllim: > diffing failures over time

Ah, that would be bad. I'll change it to `finally` as you suggested. Thank you!

tests/helpers.py
```diff
@@ -0,0 +1,389 @@
...
+
+
+def ref_from_image(input_image):
+    """
+    Return a list of reference filenames, as defined in the primary
+    header of the given input image, necessary for calibration; i.e.,
+    only those associated with ``*CORR`` set to ``PERFORM`` will be
+    considered.
+    """
+    # NOTE: Add additional mapping as needed.
```
stscicrawford: Can this be replaced by https://github.com/spacetelescope/crds/blob/master/crds/bestrefs/bestrefs.py?

pllim: This one in helpers is simply to make sure the necessary *keywords* are present. It does not impose any *value* requirements.

Whether to use latest CRDS *value* mapping or not is a separate discussion. Ideally, we need to be able to tell when there is a diff, whether it is caused by code change or CRDS change (we might or might not care about the latter).

pllim:
Note to self: pllim/hstcal#2

tests/helpers.py
```diff
@@ -0,0 +1,389 @@
...
+            raise AssertionError(os.linesep + creature_report)
+
+        return creature_report
+
+
+@use_calacs
+class BaseACS(BaseCal):
+    refstr = 'jref'
+    prevref = os.environ.get(refstr)
+    input_loc = 'hstcal/acs/calacs_e'
```
pllim: @mdlpstsci , in the current state of the POC, input files are assumed to be in a pre-defined location within these sub-folder, so to change the input data, you need to copy the new versions here.

Just below this line is where the "truth" folder is defined. These are following pandokia file system layout but we certainly don't have to do it this way going forward.

Does this answer you question about how to change the data files for a test?

tests/helpers.py
```diff
@@ -0,0 +1,389 @@
...
+    allow_remote_ref : bool
+        If set to `True`, instead of copying or downloading to current
+        directory, return the full path. This should only be used for
+        "truth" images.
+
+    """
+    # NOTE: This could be explicitly controlled using pytest fixture
+    #       but too many ways to do the same thing would be confusing.
+    #       Refine this logic if using pytest fixture.
+    cs_file = os.path.join('/eng/ssb2/tests', tree, project, filename)
```
pllim: @mdlpstsci , the "prepend" part is here. However, if you change this for one test, you change it for all the test. You currently cannot change this in a per-test basis.

pllim:
Update: I have received some files from @mdlpstsci (thanks, Michele!). I just need to find some zen time to get this thing back on track. Next week hopefully. Thanks for your patience!

jhunkeler:
@pllim The broken pipe is from the `yes` pipe falling through after `waf configure` failed to find `cfitsio`. Disregard.

Rebasing against `master` should fix this. Or just replace your branch's `Jenkinsfile` with the contents of the one on `master`.

jhunkeler:
```diff
$ git --no-pager diff lim/regtest-poc -- Jenkinsfile
diff --git a/Jenkinsfile b/Jenkinsfile
index 3d7f4471..8cbaa171 100644
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,8 +2,8 @@
 if (utils.scm_checkout()) return
 
 // Config data to share between builds.
-CFLAGS = 'CFLAGS="-m64"'
-LDFLAGS = 'LDFLAGS="-m64"'
+CFLAGS = ''
+LDFLAGS = ''
 DEFAULT_FLAGS = "${CFLAGS} ${LDFLAGS}"
 // Some waf flags cause a prompt for input during configuration, hence the 'yes'.
 configure_cmd = "yes '' | ./waf configure --prefix=./_install ${DEFAULT_FLAGS}"
@@ -12,27 +12,28 @@ configure_cmd = "yes '' | ./waf configure --prefix=./_install ${DEFAULT_FLAGS}"
 // Define each build configuration, copying and overriding values as necessary.
 bc0 = new BuildConfig()
 bc0.nodetype = "linux-stable"
-bc0.name = "debug"
+bc0.build_mode = "debug"
 bc0.env_vars = ['PATH=./_install/bin:$PATH']
-bc0.build_cmds = ["${configure_cmd} --debug",
+bc0.build_cmds = ["conda config --add channels http://ssb.stsci.edu/astroconda",
+                  "conda install -q -y cfitsio pkg-config pytest requests astropy",
+                  "pip install -q pytest-remotedata",
+                  "${configure_cmd} --debug",
                   "./waf build",
                   "./waf install",
                   "calacs.e --version"]
 
 
 bc1 = utils.copy(bc0)
-bc1.name = "release"
-bc1.build_cmds[0] = "${configure_cmd} --release-with-symbols"
-bc1.test_cmds = ["conda install -q -y pytest requests astropy",
-                 "pip install -q pytest-remotedata",
-                 "pytest tests --basetemp=tests_output --junitxml results.xml --remote-data"]
+bc1.build_mode = "release"
+bc1.build_cmds[3] = "${configure_cmd} --release-with-symbols"
+bc1.test_cmds = ["pytest tests --basetemp=tests_output --junitxml results.xml --remote-data -v"]
 bc1.failedUnstableThresh = 1
 bc1.failedFailureThresh = 6
 
 
 bc2 = utils.copy(bc0)
-bc2.name = "optimized"
-bc2.build_cmds[0] = "${configure_cmd} --O3"
+bc2.build_mode = "optimized"
+bc2.build_cmds[3] = "${configure_cmd} --O3"
 
 
 // Iterate over configurations that define the (distibuted) build matrix.
```

pllim:
`13 failed, 8 passed, 2 skipped, 1 xfailed in 8.89 seconds`

OK this I can live with. Thanks, Joe!

pllim:
Hmm... Any idea how to fix this, @jhunkeler ?

```
+ pip install git+git://github.com/spacetelescope/ci_watson.git@master#egg=ci-watson
Collecting ci-watson from git+git://github.com/spacetelescope/ci_watson.git@master#egg=ci-watson
  Cloning git://github.com/spacetelescope/ci_watson.git (to master) to /tmp/pip-build-rofpyhv8/ci-watson
fatal: unable to connect to github.com:
github.com[0: 192.30.253.112]: errno=Connection timed out
github.com[1: 192.30.253.113]: errno=Connection timed out

Command "git clone -q git://github.com/spacetelescope/ci_watson.git /tmp/pip-build-rofpyhv8/ci-watson" failed with error code 128 in None

You are using pip version 9.0.1, however version 18.0 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

script returned exit code 1
```

pllim:
**Notes**

Did a test with `--slow` option per push event "for fun". This POC (which is a small subset of the actual nightly tests) took about 18 minutes. It took 6 minutes without `--slow` but with one (and only one) fullframe test using `PCTECORR=PERFORM`. And it took less than 4 minutes without any `PCTECORR`.

Attempted to run `--slow` on my own machine but it ran out of disk space in `/tmp`.

jhunkeler:
Gotta love those worthlessly tiny `/tmp`s... Try `pytest --basetemp=/somewhere/not/tmp` or `mkdir -p /somewhere/not/tmp && export TMPDIR=/somewhere/not/tmp` and run pytest like usual.

tests/helpers.py
pllim: @stscicrawford , let's table this one for "future work" (if necessary). Current `/grp/hst/cdbs` look-up works fine and does not introduce extra `crds` package dependency.

tests/stis/test_lev1_fuvspec.py
```diff
@@ -0,0 +1,26 @@
...
+        self.get_input_file(raw_file)
+        get_bigdata(self.env, 'stis', 'fuv-mama', 'input', wav_file)
+
+        # Run CALSTIS (equivalent to stistools.calstis.calstis)
+        subprocess.call(['cs0.e', raw_file, '-v'])
+
+        # Compare results
+        outputs = [('o5cl02040_flt.fits', 'calstis_lev1_FUVspec_flt.fits'),
+                   ('o5cl02040_x1d.fits', 'calstis_lev1_FUVspec_x1d.fits')]
+        self.compare_outputs(outputs)
```
pllim: Note: This is a weird one, where if I run this same test on my own machine, the comparison would fail until I set `rtol=0.01` (1%). 🤷‍♀️  But it passes if run on Jenkins. Investigating this is out of scope of this PR.

tests/wfc3/test_ir_01.py
```diff
@@ -0,0 +1,31 @@
...
+        asn_file = 'iabg31030_asn.fits'
+
+        # Prepare input file.
+        self.get_input_file(asn_file)
+
+        # Run CALWF3
+        subprocess.call(['calwf3.e', asn_file, '-v'])
+
+        # Compare results
+        # Excluded ('iabg31031_spt.fits', 'iabg31031_spt_ref.fits')
```
pllim: Note: There is a "truth" for SPT but local run does not generate SPT. Had an old note to ask @sosey but resolving this is also out of scope of this PR.

jhunkeler:
> Joe does not want me to kill all cores

No he does not. 😁

tests/stis/test_lev1_fuvspec.py
pllim: Attached is the failure I see locally, which Jenkins does not reproduce.

[stis_local_run.txt](https://github.com/spacetelescope/hstcal/files/2303410/stis_local_run.txt)

pllim:
As discussed in meeting today, someone please merge when tests pass. :pray: :pray: :pray:

pllim:
p.s. "Squash and merge" is probably fine if you want a cleaner history in "master". Or just regular merge is also fine if you prefer individual authors get proper credit (and blame).

jhunkeler:
@stscicrawford @mdlpstsci Merging this so we can start porting the remaining tests.

pllim:
Only took 1 year and 4 months!

PR #82, opened 2017-04-12
Add pointer register for memory allocation book keeping

### LINKED ISSUE #81: Add pointer register for easy book keeping of allocated memory
Since hstcal is not written in C++, error handling is cumbersome. Exiting at some point during the code, after various memory has been allocated, requires manually keeping track (book keeping) of already allocated memory so as to free said memory upon return. From what I have seen in calacs and calwf3, this manual booking isn't even being implemented. What I see is that normally only one allocated pointer being freed per return with little regard to anything else also previously allocated that is, perhaps, out of the developers 'focus'.

For the cte work I have implemented a simple register struct with functions to make this significantly easier for us to do.

NOTE: the existing PR doesn't not use this register, it simply adds it so that it can be used. This is effectively me splitting out the cte work into more logical commits for git hygiene.

jamienoss:
Adds the struct:
PtrRegister

and the folowing 'methods'
void initPtrRegister(PtrRegister * reg);
void addPtr(PtrRegister * reg, void * ptr, void * freeFunc); // ptr list is self expanding
void freePtr(PtrRegister * reg, void * ptr);
void freeOnExit(PtrRegister * reg); //only calls freeAll() followed by freeReg()
void freeAll(PtrRegister * reg); // frees all ptrs registered (excluding itself)
void freeReg(PtrRegister * reg); // frees itself i.e. the register array holding the ptrs

Fixes #81

### DIFF
--- a/hstio/hstio.c
+++ b/hstio/hstio.c
@@ -101,7 +101,6 @@
 ** Section 6.
 **      Functions to manipulate the header array.
 */
-# include "hstio.h"
 # include <fitsio.h>
 # include <ctype.h>
 # include <stdio.h>
@@ -110,12 +109,125 @@
 # include <sys/stat.h>
 # include <time.h>
 # include <unistd.h>
+# include <assert.h>
+#include <stdlib.h>
+
+# include "hstio.h"
 
 /*
 ** String defined to allow determination of the HSTIO library version
 ** from the library file (*.a) or the executable using the library.
 */
 const char *hstio_version = HSTIO_VERSION;
+
+void initPtrRegister(PtrRegister * reg)
+{
+    reg->cursor = 0; //points to last ptr NOT next slot
+    reg->length = PTR_REGISTER_LENGTH_INC;
+    reg->ptrs = calloc(reg->length+1, sizeof(*reg->ptrs));
+    assert(reg->ptrs);
+    reg->freeFunctions = calloc(reg->length+1, sizeof(*reg->freeFunctions));
+    if (!reg->freeFunctions)
+    {
+        free(reg->ptrs);
+        assert(0);
+    }
+    reg->ptrs[0] = reg; //this ptr
+    reg->freeFunctions[0] = &free;
+}
+void addPtr(PtrRegister * reg, void * ptr, void * freeFunc)
+{
+    if (!reg || !ptr || !freeFunc)
+        return;
+
+    //check ptr isn't already registered? - go on then.
+    {int i;
+    for (i = reg->cursor; i >= 0 ; --i)// i >= 0 prevents adding self again
+    {
+        if (reg->ptrs[i] == ptr)
+            return;
+    }}
+
+    if (++reg->cursor >= reg->length)
+    {
+        reg->length += PTR_REGISTER_LENGTH_INC;
+        assert(realloc(&reg->ptrs, reg->length*sizeof(*reg->ptrs)));
+        assert(realloc(reg->freeFunctions, reg->length*sizeof(*reg->freeFunctions)));
+    }
+    reg->ptrs[reg->cursor] = ptr;
+    reg->freeFunctions[reg->cursor] = freeFunc;
+}
+void freePtr(PtrRegister * reg, void * ptr)
+{
+    //Can't be used to free itself, use freeReg(), use of i > 0 in below for is reason.
+    if (!reg || !ptr)
+        return;
+
+    int i;
+    for (i = reg->cursor; i > 0 ; --i)
+    {
+        if (reg->ptrs[i] == ptr)
+            break;
+    }
+
+    //call function to free ptr
+    reg->freeFunctions[i](ptr);
+
+    if (i == reg->cursor)
+    {
+        reg->ptrs[i] = NULL;
+        reg->freeFunctions[i] = NULL;
+    }
+    else
+    {
+        //move last one into gap to close - not a stack so who cares
+        reg->ptrs[i] = reg->ptrs[reg->cursor];
+        reg->ptrs[reg->cursor] = NULL;
+        reg->freeFunctions[i] = reg->freeFunctions[reg->cursor];
+        reg->freeFunctions[reg->cursor] = NULL;
+    }
+    --reg->cursor;
+}
+void freeAll(PtrRegister * reg)
+{
+    if (!reg || reg->length == 0 || reg->cursor == 0)
+        return;
+
+    {unsigned i;
+    for (i = 1; i <= reg->cursor; ++i)
+    {
+        if (reg->freeFunctions[i] && reg->ptrs[i])
+        {
+            reg->freeFunctions[i](reg->ptrs[i]);
+            reg->ptrs[i] = NULL;
+            reg->freeFunctions[i] = NULL;
+        }
+    }}
+    reg->cursor = 0;
+}
+void freeReg(PtrRegister * reg)
+{
+    if (!reg || reg->length == 0)
+        return;
+
+    if (reg->cursor > 0)
+        freeAll(reg);
+
+    reg->cursor = 0;
+    reg->length = 0;
+    // free 'itself'
+    reg->freeFunctions[0](reg->ptrs);
+    reg->ptrs[0] = NULL;
+    reg->freeFunctions[0](reg->freeFunctions);
+    reg->freeFunctions[0] = NULL;
+}
+void freeOnExit(PtrRegister * reg)
+{
+    //Free everything registered
+    freeAll(reg);
+    //Free itself
+    freeReg(reg);
+}
 
 /*
 ** Section 1.
--- a/include/hstio.h
+++ b/include/hstio.h
@@ -119,6 +119,45 @@ extern "C" {
 
 # include <stdlib.h>
 
+/* Below is a general use pointer register (book keeping) to allow easy cleanup upon exit (pseudo C++ destructor)
+ * Use example:
+ *
+ * PtrRegister ptrReg;
+ * PtrRegister initPtrRegister(&ptrReg); // must always initialize before further use
+ *
+ * void * array = malloc(size);
+ * addPtr(&ptrReg, array, &free); // returns if array == NULL so don't need to pre check - this is also self expanding
+ *
+ * Bool cancel = False;
+ * ...
+ * do something
+ * cancel = True; //uh oh, something went wrong
+ * ...
+ * if (cancel)
+ * {
+ *     freeOnExit(&ptrReg); // This frees itself, i.e. the register array holding the ptrs
+ *     return;
+ * }
+ * ...
+ *
+ *NOTE: This pattern is considered integral to all use and as such internal failed allocations are asserted
+ */
+
+typedef void (*FreeFunction)(void*); // Only trivial functions accepted
+#define PTR_REGISTER_LENGTH_INC 10
+typedef struct {
+    unsigned cursor;
+    unsigned length;
+    void ** ptrs;
+    FreeFunction * freeFunctions;
+} PtrRegister;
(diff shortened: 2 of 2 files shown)


jamienoss:
@sosey, @stsci-hack, @jhunkeler, @pllim please may someone review this.

I wasn't entirely sure where it should live, hstcal doesn't seem to have a 'systems' ilb or something equivalent for infrastructure code. From what I could see, regarding the majority of allocation routines living in hstio.c, this would be the best guess, It is also pretty much central to all.

sosey:
There are some higher libraries that all the instruments have pulled from, in addition to hstio, for shared things that aren't in the specific instrument libs.  This is a big change to hstio, if none of the current code is referencing it, we should probably hold off on accepting this pull request until after 2017.2 delivery so that more extensive testing can be done. @philhodge knows the most about hstio at this point and should look at this change.

jamienoss:
[Example description](https://github.com/jamienoss/hstcal/blob/e1860ac77799702b6f7d5672644f166312e93f7f/include/hstio.h#L122-L144)

jamienoss:
@sosey This will be going in as part of 2017.2 as part of the CTE work. I have fractioned it out for commit/git correctness.

sosey:
you're updating HSTIO with this code, or is it self contained in ACS?

jamienoss:
@sosey #81 It is unused within this PR but is used within the acscte changes.

sosey:
In the past, code changes outside of calxxx which touched shared software needed to be validated by all teams. Since this is brand new code, which no other instruments are using it's probably fine, additionally this time (which is rare) all instruments are submitting changes and will need to verify their regression suite. In the future, these types of changes should be choreographed with all the teams, much earlier than this.

jamienoss:
@jhunkeler When you get a sec, could you take a look at the Travis fails, they look un code related. Just chatted with @pllim and she noted that it was using python 3.6 when it perhaps shouldn't be.

jamienoss:
@jhunkeler Actually I've just restarted it all...

pllim:
@jamienoss , please rebase to `master` to pick up the latest `waf`.

jamienoss:
@pllim copy that.

jamienoss:
Build restarted...

pllim:
Tests still failed but with more meaningful errors now. They look real.

hstio/hstio.c
```diff
@@ -102,13 +101,122 @@
...
+    reg->length = 0;
+}
+void freeReg(PtrRegister * reg)
+{
+    if (!reg || reg->length == 0)
+        return;
+
+    reg->cursor = 0;
+    reg->length = 0;
+    // free 'itself'
```
pllim: I vaguely remember that `//` is not supported. Can you change it to `/* */`?

jamienoss: No! It is supported by C99. Would you mind holding off from doing anything as I'm currently doing stuff. If you are planning on merging without a final notification to myself, I would advise not doing so. Please make a final inquiry to myself before attempting to merge, as we discussed a while back (git PR/merge/review etiquette). Cheers

jamienoss:
@pllim I am currently investigating travis and what we are actually trying to build with...

hstio/hstio.c
pllim: I don't plan on merging anything. I just remember that a long time ago, I tried inserting a `//` comment and it wouldn't build. I am not touching your PRs, don't worry.

jamienoss: Cool thanks. So // (C++ style comments) were added to the C language in C99 along with a load of other useful things, e.g.
 
```
for(unsigned int = 0; i < N, ++i)
{
//i confined to the scope of this block
}
```
as apposed to 
```
unsigned i;
for (i = 0; i < N; ++i)
{
}
//i still in scope
```

pllim: Sure. But does HSTDP machines support that? Whatever we do here, they need to be able to build. I guess @jhunkeler or @rendinam can confirm.

sosey: hstcal is written following c89, until right now (as far as I know) we were compiling with c89 standards. You can't force all of hstcal into c99 standards for this build,  none of the other instrument code has been updated for that. A standards update like that should be directed by @stsci-hack

jamienoss: I wouldn't even dream of doing it for this one build, in fact, I'm in the middle of opening an issue such that we _do_ update the use of the standard to a min of C99 - to which I would obviously let all relevant parties know. FYI most modern compilers default to at least C99 anyhow. In addition, it is backward compatible with C89 (unless we turn on pedantic warnings). I am just in the workings of figuring out how to extract compiler versions in with waf...

jamienoss: I'll open it as is, and we can continue any discussion there...

jhunkeler: @pllim `gcc` uses the `-std=[standard]` argument to lock the code standard during compilation, but as @sosey commented, using it would probably cross the streams. As far as I know... gcc compiles the source code with default standard (c99?) and issues warnings if it detects the usage of a different coding standard (c89, c90, c11, etc). I have no idea if it makes an attempt to fix the the "problem" in-place, or just continues on its merry way if the "problem" is simply informative and not critical to byte code generation.


### Reference (man 1 gcc)
```
-std=
    Determine the language standard.   This option is currently only supported when
    compiling C or C++.

    The compiler can accept several base standards, such as c90 or c++98, and GNU dialects
    of those standards, such as gnu90 or gnu++98.  When a base standard is specified, the
    compiler accepts all programs following that standard plus those using GNU extensions
    that do not contradict it.  For example, -std=c90 turns off certain features of GCC that
    are incompatible with ISO C90, such as the "asm" and "typeof" keywords, but not other
    GNU extensions that do not have a meaning in ISO C90, such as omitting the middle term
    of a "?:" expression. On the other hand, when a GNU dialect of a standard is specified,
    all features supported by the compiler are enabled, even when those features change the
    meaning of the base standard.  As a result, some strict-conforming programs may be
    rejected.  The particular standard is used by -Wpedantic to identify which features are
    GNU extensions given that version of the standard. For example -std=gnu90 -Wpedantic
    warns about C++ style // comments, while -std=gnu99 -Wpedantic does not.

    A value for this option must be provided; possible values are

    c90
    c89
    iso9899:1990
        Support all ISO C90 programs (certain GNU extensions that conflict with ISO C90 are
        disabled). Same as -ansi for C code.

    iso9899:199409
        ISO C90 as modified in amendment 1.

    c99
    c9x
    iso9899:1999
    iso9899:199x
        ISO C99.  This standard is substantially completely supported, modulo bugs and
        floating-point issues (mainly but not entirely relating to optional C99 features
        from Annexes F and G).  See <http://gcc.gnu.org/c99status.html> for more
        information.  The names c9x and iso9899:199x are deprecated.

    c11
    c1x
    iso9899:2011
        ISO C11, the 2011 revision of the ISO C standard.  This standard is substantially
        completely supported, modulo bugs, floating-point issues (mainly but not entirely
        relating to optional C11 features from Annexes F and G) and the optional Annexes K
        (Bounds-checking interfaces) and L (Analyzability).  The name c1x is deprecated.

    gnu90
    gnu89
        GNU dialect of ISO C90 (including some C99 features).

    gnu99
    gnu9x
        GNU dialect of ISO C99.  The name gnu9x is deprecated.

    gnu11
    gnu1x
        GNU dialect of ISO C11.  This is the default for C code.  The name gnu1x is
        deprecated.

    c++98
    c++03
        The 1998 ISO C++ standard plus the 2003 technical corrigendum and some additional
        defect reports. Same as -ansi for C++ code.

    gnu++98
    gnu++03
        GNU dialect of -std=c++98.  This is the default for C++ code.

    c++11
    c++0x
        The 2011 ISO C++ standard plus amendments.  The name c++0x is deprecated.

    gnu++11
    gnu++0x
        GNU dialect of -std=c++11.  The name gnu++0x is deprecated.

    c++14
    c++1y
        The 2014 ISO C++ standard plus amendments.  The name c++1y is deprecated.

    gnu++14
    gnu++1y
        GNU dialect of -std=c++14.  The name gnu++1y is deprecated.

    c++1z
        The next revision of the ISO C++ standard, tentatively planned for 2017.  Support is
        highly experimental, and will almost certainly change in incompatible ways in future
        releases.

    gnu++1z
        GNU dialect of -std=c++1z.  Support is highly experimental, and will almost
        certainly change in incompatible ways in future releases.

```

jamienoss: See #88

jamienoss:
@jhunkeler @pllim see the latest round of comments on #88 - how do we go about preventing the above travis failures? Do we ignore them, do we add -std=c99 to some of the tests in .travis.yml (is this even an option)?

pllim:
I see a lot of discussions indeed but the unanimous decision is not clear to me. What is the *final* decision?

jamienoss:
@pllim use of c99 is allowed but not to mandate strict conformity, i.e. no use of pedantic warnings/errors.

pllim:
Does `-std=c99` satisfy that requirement?

pllim:
Also I saw Megan said, "My personal preference would be to convert the whole package, rather than have mixed standards throughout." Does this option also address Megan's comment?

jamienoss:
@pllim No it doesn't, you'd have to turn on pedantic warnings, upgrade this to an error and then go through and fix the code, leaving the upgraded pedantic warning -> error when complete. This is the difference between conforming and allowing. -std=c99 will allow the use and is backward compatible (at least in gcc) with c90 as long as the pedantic warnings/errors is not used. Though this may not be entirely true with clang which is a lot more strict and more encompassing with its warnings. p.s. I have no idea, not yet having looked, how much work that would be.

pllim:
As long as it does not mess with HSTDP machines, I am okay with this, since Megan and Warren appear to be okay as well. Only @jhunkeler has access to that machine?

jamienoss:
I have removed the C99 for loop declaration instances.

@philhodge This new addition perhaps doesn't belong in hstio.c and nor does it actually change any of the existing code in hstio.c, but perhaps, as Megan suggested, you might fancy taking a quick look to see if I have made any glaring mistakes, please?

hstio/hstio.c
```diff
@@ -110,13 +109,126 @@
...
+    }
+    --reg->cursor;
+}
+void freeAll(PtrRegister * reg)
+{
+    if (!reg || reg->length == 0 || reg->cursor == 0)
+        return;
+
+    {unsigned i;
+    for (i = 1; i < reg->cursor; ++i)
```
jamienoss: As @philhodge pointed out to me, this needs to be ``i <= reg->cursor`` otherwise we leak the last one added. Good catch.

hstio/hstio.c
```diff
@@ -110,13 +109,126 @@
...
+    {int i;
+    for (i = reg->cursor; i >= 0 ; --i)// i >= 0 prevents adding self again
+    {
+        if (reg->ptrs[i] == ptr)
+            return;
+    }}
+
+    if (++reg->cursor >= reg->length)
+    {
+        reg->length += 10;
```
jamienoss: Also, this would be better if it were ``+= PTR_REGISTER_LENGTH``.

jamienoss:
@philhodge please review the [corrections](https://github.com/spacetelescope/hstcal/pull/82/commits/ddd9d854e6f2003749b070603c003ef1da785193) and let me know if all is ok and whether you have finished reviewing everything, please. I can then squash these commits and merge. Thanks again for taking the time to review this for me.

jamienoss:
BTW, I incorporated this change into the new acscte code, that useses this code, and it works.

philhodge:
That looks fine to me.  I'm through reviewing it.

jamienoss:
Thanks, Phil.

Issue #85, opened 2017-04-13
Travis ignoring python setting and uses Python 3.6 anyway


pllim:
For example:

https://travis-ci.org/spacetelescope/hstcal/jobs/221469448 (#82) -- Failing due to https://bugs.python.org/issue29174

https://travis-ci.org/spacetelescope/hstcal/jobs/221460615 (#71)

Issue #73, opened 2017-04-10
Switch to Python 3.6 in Travis?


pllim:
I notice `.travis.yml` still uses Python 3.5. Should we just use 3.6 now, or is that somehow coordinated with HSTDP?

sosey:
@stsci-hack should verify, but delivery of updated python kernel needs to be coordinated with HSTDP. That doesn't mean we can't test 3.6, but we need to be at least testing what's being delivered to DP right now and anything cos/drizzlepac as well as the other python packages are supporting.

pllim:
Would be nice to make sure `waf` still works in Python 3.6 when we are able to switch. Or if Joe and Jamie manage to replace it with `cmake`, then we can completely remove Python from the picture here. :wink:

pllim:
Actually, I take a little back... If #71 is merged, we do still use Python but only for testing. In that case, I think it is okay to just use latest Python (e.g., 3.6) since it is not something HSTDP would care about.

PR #87, opened 2017-04-13
Upgrade waf to 1.9.9

### LINKED ISSUE #86: waf is not compatible with Python 3.6
As evident in our accidental Travis failures and also reported at waf-project/waf#1889. An upgrade to 1.9.9 like #15 should fix it, hopefully.

pllim:
Fix #86 

Needed by #82 to debug actual errors. Otherwise, get `NoneType` exceptions from Python 3.6.

Note that #85 still needs to be addressed because we are not supposed to be using Python 3.6 yet in the first place (see #73) but this PR will future-proof HSTCAL installation.

c/c @jamienoss @sosey @jhunkeler


pllim:
@jamienoss , I got rid of the `NoneType` errors here. But I don't know what these new errors are.

@sosey , should we go ahead and upgrade `waf` to 1.9.9 outside Jamie's changes anyway?

jamienoss:
@pllim Cool, thanks. It seems that the errors are now genuine - it's complaining about my use of the C99 standard.

pllim:
@jamienoss , you can either push your fixes here (to my branch as a maintainer) or continue in #82. Theoretically, if you fixed your actual errors, the Python 3.6 `NoneType` errors will go away as well, as the latter was triggered by the former.

If you want to work in #82, I'll remove your commit from this PR, and make this a pure `waf` upgrade PR.

sosey:
I think it would be nicer to put the waf in separately

pllim:
@sosey , in that case, would be nice to merge the `waf` only PR in ASAP, so Jamie can rebase and take advantage of it. What do you think?

sosey:
@pllim I'm fine with that

pllim:
Okay. I removed changes from #82 here. Will merge when tests pass.

PR #92, opened 2017-04-20
Dms2017 2

stscirij:
new version of calstis for dms 2017.2


pllim:
Is this supposed to fix the failing regression test?

stscirij:
It should fix the failing calstis test in the sense that the new calstis output will be correct, even if it doesn't match the input.  So we'll need to copy the new output to the reference.

PR #93, opened 2017-04-20
wf3cte unessacary re-computation of cte correction due to over-subtra…

### LINKED ISSUE #46: wf3cte unnecessary re-computation of cte correction due to over-subtracted tails
This seems like a bug to me. 

https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/wf3cte/wf3cte.c#L1344

As is, the code checks whether the cte correction has over-subtracted a tail, if so it downscales the cte scaling and re-runs the correction for that given column, however, it is possible for it to re-run the exact same column correction 4 times - 'exact' because it's possible for the conditional checking for the tails to not be reentered (i.e. the first re-run was enough) and thus never downgrading the cte scaling on subsequent re-runs, thus nothing having changed for the other 3 re-runs.

It could be possible that the first re-run was sufficient and as such should prevent further re-computations of the column's correction.

I think ```REDO``` should be reset on each loop?
https://github.com/spacetelescope/hstcal/compare/master...jamienoss:unnecessary-re-comp-of-cte-cor-cols-for-over-sub-tails-wf3cte

jamienoss:
…cted tails

As is, the code checks whether the cte correction has over-subtracted a tail, if so it downscales the cte scaling and re-runs the correction for that given column, however, it is possible for it to re-run the exact same column correction 4 times - 'exact' because it's possible for the conditional checking for the tails to not be reentered (i.e. the first re-run was enough) and thus never downgrading the cte scaling on subsequent re-runs, thus nothing having changed for the other 3 re-runs.

It could be possible that the first re-run was sufficient and as such should prevent further re-computations of the columns correction.

Fixes #46 

### DIFF
--- a/pkg/wfc3/calwf3/wf3cte/wf3cte.c
+++ b/pkg/wfc3/calwf3/wf3cte/wf3cte.c
@@ -1263,8 +1263,8 @@ int inverse_cte_blur(SingleGroup *rsz, SingleGroup *rsc, SingleGroup *fff, CTEPa
 
         if (totflux >= 1) {/*make sure the column has flux in it*/
             NREDO=0; /*START OUT NOT NEEDING TO MITIGATE CRS*/
-            REDO=0; /*FALSE*/
             do { /*replacing goto 9999*/
+                REDO=0; /*FALSE*/
                 /*STARTING WITH THE OBSERVED IMAGE AS MODEL, ADOPT THE SCALING FOR THIS COLUMN*/
                 for (j=0; j<RAZ_ROWS; j++){
                     pix_modl[j] =  Pix(rz.sci.data,i,j);


jamienoss:
@sosey please review.

sosey:
this is fine, but it wont go into 2017.2. I don't have time to validate that it doesn't change data results and the team is not accepting algorithm changes with this delivery.

jamienoss:
Tested this against the WFC3 team's own test dataset /grp/hst/adrizzle/HSTDP-2017_2/.
Out of the 1116 `*raw.fits` files, 147 have `PCTECOR = PERFORM`. However, none of these have the fix-readout-cosmic-ray flag set to `FIXROCR = 1`. So I turned this on for all. Running this against [candidate-1.3.1](https://github.com/spacetelescope/hstcal/tree/candidate-1.3.1) (this is 1.3.0 + the buffer overflow fix #157 ) found no differences in the output data. All corrected files triggered this step/correction, I know this from running candidate-1.3.1 once with `FIXROCR` ON and another OFF, the output `*flc.fits` files all differed. This states that all 147 `raw.fits` files actually test this PR.

PR #94, opened 2017-04-21
Consolidate stiserr.h, wf3err.h, & acserr.h to hstcal/include/hstcalerr.h

jamienoss:
Fixes #95 

Signed-off-by: James Noss <jnoss@stsci.edu>

PR #96, opened 2017-04-21
Consolidate stiserr.h, wf3err.h, acserr.h, & the errors in hstcal/inc…

### LINKED ISSUE #95: Consolidate stiserr.h, wf3err.h, & acserr.h to hstcal/include/err.h

jamienoss:
…lude/imphttab.h to hstcal/include/hstcalerr.h

Fixes #95 


sosey:
I would say it's too late to start inserting package changes to this build. We should keep this on hold until after the delivery and then do package restructuring if it's been requested and the resources approved by @stsci-hack

jamienoss:
See old PR #94 for some comments.

pkg/stis/calstis/lib/mkpkg
```diff
@@ -11,67 +11,67 @@ install:
 	;
 
 calstis.a:
-	add2d.c		<c_iraf.h> <hstio.h> ../stis.h ../stiserr.h ../stisdef.h
```
jamienoss: @jhunkeler I think the other relative paths are incorrect(?) Is this file even still needed or can we just delete it?

jamienoss:
@pllim and/or @jhunkeler can you please review this.

jhunkeler:
So nothing in the code directly references `pkg/stis/calstis/lib/mkpkg`. The only script of this nature generated at compile-time is `cfitsio/mkpkg`, *and* only because it was a direct port from IRAF, however it isn't used by anything.

Restructuring efforts aside -- `pkg/stis/calstis/lib/mkpkg` and `cfitsio/mkpkg` are dead files and should be purged.

jamienoss:
@pllim can you please spend a quick 5min and look over this for me, cheers.

pllim:
I looked at ACS changes and they seem okay. Extrapolating that to WFC3 and STIS, should be okay. I do like how all the error codes are now properly sorted in the same file.

One way to make sure this works is to install this branch, and call one of the CALs with a bad input (e.g., trying to run CALACS with undefined "jref" or trying to run CALWF3 on ACS input file), then make sure the proper error is returned.

jamienoss:
@pllim thank you. This sort of change requires only a build and not a run. As long as the numbers match up, which can be shown by reviewing the git comparison, the only thing left is that something hasn't been left out of the new consolidated header, if this were the case the build would fail.

PR #101, opened 2017-04-27
ACS CTE second generation correction algorithm

### LINKED ISSUE #100: Implement gen2 CTE algorithm in calacs
It has been requested by the ACS INS team that they wish to have calacs upgraded to use the second generation CTE algorithm designed by Jay Anderson (@JayVB.) This is is the algorithm currently used in calwf3, however, the implementation of this, for calacs, is not that implemented in calwf3 - it is instead a newer and faster version.

The primary performance aspects of this new implementation come from removing unnecessary zero initializations, correct storage access to better fit the problem (now column major instead of row major*), and more uses of openmp.

The implementation of these algorithms are not identical though they have been tested to yield identical results.

This change introduces three new command line options to acscte.e and calacs.e.
These are:

``--nthreads <N>`` : specifies the number of threads to use for the CTE correction.

``--pctetab <filename>`` : use 'filename' as the PCTETAB cte reference file instead of that specified by PCTETAB in the image header.

``--ctegen <1|2>`` : indicates intent of which cte algorithm to use, gen1 or gen2. This is a safeguards against incorrect use when the code determines which to use dependent on which PCTETAB generation was specified. See the below description.

The code looks for ``CTE_NAME`` in the specified ``PCTETAB`` header, if it is missing it assumes gen1 cte. If it is = "pixelCTE 2017", gen2 is used. Defaults to gen1 for anything else. The cmd line opt ``--ctegen <1|2>`` can be used to give intent and safeguard, failing if incorrect ``PCTETAB`` given.

*Since the CTE correction algorithm acts along columns and not rows it makes more sense to store the data in column major storage order so as to give contiguous memory access when iterating over rows, i.e. along a column.

jamienoss:
Resolves #100

### DIFF
--- a/pkg/acs/calacs/acscte/dopcte.c
+++ b/pkg/acs/calacs/acscte/dopcte.c
@@ -12,7 +12,7 @@
 #include "pcte.h"
 
 
-static int get_amp_array_size(const ACSInfo *acs, SingleGroup *x,
+int get_amp_array_size_acs_cte(const ACSInfo *acs, SingleGroup *x,
                               const int amp, char *amploc, char *ccdamp,
                               int *xsize, int *ysize, int *xbeg,
                               int *xend, int *ybeg, int *yend);
@@ -39,7 +39,7 @@ static int unmake_amp_array(const ACSInfo *acs, const SingleGroup *im,
 
    MRD 10 Mar 2011
 */
-int doPCTE (ACSInfo *acs, SingleGroup *x) {
+int doPCTEGen1 (ACSInfo *acs, SingleGroup *x) {
 
     /* arguments:
        acs     i: calibration switches, etc
@@ -61,7 +61,7 @@ int doPCTE (ACSInfo *acs, SingleGroup *x) {
     double dpde_l[NUM_LEV];
 
     /* structure to hold CTE parameters from file */
-    CTEParams pars;
+    ACSCTEParams pars;
 
     /* temporary variable used during final error calculation */
     double temp_err;
@@ -164,7 +164,7 @@ int doPCTE (ACSInfo *acs, SingleGroup *x) {
         amp = *amploc - AMPSORDER[0];
 
         /* get amp array size */
-        if (get_amp_array_size(acs, x, amp, amploc, ccdamp,
+        if (get_amp_array_size_acs_cte(acs, x, amp, amploc, ccdamp,
                                &amp_xsize, &amp_ysize, &amp_xbeg,
                                &amp_xend, &amp_ybeg, &amp_yend)) {
             return (status);
@@ -212,9 +212,10 @@ int doPCTE (ACSInfo *acs, SingleGroup *x) {
         }
 
         /* perform CTE correction */
+        int onecpu = acs->nThreads <= 1 ? YES : NO;
         if (FixYCte(amp_arr1, amp_arr2, amp_sig_arr, amp_cor_arr, pars.sim_nit,
                     pars.shft_nit, pars.sub_thresh, cte_frac_arr, pars.levels,
-                    dpde_l, chg_leak_lt, chg_open_lt, acs->onecpu)) {
+                    dpde_l, chg_leak_lt, chg_open_lt, onecpu)) {
             return (status);
         }
 
@@ -265,7 +266,7 @@ int doPCTE (ACSInfo *acs, SingleGroup *x) {
    the logic for figuring out the amp regions has been copied from doblev.
    - MRD 14 Mar 2011
 */
-static int get_amp_array_size(const ACSInfo *acs, SingleGroup *x,
+int get_amp_array_size_acs_cte(const ACSInfo *acs, SingleGroup *x,
                               const int amp, char *amploc, char *ccdamp,
                               int *xsize, int *ysize, int *xbeg, int *xend,
                               int *ybeg, int *yend) {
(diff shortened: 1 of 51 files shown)


jamienoss:
FYI: This includes #81 which should be merged first.

jamienoss:
@stsci-hack @sosey FYI

pkg/acs/calacs/Updates
```diff
@@ -49,10 +49,10 @@ Update for version 8.2.0 - 12-Aug-13 (PLL)
     acscte/getctesw.c
     calacs/acsinit.c
     calacs/calacs.c
-    calacs/calacs.h
+    calacs/acsinfo.h
```
pllim: Why are you changing old history? You should have created your own entry at the top.

pllim:
So, if I understood correctly, this is going in but **not** used in the pipeline by default? If not, has ACS Team signed off on it?

jamienoss:
@pllim The code's going in, it now self-determines which algorithm to use so that whether the new code is used or not is entirely dependent on which PCTETAB file is given.

This is still in the last stages of being hashed out by the ACS team. We have a meeting at 4:15 today to discuss and update.

pllim:
Ah, I did not see #100 . Thanks for the update.

pkg/acs/calacs/Updates
jamienoss: Good spot. I wasn't, at least not intentionally, at one point I attempted to disambiguate the dangerous case of having two ACSInfo structs. I gave up at some point and this must be the remainder of that alteration that I hadn't noticed was still present. Thanks.

jamienoss:
@stsci-hack short-circuited.

jamienoss:
@stsci-hack do you want someone to look over this or should I just merge it?

pkg/acs/calacs/acscte/dopcte.c
```diff
@@ -12,7 +12,7 @@
 #include "pcte.h"
 
 
-static int get_amp_array_size(const ACSInfo *acs, SingleGroup *x,
```
pllim: Is there a reason why you removed "static"?

pkg/acs/calacs/acscte/getcteflag.c
```diff
@@ -91,9 +92,21 @@ static int checkPCTE (Hdr *phdr, ACSInfo *acs, int *missing, int *nsteps) {
             return (status);
         }
 
-        if (GetTabRef (acs->refnames, phdr,
-                       "PCTETAB", &acs->pcte, &acs->pctecorr))
-            return (status);
+        if (acs->pcteTabNameFromCmd && *acs->pcteTabNameFromCmd != '\0')
+        {
+            char msgBuffer[256];
```
pllim: Instead of magic number, I suggest you use one of the pre-defined constants. For example, `ACS_LINE` from `acs.h` should do well enough for this purpose.

pkg/acs/calacs/acscte/maincte.c
```diff
@@ -33,8 +37,12 @@ int main (int argc, char **argv) {
     /*int switch_on = 0;*/	/* was any switch specified? */
     int printtime = NO;	/* print time after each step? */
     int verbose = NO;	/* print additional info? */
-    int quiet = NO;	/* print additional info? */
     int onecpu = NO; /* Use OpenMP (multi vs single CPU mode), if available? */
+    int quiet = NO;	/* print additional info? */
+    unsigned cteAlgorithmGen = 0; //Use gen1cte algorithm rather than gen2 (default)
+    unsigned nThreads = 0;
+    char pcteTabNameFromCmd[255];
```
pllim: Replace magic number with, say, `ACS_LINE` from `acs.h` (see my comment above).

pkg/acs/calacs/calacs/acsmain.c
```diff
@@ -21,12 +25,16 @@ int main(int argc, char **argv) {
 	int debug = NO;		/* print debug statements during processing? */
 	int quiet = NO;		/* suppress STDOUT messages? */
 	int onecpu = NO;		/* suppress OpenMP usage? */
+    unsigned cteAlgorithmGen = 0; //Use gen1cte algorithm rather than gen2 (default)
+    char pcteTabNameFromCmd[255];
```
pllim: Replace magic number. Maybe this one can be `SZ_LINE` from `pcte.h`.

pkg/acs/calacs/include/acsinfo.h
```diff
@@ -120,6 +126,7 @@ typedef struct {
 
     /* calibration images and tables for ACSCTE */
     RefTab pcte;        /* Pixel CTE parameters table */
+    char pcteTabNameFromCmd[255];
```
pllim: Magic number (see comment above).

pllim:
And I noticed that the following files are not updated but they should be for book-keeping:

* `calacs/Dates`
* `calacs/History`
* `calacs/Updates` (for such big changes, I think it is okay to be a bit lazy here and use wildcards; e.g. `acscte/*`)
* `calacs/include/acsversion.h`

pkg/acs/calacs/acscte/dopcte.c
jamienoss: Yes, so that it can be called by both algorithms. The gen2 exists in a different file.

pkg/acs/calacs/acscte/getcteflag.c
jamienoss: Cool, thanks.

pkg/acs/calacs/acscte/dopcte.c
pllim: Okay, that is fine. Just want to make sure this is not another rebase error.

jamienoss: Cool, thanks for pointing this out.

jamienoss:
Just waiting on travis...

PR #122, opened 2017-05-12
Guarded all headers

jhunkeler:
I'm not sure much explanation is required here. Unguarded headers in project of this size is a bad thing.

EDIT:
- Headers that already had guards in place were not touched.
- `cfitsio` was not touched. It's not my place to guard their headers... or is it?


jamienoss:
Cheers dude.

PR #124, opened 2017-05-16
ACS CTE: correct PCTETAB in corrected image file header from --pctetab <file>

### LINKED ISSUE #123: ACS CTE: correct PCTETAB in corrected image file header from --pctetab <file>
[The following code](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/acscte/docte.c#L215-L219) should now be activated.

```c
...
/*if ((status = PutKeyStr(x[0].globalhdr, "PCTETAB", cteTabFilename, "CTE Correction Table")))
{
    trlmessage("(pctecorr) Error updating PCTETAB keyword in image header");
    return status;
}*/
...
```

Without the above code, the PCTETAB filename provided via ``--pctetab <filename>``, that used for the cte correction, will not be populated in the resultant corrected image header.

It was previously decided for this not to be the case. After discussing the options with the ACS team lead, either to remove the cmd line option ``--pctetab <filename>`` or to activate the above code, the latter was decided upon. I believe this to be the correct course of action, leaving the cmd option available but not updating the resultant header should never have been the decided design.

jamienoss:
This also corrects some trlmessages() being used as errors to trlerror

resolves #123

### DIFF
--- a/ctegen2/ctehelpers.c
+++ b/ctegen2/ctehelpers.c
@@ -339,7 +339,6 @@ No.    Name         Type      Cards   Dimensions   Format
         status = KEYWORD_MISSING;
         return status;
     }
-
     sprintf(MsgText,"PCTETRSH: %g",pars->thresh);
     trlmessage(MsgText);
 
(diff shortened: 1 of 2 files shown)


jamienoss:
@pllim can you take a quick look at this before I merge, please?

pllim:
LGTM

PR #126, opened 2017-05-17
Issue61 out of bounds memory access

### LINKED ISSUE #61: wf3cte - BIG BOSS LEVEL COMPLETE!
[This is it](https://github.com/spacetelescope/hstcal/blob/1.2.0/pkg/wfc3/calwf3/wf3cte/wf3cte.c#L999-L1010), _the_ bug in wf3cte.

```C
...
  if (i==1 &&  RAZ_ROWS-1>=j  && j>0 ) {

        dval9 = obsloc[i][j-1]  - rszloc[i][j-1] +
            obsloc[i][j]    - rszloc[i][j]  +
            obsloc[i][j+1]  - rszloc[i][j+1] +
            obsloc[i-1][j-1]- rszloc[i-1][j-1] +
            obsloc[i-1][j]  - rszloc[i-1][j] +
            obsloc[i-1][j+1]- rszloc[i-1][j+1] +
            obsloc[i+1][j-1]- rszloc[i+1][j-1] +
            obsloc[i+1][j]  - rszloc[i+1][j] +
            obsloc[i+1][j+1]- rszloc[i+1][j+1];
    }
...
```

The conditional should be ``RAZ_ROWS-1>j`` __NOT__  ``RAZ_ROWS-1>=j``.

Clarification edit:
With ``RAZ_ROWS-1=>j`` for ``j = RAZ_ROWS-1`` => any ``j+1`` index ([here](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/wf3cte/wf3cte.c#L1003), [here](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/wf3cte/wf3cte.c#L1006), & [here](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/wf3cte/wf3cte.c#L1009)) makes ``j = RAZ_ROWS``, this is C with zero based indices and as such the ``RAZ_ROWS``th element doesn't exist.

__UPDATE:__
Any garbage read by this access is luckily bound by ``+/- 0.33*read_noise`` in [the following code](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/wf3cte/wf3cte.c#L1012-L1018)

```c
...
dval9 =dval9 / 9.;
dval9u = dval9;

if (dval9u > (rnsig*0.33))
    dval9u =  rnsig*0.33;
if (dval9u <  rnsig*-0.33)
    dval9u = rnsig*-0.33;
...
```

The compiler flag ``-fstack-protector-all`` was added to the main [build script](https://github.com/spacetelescope/hstcal/commit/13fc8aa3f4d8dec8168ab35b3ea1a2b9892ab60a) in order to 'somehow' fix the issue. [This is the responsible commit (github)](https://github.com/spacetelescope/hstcal/commit/13fc8aa3f4d8dec8168ab35b3ea1a2b9892ab60a), [svn](https://trac.stsci.edu/ssb/stsci_python/changeset/45938), and [Trac issue](https://trac.stsci.edu/ssb/stsci_python/ticket/1193) . Here is the commit message:

>- moved the regular compile from optimization 1 to 2, added fstack-protect-all flag, still need to figure out where exactly the stack buffer overflow is happening, but its' consistently producing the correct results for now at least.

What makes no sense to me is that if the bug truly was a buffer overflow then adding stack protection would only abort the execution upon detecting an overflow (writing beyond the allowed stack space), i.e. if it was segfaulting before (from corrupted memory) it would now only abort - either way, it still 'crashes'. This compiler flag could __NEVER__ actually 'fix' an overflow issue. As it turns out the bug in this issue is actually a read and not a write so it still wouldn't trigger the stack protector to abort, unless these are different bugs.

I identified the relation to stack protection when I  experienced the bug, tried to debug with --debug, a build that uses -O0 and no stack protection. After determining that it wasn't directly the optimization difference but the use of stack protection in the release -O2 build, I came to the above conclusion.

jamienoss:
fixes #61

### DIFF
--- a/pkg/wfc3/calwf3/wf3cte/wf3cte.c
+++ b/pkg/wfc3/calwf3/wf3cte/wf3cte.c
@@ -997,7 +997,7 @@ int find_dadj(int i ,int j, double obsloc[][RAZ_ROWS], double rszloc[][RAZ_ROWS]
     dval9 = 0.;
 
     /*COMPARE THE SURROUNDING PIXELS*/
-    if (i==1 &&  RAZ_ROWS-1>=j  && j>0 ) {
+    if (i==1 &&  RAZ_ROWS-1>j  && j>0 ) {
 
         dval9 = obsloc[i][j-1]  - rszloc[i][j-1] +
             obsloc[i][j]    - rszloc[i][j]  +


pkg/wfc3/calwf3/wf3cte/wf3cte.c
```diff
@@ -996,7 +996,7 @@ int find_dadj(int i ,int j, double obsloc[][RAZ_ROWS], double rszloc[][RAZ_ROWS]
     dval9 = 0.;
 
     /*COMPARE THE SURROUNDING PIXELS*/
-    if (i==1 &&  RAZ_ROWS-1>=j  && j>0 ) {
```
philhodge: Just to answer the specific question here:  Yes, when j equals RAZ_ROWS - 1, one element is read from both obsloc and rszloc that is out of bounds of these arrays.  There are two other points (i.e. uses of j+1) where the wrong array element is read, but those are at least within the bounds of the arrays.

jamienoss:
Rebased up to date.

jamienoss:
Closing this and moving to #366 so as to base against candidate-2.2.0

PR #132, opened 2017-05-19
PtrRegister aborts/crashed on addPtr() due to realloc mem not being allocated

### LINKED ISSUE #131: PtrRegister aborts/crashed on addPtr() due to realloc mem not being allocated
@philhodge noted that the reallocation inside of ``addPtr()`` was faulting. The below driver of his was used to demonstrate this:

```c
# define NELEM 22
# define NITER 5

int main(int argc, char **argv) {

       PtrRegister reg;
       float *x[NELEM];
       int i, j;

       for (j = 0;  j < NITER;  j++) {
           initPtrRegister(&reg);
           printf("allocated ...\n"); fflush(stdout);
           for (i = 0;  i < NELEM;  i++) {
               x[i] = malloc(23456);
               addPtr(&reg, x[i], free);
               printf("x[%d] saved\n", i); fflush(stdout);
           }
           freeAll(&reg);
           freeReg(&reg);
           printf("... freed\n"); fflush(stdout);
       }

       printf("done\n"); fflush(stdout);

       return 0;
}
```

resulting in (osx):
```
a.out(6601,0x7fff7af8f000) malloc: *** error for object 0x7fbcf34031c0: pointer being realloc'd was not allocated
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

Debugging the code I found that it was successfully reallocating on the first attempt but failing on the second. The reason for this is that the [following code](https://github.com/spacetelescope/hstcal/blob/master/hstio/hstio.c#L155-L156) is incorrect, in that the ``reg-ptrs`` and ``reg->freeFunctions`` are not being updated with the newly reallocated pointers. (Also note the ``&`` on ``reg->ptrs`` that is a bug in itself.)

```c
...
assert(realloc(&reg->ptrs, reg->length*sizeof(*reg->ptrs)));
assert(realloc(reg->freeFunctions, reg->length*sizeof(*reg->freeFunctions)));
...
```
It turns out that the first ``realloc`` was just expanding the current memory space, which is why subsequent calls to ``addPtr()`` were not segfaulting. On the second ``realloc`` it must of had  to move the memory which was obviously not being kept track of.

The code should simply read:

```c
...
assert(reg->ptrs = realloc(reg->ptrs, reg->length*sizeof(*reg->ptrs)));
assert(reg->freeFunctions = realloc(reg->freeFunctions, reg->length*sizeof(*reg->freeFunctions)));
...
```

Thanks for finding this @philhodge 👍 

This code exists in 1.2.0rc2 but is not used. It is used in [candidate-1.3.0](https://github.com/spacetelescope/hstcal/tree/candidate-1.3.0), however, not enough pointers are ever added to cause it to reallocate the registers beyond their [initial length of 10](https://github.com/spacetelescope/hstcal/blob/candidate-1.3.0/include/hstio.h#L147)

jamienoss:
fixes #131

### DIFF
--- a/hstio/hstio.c
+++ b/hstio/hstio.c
@@ -152,8 +152,8 @@ void addPtr(PtrRegister * reg, void * ptr, void * freeFunc)
     if (++reg->cursor >= reg->length)
     {
         reg->length += PTR_REGISTER_LENGTH_INC;
-        assert(realloc(&reg->ptrs, reg->length*sizeof(*reg->ptrs)));
-        assert(realloc(reg->freeFunctions, reg->length*sizeof(*reg->freeFunctions)));
+        assert(reg->ptrs = realloc(reg->ptrs, reg->length*sizeof(*reg->ptrs)));
+        assert(reg->freeFunctions = realloc(reg->freeFunctions, reg->length*sizeof(*reg->freeFunctions)));
     }
     reg->ptrs[reg->cursor] = ptr;
     reg->freeFunctions[reg->cursor] = freeFunc;


jamienoss:
@philhodge can you review, please?

philhodge:
It looks as if you haven't changed lines 221 and 223, where you assign NULL to reg->ptrs[0] and reg->freeFunctions[0].  Were you going to do that?  The changes to lines 155 and 156 (the calls to realloc) look fine to me.

jamienoss:
These changes will be in another PR associated with #133. That Pr will also deal with a bit of refactoring to remove the ``this`` pointer storage etc. I'll tag you to review that one also, when I'm done testing. Cheers

PR #142, opened 2017-05-20
Refactor PtrRegister to correctly handle 'this' pointers

### LINKED ISSUE #141: Refactor PtrRegister to correctly handle 'this' pointers

jamienoss:
resolves #141

### DIFF
--- a/hstio/hstio.c
+++ b/hstio/hstio.c
@@ -121,20 +121,26 @@
 */
 const char *hstio_version = HSTIO_VERSION;
 
+void * newPtrRegister()
+{
+    void * this = malloc(sizeof(PtrRegister));
+    initPtrRegister(this);
+    addPtr(this, this, &free); // Note: freeFunctions[0] is ignored anyhow
+    return this;
+}
 void initPtrRegister(PtrRegister * reg)
 {
     reg->cursor = 0; //points to last ptr NOT next slot
-    reg->length = PTR_REGISTER_LENGTH_INC;
-    reg->ptrs = calloc(reg->length+1, sizeof(*reg->ptrs));
+    reg->length = PTR_REGISTER_LENGTH_INC+1; //+1 to special case reg->ptrs[0] for 'this' pointer only
+    reg->ptrs = malloc(reg->length*sizeof(*reg->ptrs));
     assert(reg->ptrs);
-    reg->freeFunctions = calloc(reg->length+1, sizeof(*reg->freeFunctions));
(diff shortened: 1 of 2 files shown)


jamienoss:
I tested this with the following driver and valgrind, which showed zero leaks or errors:

```c
int main()
{
    int i, j;
    for (j = 0;  j < NITER;  j++) {
       PtrRegister * reg = newPtrRegister();
       addPtr(reg, reg, &free); // test (not actually needed)
       printf("allocated ...\n"); fflush(stdout);
       for (i = 0;  i < NELEM;  i++) {
           addPtr(reg, malloc(23456), &free);
           printf("x[%d] saved\n", i); fflush(stdout);
       }
       freeOnExit(reg);
       printf("... freed\n"); fflush(stdout);
    }
    printf("done\n"); fflush(stdout);

    return 0;
}
```

hstio/hstio.c
```diff
@@ -121,20 +121,26 @@
...
+    addPtr(this, this, &free); // Note: freeFunctions[0] is ignored anyhow
+    return this;
+}
 void initPtrRegister(PtrRegister * reg)
 {
     reg->cursor = 0; //points to last ptr NOT next slot
-    reg->length = PTR_REGISTER_LENGTH_INC;
-    reg->ptrs = calloc(reg->length+1, sizeof(*reg->ptrs));
+    reg->length = PTR_REGISTER_LENGTH_INC+1; //+1 to special case reg->ptrs[0] for 'this' pointer only
+    reg->ptrs = malloc(reg->length*sizeof(*reg->ptrs));
```
jamienoss: Since we call ``realloc`` which doesn't zero initialize don't bother here.

jamienoss:
@philhodge please review.

philhodge:
The only thing that doesn't look quite right to me is in function freePtr, lines 181 - 185.  If ptr has a non-zero value but isn't found in reg->ptrs (i.e. if the calling function passes some random value instead of an actual, saved pointer), the `break` will not execute, and i will be 0 after the for loop completes.  Then reg->freeFunctions[0] will be called to free `ptr`.

Otherwise, this looks fine.

jamienoss:
@philhodge thanks, that's actually something that already exists, unrelated to this PR, good catch. I have opened #143 to deal with this as that will have to be merged into an earlier branch.

FYI, it's easier to understand your review when you make line comments, please.

Thanks again.

PR #144, opened 2017-05-22
PtrRegister: freePtr() assumes ptr passed has definitely been added

### LINKED ISSUE #143: PtrRegister: freePtr() assumes ptr passed has definitely been added
The following [code](https://github.com/spacetelescope/hstcal/blob/master/hstio/hstio.c#L167-L175) is incorrect.

```c
...
    int i;
    for (i = reg->cursor; i > 0 ; --i)
    {
        if (reg->ptrs[i] == ptr)
            break;
    }

    //call function to free ptr
    reg->freeFunctions[i](ptr);
...
```

It assumes that ``ptr`` passed in will allays be found, i.e. that it has definitely been added to the register already.

It should probably be something like:

```c
...
    int i;
    Bool found = False;
    for (i = reg->cursor; i > 0 ; --i)
    {
        if (reg->ptrs[i] == ptr)
       {
            found = True;
            break;
        }
    }
    if (!found)
        return; //or better still assert(0) as this is an internal coder fault

    //call function to free ptr
    reg->freeFunctions[i](ptr);
...
```

jamienoss:
fixes #143

### DIFF
--- a/hstio/hstio.c
+++ b/hstio/hstio.c
@@ -165,11 +165,17 @@ void freePtr(PtrRegister * reg, void * ptr)
         return;
 
     int i;
+    Bool found = False;
     for (i = reg->cursor; i > 0 ; --i)
     {
         if (reg->ptrs[i] == ptr)
+        {
+            found = True;
             break;
+        }
     }
+    if (!found)
+        assert(0); //internal error: the ptr trying to be freed was not added to the register
 
     //call function to free ptr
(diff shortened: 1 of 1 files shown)


jamienoss:
@philhodge FYI

philhodge:
That looks good.

jamienoss:
Woops, just realized that I pushed this to upstream, never mind.

PR #145, opened 2017-05-23
STIS: cs2: Immediate SIGSEGV

jhunkeler:
ATTN @stscirij 

Just following up on what we talked about the other day.

Other stuff:
- Tried to normalize path and filename lengths with CFITSIO and the rest of HSTCAL.
- There's already an if statement checking whether argc is less than 3, so it seemed odd to check if argc was greater than 3 immediately after. Also, the while loop: `ctoken = 3; while (ctoken < argc)` still performs proper argument scanning, assuming additional arguments were issued past `input` and `output`.

### DIFF
--- a/pkg/stis/calstis/cs2/cs2.c
+++ b/pkg/stis/calstis/cs2/cs2.c
@@ -23,7 +23,7 @@ int main (int argc, char **argv) {
 
 	c_irafinit (argc, argv);
 
-        /* 1Kb padding per path, to allow for string modifications later */
+	/* 1Kb padding per path, to allow for string modifications later */
 	if ((input = calloc(argc * (SZ_PATHNAME * 2), sizeof(char))) == NULL) {
 	    printf ("ERROR    out of memory in cs2.c\n");
 	    exit (ERROR_RETURN);
--- a/pkg/stis/calstis/cs2/cs2.c
+++ b/pkg/stis/calstis/cs2/cs2.c
@@ -14,7 +14,7 @@ int main (int argc, char **argv) {
 
 	int status = 0;			/* zero is OK */
 
-	char	*input;			/* list of input file names */
+	char	*input = NULL;		/* list of input file names */
 	char 	output[STIS_LINE];	/* output file name */
 	clpar 	par;			/* parameters used */
 	int 	newpar[MAX_PAR+1];	/* user specifiable parameters */
@@ -23,8 +23,8 @@ int main (int argc, char **argv) {
 
 	c_irafinit (argc, argv);
 
-	/* 1Kb padding per path, to allow for string modifications later */
-	if ((input = calloc(argc * (SZ_PATHNAME * 2), sizeof(char))) == NULL) {
+	/* ~1Kb padding per path, to allow for string modifications later */
+	if ((input = calloc(argc * (STIS_FNAME * 2), sizeof(char))) == NULL) {
 	    printf ("ERROR    out of memory in cs2.c\n");
 	    exit (ERROR_RETURN);
 	}
--- a/pkg/stis/calstis/cs2/cs2.c
+++ b/pkg/stis/calstis/cs2/cs2.c
@@ -23,7 +23,8 @@ int main (int argc, char **argv) {
 
 	c_irafinit (argc, argv);
 
-	if ((input = calloc (strlen(argv[1]) + 1, sizeof(char))) == NULL) {
+        /* 1Kb padding per path, to allow for string modifications later */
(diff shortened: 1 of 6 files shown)


pllim:
I wasn't in that discussion. What is the reasoning to increase constants in HSTIO include so much?

jhunkeler:
@plim Depending on the number of comma delimited filenames passed to cs2.e as `input`, *and* a file read error occurs, the string copied to `error_message` in HSTIO will overrun ERRLINEWIDTH and trigger a segfault. I admit, the 2KBs of extra padding is a stop-gap measure not a solution.

CFITSIO accepts input filename up to 1025 bytes in length. The constants in HSTIO were much smaller than the maximum allowed, so if when you exceeded 511 bytes you'd segfault the program. 

Just off the top of my head...

- There are few sanity checks in place to determine if input strings are too long.
- If any sanity checks do exist they are not implemented consistently.
- There's no file length consistency between modules.
- There's a lot of `# define SZ_* X` redefinitions peppered throughout HSTCAL in general, so it's difficult to know which "SZ_*" or "MODULE_*" definition is getting pulled in at any given time.
- Insecure string operations are used all over the place (`strcpy` vs `strncpy`) so is difficult to track which string caused which failure.

pllim:
513 characters can fit over 25 comma separated input files, so the problem will only manifest if there is somehow a need to run this executable on so many files at the same time. ACS did encounter such a problem for ACSREJ but that was a special case that only happened after 10 years in orbit.

philhodge:
Regarding calstis, different parts ("modules") were written by three different people, and styles differed.  Some modules allocate memory for input or output based on strlen(argv[i]), but that's a recent change, maybe five years ago.  The input and/or output can be a comma-separated list of names, and it can be long (as @pllim pointed out) for cosmic-ray rejection.  The individual file names will not likely be extremely long, and if they are the user can rename them.

I certainly agree regarding consistency in buffer lengths and using strncpy.

pkg/stis/calstis/cs2/cs2_command.c
```diff
@@ -54,103 +54,100 @@ int newpar;		o: parameter been set by the user?
...
-
-	            /* No match. */
-	            } else
-	                return (syntax_error (argv[ctoken]));
-	        } 
-	    }
-	}
-
-	return (0);
+	strncpy (input, argv[1], strlen(argv[1])+1);
```
jamienoss: ``strncpy (input, argv[1], strlen(argv[1])+1);`` and ``strcpy (input,  argv[1]);`` are identicle.

jamienoss:
FYI - I'd be more worried bout the use of uninitialized instantiated char arrays i.e.
``char buffer[SOME_MACRO_LENGTH]`` without the subsequent initialization ``*buffer = '\0'``

pkg/stis/calstis/cs2/cs2.c
```diff
@@ -23,7 +23,8 @@ int main (int argc, char **argv) {
 
 	c_irafinit (argc, argv);
 
-	if ((input = calloc (strlen(argv[1]) + 1, sizeof(char))) == NULL) {
+	/* ~1Kb padding per path, to allow for string modifications later */
+	if ((input = calloc(argc * (STIS_FNAME * 2), sizeof(char))) == NULL) {
```
philhodge: What was wrong with allocating the buffer size based on the length of the argument?  And why should argc have anything to do with it?

jamienoss: I would guess that it is an attempt to handle the worse case scenario that all (or significantly all) of the passed in arguments are filenames?
To me, this seems like a dangerous amount of stack allocation.
I have only briefly looked at the code but I'm not sure why this is the case at all.

jhunkeler: I agree, Phil, this calloc should just allocate the length of the input string and nothing more. Jamie -- yeah exactly. Sorry for the confusion 😬. 

The massive calloc was an attempt to brute-force "success" by allocating far more memory than required (in a silly way no less). Of course, this lead to some interesting discoveries within HSTIO's error handling routines, but that's about it.

I've been digging through the POSIX docs as well as various filesystem comparison charts all night trying to figure out a good way to handle path/file length limits. I think it may be possible to consolidate some of SZ_* definitions into a common place (either a header or dynamically generate them at run-time) and not adversely effect anything.

pllim: > digging through the POSIX docs 

You should close that can o' worms before they evolved into face-huggers.

jamienoss:
@jhunkeler could you please collate all of this info into an issue and open it please? This should not get lost.

P.s. we should not be opening orphaned PRs, i.e. without an associated issue.

PR #147, opened 2017-06-01
Bump ACS version to 9.2.0

### LINKED ISSUE #146: Prep CALACS for 2017.2a delivery
ACS Team has approved the CTE code for delivery this morning during ACS meeting. Now, the following needs to be done to wrap things up:

* Add an entry to `pkg/acs/calacs/Updates` (see "History" and "Dates")
* Updated date stamp in `ACS_CAL_VER` to "28-Apr-2017" to be consistent with "History"
* Tag a new release candidate for 2017.2a -- This is needed for delivery to HSTDP for testing

c/c @jhunkeler @rendinam @stsci-hack

jamienoss:
This updates the following files:

pkg/acs/calacs/Dates
pkg/acs/calacs/History
pkg/acs/calacs/Updates
pkg/acs/calacs/include/acsversion.h

resolves #146

### DIFF
--- a/pkg/acs/calacs/Dates
+++ b/pkg/acs/calacs/Dates
@@ -1,4 +1,4 @@
- 28-Apr-2017:  CALACS 9.1.0 New CTE algorithm implemented
+ 01-Jun-2017:  CALACS 9.2.0 New CTE algorithm implemented.
  21-Feb-2017:  CALACS 9.1.0 New SINKCORR step added in ACSCCD for WFC.
  22-Nov-2016:  CALACS 9.0.0 BLEVCORR now uses new OSCNTAB that correctly
                             process all subarrays.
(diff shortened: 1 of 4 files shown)


jamienoss:
@pllim please review, have I missed something?

pllim:
I thought we discussed about not bumping the version and doing it your way? But even as-is, this looks okay to me. We can worry about tagging RC once you get the final reference file and regression tests sorted out. Thanks! 👍

jamienoss:
But like you said, the instrument team had been testing your sinkcorr code etc which was versioned as 9.1.0, as you pointed out, they may get confussed if the new cte is added as the same version. Your were/are right about not changing history, so I even moved my cte changes to 9.2 rather than 9.1.

In the future, we should not be doing this, pkg version changes should only exist with an actual corresponding release version instead of being orphaned as they have been.

pllim:
Sure. Thanks!

pllim:
@jhunkeler has a `relics` package that does auto-versioning but not sure how to apply that to CALXXX each separately here. But that is a discussion for post-2017.2a.

Issue #130, opened 2017-05-18
WFC3 CTE: add error contribution from cte correction to err extension


jamienoss:
Calacs produces a contribution to the error array in the resultant image from the CTE correction. This is not the case for calwf3.

The [generation 1](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/acscte/dopcte.c#L224-L234) error computation and the [generation 2](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/acscte/dopcte-gen2.c#L226-L241) is computed as 10% of the correction itself added in quadrature.

```
correction = correctedImage - inputImage
error = sqrt(correction^2 + existingError^2)
```

This error contribution _should_ be implemented in calwf3.

### Update:
Possible hurdle - the CTE correction takes place on the raw image which has no error. The error array is introduced somewhere along the pipeline, probably in the ccd stage. So we have to make sure that the error introduced from the CTE is _not_ blown away later where it would previously have been initialized. Hopefully, it has been designed to cope with this.

jamienoss:
WIP - https://github.com/spacetelescope/hstcal/compare/master...jamienoss:issue130-wfc3-cte-add-error-contribution

jamienoss:
I just tested this on iacr51ojq_raw.fits and the CTE error contribution does indeed make its way into the final _flc.fits file after running calwf3.

This file has the following calibration switches set:
```
 / CALIBRATION SWITCHES: PERFORM, OMIT, COMPLETE, SKIPPED                                                                                          WRTERR  =                    T / write out error array extension
DQICORR = 'PERFORM '           / data quality initialization
ATODCORR= 'OMIT    '           / correct for A to D conversion errors 
BLEVCORR= 'PERFORM '           / subtract bias level computed from overscan
BIASCORR= 'PERFORM '           / Subtract bias image
FLSHCORR= 'OMIT    '           / post flash correction
CRCORR  = 'OMIT    '           / combine observations to reject cosmic rays
EXPSCORR= 'PERFORM '           / process individual observations after cr-reject
SHADCORR= 'OMIT    '           / apply shutter shading correction 
DARKCORR= 'PERFORM '           / Subtract dark image
FLATCORR= 'PERFORM '           / flat field data
PCTECORR= 'PERFORM '           / cte correction
PHOTCORR= 'PERFORM '           / populate photometric header keywords
DRIZCORR= 'PERFORM '           / drizzle processing
[...]

PR #157, opened 2017-06-26
Issue 150 - Buffer overflow

### LINKED ISSUE #150: Buffer overflow in strcpy called from calwf3
When running calw3.e on iacr51ojq_raw.fits (on OSX) in the terminal it completes without fault, however, when running with GDB it aborts [here](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/calwf3/wf3member.c#L106) in ``calwf3/wf3members.c``.

```c
int GetSingle (AsnInfo *asn, WF3Info *wf3) {
...
strcpy(outroot, asn->filename);
...
```

This is due to ``outroot`` being of length ``SZ_CBUF`` := 24 ([here](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/include/msg.h#L4)) when it should really be of length ``SZ_FNAME`` (in this instance ``strlen(asn->filename)`` = 62). However, it seems to me like the entire use of outroot is incorrect in [int GetSingle (AsnInfo *asn, WF3Info *wf3)](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/calwf3/wf3member.c#L95-L134)

```c
int GetSingle (AsnInfo *asn, WF3Info *wf3) {

/* arguments:
AsnInfo *asn      	i: calibration flags and other info
WF3Info *wf3		o: exposure specific flags and info
*/
	extern int status;
	char rootname[SZ_FNAME+1];
	char outroot[SZ_CBUF+1];
	void FindAsnRoot (char *, char *);
	
	strcpy(outroot, asn->filename);
	
	/* Make sure we are only passing a rootname, and not a full filename.*/
	FindAsnRoot (outroot, rootname);
	strcpy (wf3->outroot, rootname);
	strcpy (wf3->rootname, rootname);
	
	if (asn->debug) {
	    sprintf (MsgText, "GetSingle: Rootname: %s, Output rootname: %s",
		     rootname, outroot);
	    trlmessage (MsgText);
	}

	/* Check to see that this value of rootname is what we really need. */
	strcpy (wf3->asn_table, asn->asn_table);	
	strcpy (wf3->rawfile, asn->filename);
	
	/* Set sci_* flags for wf3 */	
	wf3->sci_crcorr  = asn->crcorr;
	wf3->sci_dthcorr = asn->dthcorr;
	wf3->sci_rptcorr = asn->rptcorr;
	wf3->detector    = asn->detector;
	wf3->nimages     = 1;

	/* Set MemType appropriate for output */
	wf3->mtype[0] = '\0';

	return (status);
}
```

So, ``FindAsnRoot`` doesn't really find the actual root, i.e. dir path to file or root dir for entire path, but instead the entire thing including the filename, so i.e. it just removes the suffix (and only a certain set of suffixes at that - also they're not really file sufixes either but '_blv', '_flc' etc). In ``void FindAsnRoot (const char * input, char * output);`` ``char * input`` should be ``const char * input``.

Anyway, ``outroot`` isn't actually used other than in a print statement so should be removed or at least made just a reference. Secondly, a copy is made of it within ``FindAsnRoot()``.

In addition, ``wf3::outroot`` is actually set to ``rootname`` ([here](https://github.com/spacetelescope/hstcal/blame/master/pkg/wfc3/calwf3/calwf3/wf3member.c#L110)) and not ``outroot`` as it probbaly should.

```c
strcpy (wf3->outroot, rootname);
```

So, even with this all fixed, it then aborts on the copy of ``strcpy (wf3->rootname, rootname);`` because of the length of ``WF3Info::rootname[SZ_CBUF+1]``. This should either be the length of ``SZ_LINE+1`` as ``WF3Info::outname`` is (it is in [ACSInfo](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/include/acsinfo.h#L27)) _OR_ ``FindAsnRoot()`` really is wrong and really should be finding the canonical notion of a file root(?).

I have no idea why this is not aborting in the terminal, especially since``--fstack-protector-all`` is set at compile time.

All of the above is true for ACS code also.

jamienoss:
fixes #150

### DIFF
--- a/pkg/wfc3/calwf3/calwf3/calwf3.h
+++ b/pkg/wfc3/calwf3/calwf3/calwf3.h
@@ -34,7 +34,7 @@ typedef struct {
 	char dthfile[CHAR_LINE_LENGTH+1];	/* dither combined science data */
 	char mtype[SZ_FITS_VAL+1];	/* Role of exposure in association */
 
-	char rootname[SZ_CBUF+1];	/* root name for set of obs */
+	char rootname[CHAR_LINE_LENGTH+1];	/* root name for set of obs */
 
 
 	/* info about input science file */
--- a/pkg/wfc3/calwf3/calwf3/wf3member.c
+++ b/pkg/wfc3/calwf3/calwf3/wf3member.c
@@ -27,10 +27,9 @@ WF3Info *wf3		o: exposure specific flags and info
 	extern int status;
 	
 	char rootname[CHAR_FNAME_LENGTH+1];
-	char outroot[SZ_CBUF+1];
 	char mtype[SZ_FITS_VAL+1];
 	int  mlen;
-	void FindAsnRoot (char *, char *);
+	void FindAsnRoot (const char *, char *);
 	void UpperAll (char *, char *, int);
     int MkName(char *, char *, char *, char *, char *, int);
     
@@ -44,10 +43,9 @@ WF3Info *wf3		o: exposure specific flags and info
 
 	/* Initialize local strings */
 	rootname[0] = '\0';
-	outroot[0]  = '\0';
 	mtype[0]    = '\0';
 
-	strcpy (outroot, asn->product[prodid].subprod[posid].name);
+	const char * outroot = asn->product[prodid].subprod[posid].name;
 	
 	/* MAKE SURE WE ARE ONLY PASSING A ROOTNAME, AND NOT A FULL FILENAME.*/
 	FindAsnRoot (outroot, rootname);
@@ -101,10 +99,10 @@ WF3Info *wf3		o: exposure specific flags and info
 */
 	extern int status;
 	char rootname[CHAR_FNAME_LENGTH+1];
-	char outroot[SZ_CBUF+1];
-	void FindAsnRoot (char *, char *);
-	
-	strcpy(outroot, asn->filename);
+	*rootname = '\0';
+	void FindAsnRoot (const char * input, char * output);
+
+	const char * outroot = asn->filename;
 	
 	/* Make sure we are only passing a rootname, and not a full filename.*/
 	FindAsnRoot (outroot, rootname);
--- a/pkg/wfc3/calwf3/calwf3/wf3table.c
+++ b/pkg/wfc3/calwf3/calwf3/wf3table.c
@@ -183,7 +183,7 @@ int SetInput (AsnInfo *asn) {
     int  DoesFileExist (char *);
     char *lowcase (char *, char *);
     int  GetAsnName (char *, char *);
-    void FindAsnRoot (char *, char *);
+    void FindAsnRoot (const char *, char *);
 
     /* Initialize internal variables here... */
     filename[0] = '\0';
--- a/pkg/wfc3/calwf3/include/wf3info.h
+++ b/pkg/wfc3/calwf3/include/wf3info.h
@@ -104,7 +104,7 @@ typedef struct {
     char input[CHAR_LINE_LENGTH+1];	   /* input image to be calibrated */
     char output[CHAR_LINE_LENGTH+1];        /* output calibrated image */
 
-    char rootname[SZ_CBUF+1];      /* root name for set of obs */
+    char rootname[CHAR_LINE_LENGTH+1];      /* root name for set of obs */
 
     /* command-line flags */
     int printtime;                  /* print time after each step? */
--- a/pkg/wfc3/calwf3/lib/findroot.c
+++ b/pkg/wfc3/calwf3/lib/findroot.c
@@ -6,7 +6,7 @@
 
 # define NSUF	16
 
-void FindAsnRoot (char *input, char *root) {
+void FindAsnRoot (const char *input, char *root) {
 
 	int i;
 	int in_len;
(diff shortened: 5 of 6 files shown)


jamienoss:
[Pandokia regression test results.](https://ssb.stsci.edu/pandokia/pandokia.cgi?query=treewalk&host=%25&project=hstcal&test_run=jnoss-issue150-2017-06-26-15-38-14)

jamienoss:
@pllim & @sosey please review.

pkg/wfc3/calwf3/calwf3/calwf3.h
```diff
@@ -34,7 +34,7 @@ typedef struct {
 	char dthfile[SZ_LINE+1];	/* dither combined science data */
 	char mtype[SZ_FITS_VAL+1];	/* Role of exposure in association */
 
-	char rootname[SZ_CBUF+1];	/* root name for set of obs */
+	char rootname[SZ_LINE+1];	/* root name for set of obs */
```
jamienoss: This is the core fix.

jamienoss:
@sosey & @pllim Hold on the review, please. I just looked back over this and realized that this bug is also present in the function above, ``GetAsnMember``. I'm going to add this to this PR.

jamienoss:
@sosey & @pllim code update, please review. I will be re-running the regression test once the daily one has completed.

pllim:
@sosey is on vacation.

jamienoss:
Ok thanks.

@pllim  The bigger design changes (``ACS_FNAME 162 -> 255``) are in calcas anyhow, the rest is really just a mirror. If you can review the ACS changes and then also take a look at the WFC3 for any inconsistencies that would be great, thanks. I'll merge after that.

pkg/wfc3/calwf3/calwf3/wf3member.c
```diff
@@ -43,10 +42,9 @@ WF3Info *wf3		o: exposure specific flags and info
 
 	/* Initialize local strings */
 	rootname[0] = '\0';
-	outroot[0]  = '\0';
 	mtype[0]    = '\0';
 
-	strcpy (outroot, asn->product[prodid].subprod[posid].name);
+	const char * outroot = asn->product[prodid].subprod[posid].name;
```
pllim: Changing value of `ACS_FNAME` seems straight-forward enough but what is the reason behind this? If I read this correctly, you are not copying value over anymore, but assign a pointer to `.name`; so theoretically, if `.name` changes or `asn` is removed from memory, then `outroot` points to garbage. Did I read it right?

Best to wait till Megan come back and have a look at this.

jamienoss: The reason for making both ``*_LINE`` & ``*_FNAME`` the same is that the seem to be interchanged throughout the code when they shouldn't be. Having these be identical is the fastest fix (or now).

``FindAsnRoot()`` makes a copy anyhow and then any further persistence use is also copied as it would have to be even if first copied to ``char outroot[SZ_FNAME]``. Furthermore, in this case, there is no further use of ``outname`` beyond ``FindAsnRoot()``. I have to say I'm confused, how exactly is it that you think the thing that ``const char * outname`` points to is going to 'disappear'? If this were true then even the original ``strcpy`` is in jeopardy, regardless of this change.

pllim: If you thought about it, then it's fine by me. But I would still want to wait and see if Megan has anything to add when she returns.

If you want the `SZ_LINE` stuff to go in 2017.2a now for testing, I would suggest you split these changes into 2 PRs.

jamienoss: The idea behind a code review is that the reviewer also thinks about it - otherwise what is the point? This change is also in calacs, in identical code, if you have a concern here you should have an identical concern there - do you?

jamienoss: BTW I'm __not__ saying that we don't wait for Megan to review it also, just that you should be able to review this and that your comment:
> If you thought about it, then it's fine by me.

leads me to believe that you haven't been able to review it and therefore makes me question what you have reviewed on the ACS side of changes.

jamienoss: Though good point/idea about this going into 2017.2a. I hadn't intended for it to but perhaps it should.

jamienoss: _c.f._ #163

jamienoss:
I rebased this with master. It now includes the ACS equivalent changes from #163 making this PR WFC3 changes only.

jamienoss:
@sosey, hey, I was just wondering if you had an ETA for possibly getting around to reviewing this, please? It's kinda holding up #130

sosey:
@jamienoss I don't have time to look at this today and will get back to it monday. #130 is not approved by the science team and will not be merged.

jamienoss:
@sosey awesome, thank you.

jamienoss:
@sosey did you manage to look at this?

sosey:
No, I have other priorities, will get to it when I can

PR #170, opened 2017-07-06
ACS "unrecognised option" error should also print syntax

### LINKED ISSUE #169: ACS "unrecognised option" error should also print syntax
The following code present in [pkg/acs/calacs/acscte/maincte.c](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/acscte/maincte.c#L157) and [pkg/acs/calacs/calacs/acsmain.c](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/calacs/acsmain.c#L113):

```c
...
printf ("Unrecognized option %s\n", argv[i]);
exit (ERROR_RETURN);
...
```

should also print syntax usage to aid user correction.

_Note: There are multiple places in the command line option parsing where this is also true. This issue should be expanded to include all such instances throughout hstcal. c.f. issue #171 ._

jamienoss:
resolves #169

Note: This also corrects the acscte syntax listing 'output' as optional
i.e. [output].

### DIFF
--- a/pkg/acs/calacs/acscte/maincte.c
+++ b/pkg/acs/calacs/acscte/maincte.c
@@ -30,6 +30,11 @@ static void FreeNames (char *, char *, char *, char *);
 
  */
 
+static void printSyntax()
+{
+    printf ("syntax:  acscte.e [-t] [-v] [-q] [-1|--nthreads <N>] [--ctegen <1|2>] [--pctetab <path>] input [output]\n");
+}
+
 int main (int argc, char **argv) {
 
     char *inlist;		/* input blv_tmp file name */
@@ -155,6 +160,7 @@ int main (int argc, char **argv) {
                 if (argv[i][1] == '-')
                 {
                     printf ("Unrecognized option %s\n", argv[i]);
+                    printSyntax();
                     exit (ERROR_RETURN);
                 }
                 for (j = 1;  argv[i][j] != '\0';  j++) {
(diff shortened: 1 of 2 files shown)


jamienoss:
@pllim can you take a quick look at this, please?

pkg/acs/calacs/acscte/maincte.c
```diff
@@ -30,6 +30,11 @@ static void FreeNames (char *, char *, char *, char *);
 
  */
 
+static void printSyntax()
+{
+    printf ("syntax:  acscte [-t] [-v] [-q] [-1|--nthreads <N>] [--ctegen <1|2>] [--pctetab <path>] input [output]\n");
```
pllim: It is actually `acscte.e`

jamienoss: Cool, will update...

jamienoss:
@pllim change made, back to you.

pllim:
Actually looking at this more, I wonder if it is simpler to just assign the syntax string to a `char` array and then print that, instead of calling a whole new `printSyntax` function.

jamienoss:
That is an option I had thought of, however, for future proofing i.e. maintenance, pulling this out is better as it keeps future changes to the syntax self-contained. It creates a platform to allow for changes with a minimum of ease.

The only real argument for your case would be if it were ever needed other than to be _just_ printed for the syntax usage. In which case, I would do both - make it a ``const char*`` but still have the ``printSyntax()`` function. I cannot think of any use for it outside of this simple exe usage print statement. Is there one?

PR #173, opened 2017-07-10
Update SINKCORR keyword and HISTORY when done

### LINKED ISSUE #172: SINKCORR not set to COMPLETE even when successful
Header still says `SINKCORR=PERFORM` even when trailer file suggests otherwise below:

```
SINKCORR PERFORM
SNKCFILE /Users/ryon/Documents/sinkpix/test_mri/reference_2015-06-03.fits
SNKCFILE PEDIGREE=INFLIGHT 03/06/2015 01/07/2015
SNKCFILE DESCRIP =Sink pixel reference file ----------------------------
Ratio of (1,1) with offset =(0,0)
SINK image and input are the same size
Sink pixels flagged = 133237
Ratio of (1,1) with offset =(0,0)
SINK image and input are the same size
Sink pixels flagged = 124002
SINKCORR COMPLETE
Writing out image with trimx = 24,24, trimy = 0,20
Outputting chip 2
Writing out image with trimx = 24,24, trimy = 20,0
Outputting chip 1


End      01-Mar-2017 14:16:29 EST


*** ACSCCD complete ***
```

The fix should be patched for 2017.2a delivery. Noticed by Lisa Sherbert, reported by @dborncamp, and confirmed by @jryon.

pllim:
Fix #172 

Follow-up of #64

Update -- Output with this fix will have the following in its primary header when SINKCORR is applied successfully:

```
SINKCORR= 'COMPLETE'           /                  
SNKCFILE= 'xxxxxxxx_snk.fits'     
...
HISTORY SINKCORR complete ...                                                   
HISTORY   reference image xxxxxxxx_snk.fits                                        
HISTORY     INFLIGHT 27/07/2016 22/08/2016                                      
HISTORY     Sink pixel reference file ---------------------------- 
```

### DIFF
--- a/pkg/acs/calacs/acsccd/doccd.c
+++ b/pkg/acs/calacs/acsccd/doccd.c
@@ -69,6 +69,7 @@ int DoCCD (ACSInfo *acs_info) {
     int doNoise (ACSInfo *, SingleGroup *, int *);
     int noiseHistory (Hdr *);
     int SinkDetect (ACSInfo *, SingleGroup *);
+    int sinkHistory(ACSInfo *, Hdr *);
     int GetACSGrp (ACSInfo *, Hdr *);
     int OmitStep (int);
     int PutKeyFlt (Hdr *, char *, float, char *);
@@ -438,6 +439,10 @@ int DoCCD (ACSInfo *acs_info) {
         if (acs_info->printtime)
             TimeStamp("SINKCORR complete", acs_info->rootname);
     }
+
+    if (!OmitStep(acs[0].sinkcorr))
+        if (sinkHistory(&acs[0], x[0].globalhdr))
+            return (status);
     /************************************************************************/
 
     /* Write this imset to the output file.  The
--- a/pkg/acs/calacs/lib/acshist.c
+++ b/pkg/acs/calacs/lib/acshist.c
@@ -13,10 +13,10 @@
  "complete" or "skipped"), the value of the pedigree keyword, and the
  value of the descrip keyword.  The calibration switch (e.g. FLATCORR)
  in the header will be updated from PERFORM to either COMPLETE or SKIPPED.
- 
+
  Note that since this info is written to the primary header, these
  routines should only be called for EXTVER = 1.
- 
+
  Warren Hack, 1998 June 10:
  Initial ACS version.
  Warren Hack, 2000 Sept 12:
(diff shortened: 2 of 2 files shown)


jamienoss:
On the case.

_Note: best viewed with ``?w=1`` appended to the url._

pllim:
No time to test locally today. I'll do it tomorrow. Don't merge yet.

jamienoss:
Roger that. I'd always ask before finally merging anyhow.

pllim:
@jamienoss , if you don't see anything shocking here, please approve so we can merge this soon. Thanks!

pkg/acs/calacs/lib/acshist.c
```diff
@@ -160,39 +162,63 @@ int blevHistory (ACSInfo *acs, Hdr *phdr, int done, int driftcorr) {
...
-  
+
 	if (hstio_err())
     return (status = HEADER_PROBLEM);
-  
+
 	return (status);
 }
 
+int sinkHistory (ACSInfo *acs, Hdr *phdr) {
```
jamienoss: ``*acs`` should be ``const``. I think ``*phdr`` should be also but this would require a bit of a refactor so leave ``*phdr`` for now.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -438,6 +439,10 @@ int DoCCD (ACSInfo *acs_info) {
         if (acs_info->printtime)
             TimeStamp("SINKCORR complete", acs_info->rootname);
     }
+
+    if (!OmitStep(acs[0].sinkcorr))
```
jamienoss: Since ``sinkhistory()`` calls ``OmitStep()`` this outer if should be removed.

jamienoss: I would have also changed line 431 ``if (acs->sinkcorr == PERFORM)`` to call ``OmitStep()``, which should probably be made inline also. Given the hastiness for this fix, leave this as is.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -438,6 +439,10 @@ int DoCCD (ACSInfo *acs_info) {
...
+        if (sinkHistory(&acs[0], x[0].globalhdr))
```
jamienoss: This should really be ``if ((status = sinkHistory(&acs[0], x[0].globalhdr)))``. I'm not convinced that ``status`` is as global as it needs to be and either way, we shouldn't be using a global for this.

Same goes for all the others in this PR.

pkg/acs/calacs/lib/acshist.c
```diff
@@ -160,39 +162,63 @@ int blevHistory (ACSInfo *acs, Hdr *phdr, int done, int driftcorr) {
...
+    if (UpdateSwitch ("SINKCORR", acs->sinkcorr, phdr, &logit))
+        return (status);
+
+    /* Write history records for the bias image. */
+    if (logit) {
+        if (ImgHistory (&acs->sink, phdr))
+            return (status);
+    }
+
+    return (status);
```
jamienoss: This should be ``return (HSTCAL_OK)``

pllim:
@jamienoss , the changes you suggested are out of scope for this PR, as they affect the entire CALACS, not just SINKCORR. I simply copied the code from another step and modified it for SINKCORR. I suggest we address any general code clean-up as a separate issue outside 2017.2a delivery. To only clean up SINKCORR now but not other steps would not make much sense.

jamienoss:
@pllim There are some comments that I have made that refer to changes outside of the scope of this PR but I have explicitly stated that this is true and stated not to fix them. The other changes, however, are perfectly within the scope of this PR and do not affect the entirety of CALACS as you suggest.

pllim:
Thanks. Much appreciated! :smile:

pllim:
For the dev tests to pick this up, this needs an "upmerge" to `master`.

jamienoss:
No probs. Upmerged also.

jamienoss:
Hmmm, this didn't auto close the issue. I think I had this problem before and attributed it to using an upper case 'F' in fix. Just an FYI.

The [git page](https://help.github.com/articles/closing-issues-via-commit-messages/) doesn't seem to note anything about case sensitivity, hmmm, a bug perhaps?

pllim:
Huh, that is weird it did not auto-close. I'll keep an eye out next time. Thanks.

Issue #181, opened 2017-07-24
It should be possible to expand wildcards in file names


philhodge:
It is currently possible for the user to specify a comma-separated list of input or output file names, and the c_imtopen function returns a pointer which is then used for iterating over the specified file names, returning one at a time.  The specified names must be complete file names, however.  There should be an option to allow wildcards in these names.

Issue #179, opened 2017-07-19
Allow batch processing of HSTCAL 


jamienoss:
It would be great if it were possible to process multiple raw files from a single call, for example, to calacs.

Global variables and memory leaks may be a road blocker.

pllim:
Technically, you already can but you need to provide the ASN file. And they will be done serially, not parallel.

jamienoss:
That is not quite what I had in mind. What about the CR rejection, wouldn't the ASN method cause the DQ array to differ from that when ran singularly?

pllim:
That depends on how the ASN is defined. I don't think it runs ACSREJ if not requested. The ACS Data Handbook (particularly Chapter 3) should tell you for sure.

jamienoss:
Ok, cool, thanks, I'll take a look. If it's easy to construct an ASN file/table from a list of file names perhaps the only thing to do would be to have ASN files processes the child files in parallel. Though it would be nice to just be able to do something like ``calacs *raw.fits``. Maybe add ``-b`` for batch processing to express explicitness and perhaps even ``-r`` to recurse - the latter perhaps having hurdles.

pllim:
It is not that hard to do this in Python using `multiprocessing` and `subprocess` modules. I would recommend implementing this as a Python wrapper under ACSTOOLS rather than going pure C. It will be much easier to maintain.

pllim:
Although... Does `multiprocessing` play well with the OpenMP stuff in CTE? It seems overly complicated to multiprocess something that already generates multiple threads (albeit for only one of the steps).

jamienoss:
Yeah, that is an issue, one that could at least be worked around in the same manner that operations do so, by using only a single thread ``-1``. For ``calacs`` you can use ``--nthreads <n>`` also - I should add this to ``calwf3`` independently from the new implementation in case it is not taken.

There are alternative methods too, so as to be able to specify the number of threads to spawn jobs (instances of calxxx) and then also the number of child (nested) threads that each job can themselves spawn. That maybe something for later.

jamienoss:
_c.f._ #181 

[...]

PR #182, opened 2017-07-24
Expand wildcards in file names.

philhodge:
c_imt.c was modified to allow the use of wildcards in file names by
including glob.h and using its glob function.
See hstcal issue #181.

### DIFF
--- a/tables/c_imt.c
+++ b/tables/c_imt.c
@@ -1,23 +1,33 @@
+#if defined (__unix__) || (defined (__APPLE__) && defined (__MACH__))
+#  include <glob.h>
+#  define WXYZ_ZYXW_Unix
+#endif
 # include <stdio.h>
 # include <stdlib.h>
 # include <string.h>
 # include <fitsio.h>
 # include "ctables.h"
 
-/* Image file name template descriptor.
-   NOTE:  This version handles only explicit file names, no wildcards.
+# define DELTA_MAXFILES 100
+
+/* Expand a file name template.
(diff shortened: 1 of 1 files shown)


pllim:
I added one person per instrument to review this. Others can review too if they want to even if not assigned.

jamienoss:
Sorry for not yet having reviewed this: I will take a look soon. My reservation deterring me from doing so is the following:

Whilst it is already possible (assuming this functionality works) to give a comma delimited list of input files to 'batch' process, I doubt that this is used, perhaps at all. I think though, that a wildcard ('*') filename expansion would be. My greatest concern with this issue/functionality is the iteration of the code at the root level - is this safe?! I feel that a fair bit of attention would be required to fulfill this feature request, e.g. fix all memory leaks, clean-up/reset all global variables etc.

philhodge:
A comma-separated list is useful when you want to process a few files but not all that might be picked up by using a wildcard.

A wildcard would be especially useful when running a calxxx routine on a large set of files for a regression test, something that is done every time there's a new delivery to DMS.  If there are memory leaks or other such problems, it would be nice to find them.

The cosmic-ray rejection functions are often used with a few dozen input files, sometimes more.

jamienoss:
> A wildcard would be especially useful when running a calxxx routine on a large set of files for a regression test, something that is done every time there's a new delivery to DMS. If there are memory leaks or other such problems, it would be nice to find them.

Exactly, the regression test scenario is exactly the reason that I opened #179

Issue #193, opened 2017-08-04
Add options --version and --gitinfo where missing


jamienoss:
_c.f._ #192

#194 does this other than for those below, which require more work than the others to add these options.

 ### To do:

 * wf3rej.e
 * acsrej.e
 * cs2.e
 * `cs6.e

_c.f._ #195

pllim:
@jhunkeler has implemented this.

PR #194, opened 2017-08-04
Add git tag version, branch, & commit to waf configure and into C code

### LINKED ISSUE #192: Build git tag, branch, & commit into code and print at runtime
E.g.

```
HSTCAL
git tag: 1.3.0-17-gfbf1650f-dirty
git branch: issue192-inc-git-version-and-hash-in-code
HEAD @: fbf1650f5836ad6a16d81aef3a90c61062f74d13
```

The idea behind this to significantly aid in testing such that the file output records git versioning, i.e. adding this to the build string is not enough, we would have to then add that build string into the code - this effectively does just that. 

A typical configure output may look like:
```
Checking for WAF			:  /Users/jnoss/dev/hstcal/waf
Setting top to                           : /Users/jnoss/dev/hstcal 
Setting out to                           : /Users/jnoss/dev/hstcal/build.Darwin-15.6.0-x86_64-i386-64bit 
Building app                             : HSTCAL 
Version                                  : 1.3.0-16-gc659f43-dirty 
git branch                               : issue192-inc-git-version-and-hash-in-code 
git HEAD commit                          : c659f4371d71a2628180d8313ea4045774f694af 
Checking for 'gcc' (C compiler)          : /Users/jnoss/software/gcc-7.1.0-INSTALLATION/bin/gcc 
Checking for 'gfortran' (Fortran compiler) : /Users/jnoss/software/gcc-7.1.0-INSTALLATION/bin/gfortran 
Compiling a simple fortran app             : yes 
Checking Mac OS-X version                  : done 
Checking for OpenMP                        : OpenMP found. 
Checking for sizeof(int)                   : yes 
Checking for compiler flags -O2            : yes 
Checking for compiler flags -Wall          : yes 
Checking for compiler flags -fstack-protector-all : yes 
C compiler flags (CFLAGS)                         : -D APPNAME="HSTCAL" -D VERSION="1.3.0-16-gc659f43-dirty" -D BRANCH="issue192-inc-git-version-and-hash-in-code" -D COMMIT="c659f4371d71a2628180d8313ea4045774f694af" -O2 -Wall -fstack-protector-all 
Fortran compiler flags (FCFLAGS)                  : -m64 
Linker flags (LDFLAGS)                            : None 
Using cfitsio                                     : local 
Checking for program 'make'                       : /sw/bin/make 
Copying cfitsio                                   : done 
Configuring cfitsio                               : done 
'configure' finished successfully (3.087s)
```
having the principle additions:
```
Building app                             : HSTCAL 
Version                                  : 1.3.0-16-gc659f43-dirty 
git branch                               : issue192-inc-git-version-and-hash-in-code 
git HEAD commit                          : c659f4371d71a2628180d8313ea4045774f694af 
```

jamienoss:
resolves #192 

### DIFF
--- a/pkg/acs/calacs/acssum/mainsum.c
+++ b/pkg/acs/calacs/acssum/mainsum.c
@@ -10,7 +10,9 @@ int status = 0;			/* zero is OK */
 
 # include "acs.h"
 # include "acssum.h"
+# include "acsversion.h"
 # include "hstcalerr.h"
+# include "hstcalversion.h"
 
 /* 
     This function will only return either 0 (ACS_OK) if everything
@@ -50,20 +52,30 @@ int main (int argc, char **argv) {
 	/* Get names of input and output files. */
 	for (i = 1;  i < argc;  i++) {
 	    if (argv[i][0] == '-') {
-		for (j = 1;  argv[i][j] != '\0';  j++) {
-		    if (argv[i][j] == 't') {
-			printtime = YES;
-		    } else if (argv[i][j] == 'v') {
-			verbose = YES;
-		    } else if (argv[i][j] == 'q') {
-			quiet = YES;
-		    } else {
-				printf (MsgText, "Unrecognized option %s\n", argv[i]);
-				free (input);
-				free (output);
-				exit (1);
-		    }
-		}
+            if (!(strcmp(argv[i],"--version")))
+            {
+                printf("%s\n",ACS_CAL_VER);
+                exit(0);
+            }
+            if (!(strcmp(argv[i],"--gitinfo")))
+            {
+                printGitInfo();
+                exit(0);
+            }
+			for (j = 1;  argv[i][j] != '\0';  j++) {
+				if (argv[i][j] == 't') {
+				printtime = YES;
+				} else if (argv[i][j] == 'v') {
+				verbose = YES;
+				} else if (argv[i][j] == 'q') {
+				quiet = YES;
+				} else {
+					printf (MsgText, "Unrecognized option %s\n", argv[i]);
+					free (input);
+					free (output);
+					exit (1);
+				}
+			}
 	    } else if (input[0] == '\0') {
 		strcpy (input, argv[i]);
 	    } else if (output[0] == '\0') {
@@ -73,7 +85,7 @@ int main (int argc, char **argv) {
 	    }
 	}
 	if (input[0] == '\0' || too_many) {
-	    printf ("syntax:  acssum [-t] [-v] [-q] input [output]\n");
+	    printf ("syntax:  acssum [-t] [-v] [-q] [--version] [--gitinfo] input [output]\n");
 		free (input);
 		free (output);
 		exit (ERROR_RETURN);
@@ -81,7 +93,8 @@ int main (int argc, char **argv) {
 
 	/* Initialize the structure for managing trailer file comments */
 	InitTrlBuf ();
-	
+    trlGitInfo();
+
 	/* Copy command-line value for QUIET to structure */
 	SetTrlQuietMode(quiet);
 
--- a/wscript
+++ b/wscript
@@ -1,6 +1,6 @@
 # vim: set syntax=python:
 
-import os, platform, shutil, sys
+import os, platform, shutil, sys, subprocess
 
 from waflib import Configure
 from waflib import Errors
@@ -11,8 +11,7 @@ from waflib import Task
 from waflib import Utils
 from waflib import TaskGen
 
-APPNAME = 'hstcal'
-VERSION = '0.1.1'
+APPNAME = "HSTCAL"
 
 top = '.'
 out = 'build.' + platform.platform()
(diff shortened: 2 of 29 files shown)


jamienoss:
@stsci-hack, @mdlpstsci, @pllim, & @sosey FYI

wscript
```diff
@@ -84,6 +86,54 @@ def _setup_openmp(conf):
...
+def _err_color(var, val):
+    if var == val:
+        return "RED"
+    else:
+        return "GREEN"
+
+def _get_git_details(conf):
+    global APPNAME, VERSION, COMMIT, BRANCH
+
+    results = subprocess.run('git describe --dirty', shell=True, check=False, universal_newlines=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
pllim: @jhunkeler , this adds `git` as a dependency to the installation process, so astroconda recipe might need to be adjusted accordingly. FYI.

jamienoss: It doesn't, if the subprocess errors upon git not existing or not being in a git dir, these globals will remain as ``"UNKNOWN"`` and thus be printed so (unless I have messed something up, that is).

pllim: Yeah, but in that case, we cannot trace the git info. Then, astroconda users will not get anything meaningful out of this feature.

jamienoss: Then they/we should make sure it gets build from a git dir. Perhaps you are suggesting that we explicitly enforce git as a dependency, rather than it being a fact from the code in this PR (which it is not) - the two are very different things.

I don't believe we should be making git a dependency, I'm not even sure how that works or, semantically, what it even means.

jamienoss:
@pllim 

>If I understand the changes correctly, this adds the info to the execution log. However, I don't see a way to inspect the version without actually running CALACS; for example calacs.e --version.

>Note that we should not change --version behavior, but rather add a new argument (say, --hstcalversion or something) to print this value to screen.

>IMHO that would be very useful, especially when delivering experimental software to instrument teams for testing.

Good idea!

wscript
jamienoss: Isn't that like saying we should make the source code a dependency?

pllim: Hmm, I guess astroconda already has git; Good point -- Maybe Joe can confirm.

But what about other people who want to build HSTCAL from source? How do you specific an equivalent to "extras_require" in `setup.py` for Python packages? :thinking:

jamienoss:
Doh - python 2.7

wscript
jamienoss: Sorry, I'm really not sure what you are on about since you can _only_ build from source and this PR only involves building from source.

What python packages (in HSTCAL?) or are you wondering if we can do the same thing for the wrappers or our python tools?

jhunkeler: @plim Even though it is rather likely `git` will exist at build-time, I'd simply add it to the recipe.
```yaml
build:
     - git
```

The Conda package run-time wouldn't need to change. As for people compiling from source without Conda... It's up to them to have a working environment with `git` available on their `PATH`. Presumably they used `git` to obtain the code anyway, so this is probably a non-issue all around :)

jamienoss:
@jhunkeler - new commit pushed to deal with Python 2.7. I changed the print format to something neater.

jamienoss:
FYI: I've just realized that this is not bulletproof as it is only in the configure script and not the build. Therefore, you could configure, change the code, commit (or not) and rebuild, and the git details will not have changed thus being inaccurate.

Suggestions welcome...

jamienoss:
@jhunkeler (if you're still up like I am) hold off on the new review, I'm making changes to accommodate Pey-Lian's good ``--version`` suggestion.

I'm going for ``--git``. It doesn't sit too well with me so suggestions would be great. ``--gitinfo`` maybe...?

jamienoss:
@jhunkeler fire at will.

pllim:
@jamienoss , `--gitinfo` would be more self-explanatory.

@jhunkeler , your explanation regarding build requirements is satisfactory. There is always a person who would prefer to download the tarball instead of using `git` (a real life case I have encountered before) but it is not a deal breaker here.

Thanks, all!

jhunkeler:
# Passing --git to all programs in suite

I understand this may not apply to every single program at the moment, but it's worth noting for future reference.


```
$ ./waf configure --prefix=$(pwd)/_install && ./waf build && ./waf install
$ for exe in _install/bin/*; do output=$($exe --git); retval=$?; printf "\`\`\`\n$exe:\n\n$output\n(returned: $retval)\n\`\`\`\n\n"; done
```

```
_install/bin/acs2d.e:

Unrecognized option --git
(returned: 1)
```

```
_install/bin/acsccd.e:


(returned: 1)
```

```
_install/bin/acscte.e:

Unrecognized option --git
syntax:  acscte.e [-t] [-v] [-q] [-1|--nthreads <N>] [--ctegen <1|2>] [--pctetab <path>] input [output]
(returned: 2)
```

```
_install/bin/acsrej.e:

Syntax error: acsrej input output [-t] [-v]
                    [-shadcorr] [-crmask] [-newbias]
                     [-table <filename>] [-scale #]
                    [-init (med|min)] [-sky (none|mode)]
                    [-sigmas #] [-radius #] [-thresh #]
                    [-pdq #]
(returned: 2)
```

```
_install/bin/acssum.e:


(returned: 1)
```

```
_install/bin/calacs.e:

git tag: 1.2.0rc2-42-g4624437d
git branch: issue192-inc-git-version-and-hash-in-code
HEAD @: 4624437d692648fc0303e59170f88a66ae577f5a
(returned: 0)
```

```
_install/bin/calwf3.e:

git tag: 1.2.0rc2-42-g4624437d
git branch: issue192-inc-git-version-and-hash-in-code
HEAD @: 4624437d692648fc0303e59170f88a66ae577f5a
(returned: 0)
```

```
_install/bin/cs0.e:

Unrecognized option --git
(returned: 2)
```

```
_install/bin/cs11.e:

ERROR:  Unrecognized option --git
(returned: 1)
```

```
_install/bin/cs12.e:

ERROR:  Unrecognized option --git
(returned: 1)
```

```
_install/bin/cs1.e:

ERROR:  Unrecognized option --git
(returned: 1)
```

```
_install/bin/cs2.e:

ERROR  Syntax error: cs2 input output
(returned: 1)
```

```
_install/bin/cs4.e:

ERROR:  Unrecognized option --git
(returned: 1)
```

```
_install/bin/cs6.e:

Error processing --git.
         HSTIO error 114:  Filename --git EXTNAME  EXTVER 0 CFITSIO status 104
failed to find or open the following file: (ffopen)--git[0]
Error opening image array.
         Open failed.
(returned: 2)
```

```
_install/bin/cs7.e:

ERROR:  Unrecognized option --git
(returned: 1)
```

```
_install/bin/wf32d.e:

git tag: 1.2.0rc2-42-g4624437d
git branch: issue192-inc-git-version-and-hash-in-code
HEAD @: 4624437d692648fc0303e59170f88a66ae577f5a
(returned: 0)
```

```
_install/bin/wf3ccd.e:

git tag: 1.2.0rc2-42-g4624437d
git branch: issue192-inc-git-version-and-hash-in-code
HEAD @: 4624437d692648fc0303e59170f88a66ae577f5a
(returned: 0)
```

```
_install/bin/wf3cte.e:


(returned: 2)
```

```
_install/bin/wf3ir.e:

git tag: 1.2.0rc2-42-g4624437d
git branch: issue192-inc-git-version-and-hash-in-code
HEAD @: 4624437d692648fc0303e59170f88a66ae577f5a
(returned: 0)
```

```
_install/bin/wf3rej.e:


(returned: 139)
```

```
_install/bin/wf3sum.e:


(returned: 1)
```

jhunkeler:
Trapping the return value didn't handle the segfault correctly above. `wf3rej.e` throws a SIGSEGV.

```
$ _install/bin/wf3rej.e --git
git tag: 1.2.0rc2-42-g4624437d
git branch: issue192-inc-git-version-and-hash-in-code
HEAD @: 4624437d692648fc0303e59170f88a66ae577f5a
Segmentation fault (core dumped)
```

jamienoss:
Cheers @jhunkeler your tests actually expose a few pre-existing bugs. So just to clarify, I only added ``--git`` to those that already have ``--version`` - I should have extended this to all and will do so (at some point).

So a few of the tests, e.g. wf3cte.e should have displayed the syntax usage but didn't.

wf3rej.e - I didn't add ``--git`` to this. The segfault looks like a pre-existing one, it does the same thing with ``--version`` or no options/args at all, i.e. just ``wf3rej.e``. The reason that it still prints the git info is that it is actually executing ``main()`` beyond the arg parsing and thus hits ``printGitIfno()``. Actually, having just looked at the code ``mainreg.c``, there are no args/parsing at all.

jhunkeler:
Roger that. Aside from the unrelated oddities the implementation looks good to me. Did you want to change the `--git` argument to be `--gitinfo` (`--git-info` maybe)? I think we accidentally skipped over @pllim's suggestion thanks to my giant wall of text.

jamienoss:
@jhunkeler Sorry, I had done so but held off on pushing (just pushed now though). I'm still wondering whether I should extend this to all executables in this PR or open another.

jhunkeler:
No problem. I'm all for doing this in one shot so long as no one objects to it. For the sake of (more) reference material I threw this together a few moments ago:

https://gist.github.com/jhunkeler/7671c45aba47f595e84d47062b50bf36

The script's output is attached to the gist.

jamienoss:
@jhunkeler haha that's awesome dude, thanks!

wscript
```diff
@@ -349,12 +349,12 @@ Press any key to continue or Ctrl+c to abort...\033[0m"""
     conf.recurse('cfitsio')
 
 def build(bld):
-    diffList = []
-    if not _is_same_git_details(bld, diffList):
-        diffString = ''
-        for item in diffList:
-            diffString = diffString + item + '\n'
-        bld.fatal("ERROR: git details differ - please re-configure.\n{0}".format(diffString))
+    #diffList = []
```
jamienoss: _c.f._ #197

jamienoss:
@jhunkeler sorry dude I did some refactoring and added some more changes. Can you please review again. _c.f._ #197 & #193

pkg/acs/calacs/acssum/mainsum.c
```diff
@@ -51,20 +52,30 @@ int main (int argc, char **argv) {
...
-		    } else if (argv[i][j] == 'q') {
-			quiet = YES;
-		    } else {
-				printf (MsgText, "Unrecognized option %s\n", argv[i]);
-				free (input);
-				free (output);
-				exit (1);
-		    }
-		}
+            if (!(strcmp(argv[i],"--version")))
```
jhunkeler: Missing `[--version] [--gitinfo]`:
```
$ ./_install/bin/acssum.e 
syntax:  acssum [-t] [-v] [-q] input [output]
```

jamienoss: Awesome, cheers for spotting this. Changes updated. BTW I would branch this such that when I squash everything you can pull the squashed version, and do a simple diff as your final review before I merge.

jhunkeler:
Travis went to lunch.

jamienoss:
@jhunkeler squashed. Travis has fallen a sleep behind the wheel again...

jamienoss:
@jhunkeler squashed. Travis has fallen a sleep behind the wheel again...

PR #209, opened 2017-08-31
char MsgText[<length>] should not be globally defined from header file

jamienoss:
resolves #207

PR #225, opened 2017-09-22
Dedup CHAR_LINE_LENGTH & CHAR_FANME_LENGTH macro defs

### LINKED ISSUE #212: #209 was incomplete
A few (4) declarations of ``CHAR_FNAME_LENGTH`` & ``CHAR_LINE_LENGTH`` exist outside of ``hstcal.h``, i.e. they were not removed in #209.

These are all a non-issue other than one, declared in [ctables.h](https://github.com/spacetelescope/hstcal/blob/master/tables/ctables.h#L8):

```c
# define CHAR_FNAME_LENGTH    1024
```

The issue here is the size, it's not 255 as the others all were. Also, it's 1024 and not 1023. I checked through all usages of it in [hstcal/tables/](https://github.com/spacetelescope/hstcal/tree/master/tables) and they are all with a ``+1`` to account for ``'\0'``. I feel that, in this case, this instance should therefore actually be 1023.

Since it is unclear what is including and thus using this dec and not the one in ``hstcal.h``, due to the plague of indirect includes throughout hstcal, the only solution I can currently think of is to expands the ``hstcal.h`` dec to the size of 1023.

## Complications
There are two cases in [tables/tbl_util.c](https://github.com/spacetelescope/hstcal/blob/master/tables/tbl_util.c), [here](https://github.com/spacetelescope/hstcal/blob/master/tables/tbl_util.c#L44) and [here](https://github.com/spacetelescope/hstcal/blob/master/tables/tbl_util.c#L59) that conditionally check ``CHAR_FNAME_LENGTH``.

```c
...
if (j >= CHAR_FNAME_LENGTH) {
                            var[j] = '\0';
                            start = j + 1;
                            break;
                        }
...
```

```c
...
if (dollar >= CHAR_FNAME_LENGTH)
                        dollar = CHAR_FNAME_LENGTH;
...
```

``CHAR_FNAME_LENGTH`` could be made equal to 1024 throughout, however, with all the use cases explicitly adding the ``+1`` I'd prefer not to make all uses slightly under-performant (weak argument perhaps). Changing it to 1023 for the uses in ``tables/tbl_util.c`` _should_ be ok. The reduction of a single character would only affect those current use cases that are obviously floating too close to the limit in the first place. We could make it larger, 2048-1, but I think it best to expand this only if actually needed, i.e. bugs detected.

### Update:
Somehow the original PR commit missed (didn't add) the following direct includes of hstcal.h for ``CHAR_FNAME_LENGTH``:

hstcal/tables/c_imt.c
hstcal/tables/c_tbfpri.c
hstcal/tables/c_tbhgt.c
hstcal/tables/c_tbtcre.c
hstcal/tables/c_tbtopn.c
hstcal/tables/ctables.c
hstcal/tables/tbl_util.c

and for ``CHAR_LINE_LENGTH``

hstcal/ctegen2/ctehelpers.c

### Update:
The problematic second definition of ``#define CHAR_FNAME_LENGTH 1024`` present in ``ctables.h`` is actually a pre-existing bug. ``ctables.h`` is only included by ``.c`` files in ``tables/`` and not in any other ``.h`` files and therefore not possible to be indirectly included elsewhere. It is entirely contained within the source code in ``tables/``. It seems that this not a redefinition either as the pre-processor yields no such warnings, implying that none of the source code in ``tables/`` is including (directly or indirectly) the definition in ``hstcal.h``. The reason that this is a bug is that the ``char`` arrays being passed into code in ``tables/`` are declared and thus instantiated externally and with the macro definition from ``hstcal.h``. This is is of length 255 and the arrays declared as length 255+1 yet the length used in ``tables/*.c`` is 1024. All of this ``char`` arrays are being accessed out of bounds.

#### Is this a problem?
All bar two references of ``CHAR_FNAME_LENGTH`` in ``tables/`` are used only to define the ``char`` array lengths and as such are completely benine. The exception are the two described at the top of this description (https://github.com/spacetelescope/hstcal/blob/master/tables/tbl_util.c), [here](https://github.com/spacetelescope/hstcal/blob/master/tables/tbl_util.c#L44) and [here](https://github.com/spacetelescope/hstcal/blob/master/tables/tbl_util.c#L59). This bug is only manifested when the passed in strings are __NOT__ null terminated. In the case that they are not, the code does read beyound the array upto 1024 characters, that's ~1024-256 = 768 beyound. There is scope for this to segfault but it is unliekely as the array would have to be pretty close to the stack boundary. This is probably why this has not been found before as it has never segfaulted before.

#### Is memory corrupted?
No. ``filename`` is not written to. This would never be the case if it was made ``const`` as it should correctly be!

DOH! The function ``expandfn(char *filename)`` that contains this bug is not actually called by anything!

jamienoss:
These changes should have been contained by #209

Resolves #212

### DIFF
--- a/ctegen2/ctehelpers.c
+++ b/ctegen2/ctehelpers.c
@@ -2,6 +2,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <assert.h>
+#include "hstcal.h"
 #include "wf3.h" //need to remove this dependency
 #include "ctegen2.h"
 
--- a/include/imphttab.h
+++ b/include/imphttab.h
@@ -4,16 +4,14 @@
 # include "c_iraf.h"  /* For Bool type */
 # include "xtables.h" /* for SZ_COLNAME*/
 # include "hstcalerr.h"
-
+#include "hstcal.h"
 /* Constants and Definitions for use with IMPHTTAB library */
 /* Definitions based on those defined in acs.h */
 
 #define YES         1
 #define NO          0
 
-# define CHAR_LINE_LENGTH      255
 # define SZ_FITS_REC   82
-# define CHAR_FNAME_LENGTH     255
 
 /* Standard string for use in Error Messages */
 /*char MsgText[CHAR_LINE_LENGTH+1];*/
--- a/pkg/acs/calacs/acscte/pcte.h
+++ b/pkg/acs/calacs/acscte/pcte.h
@@ -25,7 +25,6 @@
 
 //New params needed for second gen CTE correction algorithm
 #define TRAPS 9999//6675 //max number of traps per column = rows in pctetab[1]
-# define CHAR_LINE_LENGTH           255 //from calwf3/msg.h
 
 /* structure to hold CTE parameters from reference file */
 typedef struct {
--- a/tables/c_imt.c
+++ b/tables/c_imt.c
@@ -2,6 +2,7 @@
 # include <stdlib.h>
 # include <string.h>
 # include <fitsio.h>
+#include "hstcal.h"
 # include "ctables.h"
 
 /* Image file name template descriptor.
--- a/tables/c_tbfpri.c
+++ b/tables/c_tbfpri.c
@@ -1,5 +1,6 @@
 # include <string.h>
 # include <fitsio.h>
+#include "hstcal.h"
 # include "ctables.h"
 
 void c_tbfpri (char *intable, char *outtable, int *copied) {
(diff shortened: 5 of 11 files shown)


jamienoss:
I did a source-wide file search for macro defs of ``CHAR_FNAME_LENGTH``, ``CHAR_LINE_LENGTH``, & ``MSG_BUFF_LENGTH``, the only results are the intended definitions in ``hstcal.h``, so all is good here.

I also ran [include-headers](https://github.com/jamienoss/toolchest/blob/master/include-headers.py) ``python include-headers.py hstcal.h -p <path to hstcal> -f "\.c$" --listOnly --regex CHAR_FNAME_LENGTH`` and similarly for the other two, no files were returned indicating that all direct includes are correctly accounted for.

PR #231, opened 2017-09-22
Common up trlbuf.c into hstcal/lib

### LINKED ISSUE #191: Common up trlbuf.c into hstcal/lib
### Update:
This is dependent on #207 

__Note to self:__ Remeber to change this (as done in #194)

```c
void trlmessage (char *message) {
  
      void asnmessage (char *);
  
 -    if (!message)
 +    if (!message || !*message)
          return;
```
### UPDATE:
Done.

``trlbuf`` should also not be global as it is [here](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/lib/trlbuf.c#L127).

### UPDATE:
This was moved into trlbuf.c as the following global:
```c
struct TrlBuf trlbuf = { 0 };
```

Should this be global at all?

### Update:
Done. Converted to: ``extern struct TrlBuf trlbuf ;`` and then declared within each  ``main()`` as needed.

May have to declare as extern in a few (all) places. Already done in trlbuf.c (this might be all).

### Update:
Change ``trlbuf.c::CloseTrlBuf(void)`` to reflect NULL changes made from #196 
```c
if (trlbuf.buffer)
        {
            free (trlbuf.buffer);
            trlbuf.buffer = NULL;
        }

        if (trlbuf.preface)
        {
            free (trlbuf.preface);
            trlbuf.preface = NULL;
        }
```

### Update:
Done.

### LINKED ISSUE #228: fgets() has incorrect length passed in
``char * fgets ( char * str, int num, FILE * stream );``

Documentation from [cppreference.com](http://en.cppreference.com/w/c/io/fgets)
> Reads at most ``count - 1`` characters from the given file stream and stores them in the character array pointed to by str. Parsing stops if end-of-file occurs or a newline character is found, in which case str will contain that newline character. If no errors occur, writes a null character at the position immediately after the last character written to str.
The behavior is undefined if count is less than 1.

The key points here are:

 * Reads at most ``count - 1`` characters
 * If no errors occur, writes a null character at the position immediately after the last character written to str.

The two calls of ``fgets()`` within ``hstcal/pkg/wfc3/calwf3/lib/trlbuf.c`` pass the length as ``CHAR_LINE_LENGTH`` whilst the ``str`` (``buf``) is of array length ``CHAR_LINE_LENGTH + 1``. Therefore, this would be reading one char less than needed in the case of the full number of chars being needed. FYI this is bad design - making, ``CHAR_LINE_LENGTH`` one less than optimal, e.g. 255 rather than 256 and adding the +1 needed for null termination within the object declaration or mem allocation call.

This is not the case for calacs. The two instances can be found [here](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/lib/trlbuf.c#L245) and [here](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/lib/trlbuf.c#L289).

This is fixed as part of #191

jamienoss:
Resolves #191

Note:
There were some slight differences between some of the trailer file functions in calacs and calwf3. They are listed below:

 * InitTrlFile() - port over temp file staging from calacs
 * CatTrlFile_NoEOF() - added from calacs to accommodate the above port
 * AppendTrlFile()
		   - ``char buf[CHAR_LINE_LENGTH];`` ->  ``char buf[CHAR_LINE_LENGTH+1];``
	           - ``fgets(buf, CHAR_LINE_LENGTH, trlbuf.fp);`` -> fgets(buf, CHAR_LINE_LENGTH+1, trlbuf.fp);`` Fixes #228
	           - Better pointer safety

The following functions had slight changes made to them during the commoning-up process:

 * writeTrlFile() - Better pointer safety
 * InitTrlFile() - ``trlbuf.init = 1;`` ported from calacs


jamienoss:
To review I would pay attention to trlbuf.c and the two older files that it replaces - pkg/acs/calacs/lib/trlbuf.c pkg/wf3/calwf3/lib/trlbuf.c. You'll have to checkout master and copy these two files somewhere to ref or inversely copy lib/trlbuf.c from my dev branch.

I would do a three-way compare with meld. Tip - put the new one in the middle.

jamienoss:
FYI the three commits will be merged and not squashed.

pllim:
I don't think you need to tag me for review anymore. I defer to Michele on this. Thanks.

jamienoss:
> I don't think you need to tag me for review anymore. I defer to Michele on this. Thanks.

and

> I'm no longer keeping track of this project.

Should I take this as the official position and not run future PRs by either of you, regardless of what they are? @stsci-hack?

I wasn't actually expecting either of you to review the code but also didn't want to sneak it past you, at least in case you did happen to glance at it and spot something alarming. I know that the game plan is for @mdlpstsci to take the reins, are we saying that we are already in the future?

pllim:
I watch the repo anyway, so I'll still chime in if needed even if you don't tag me as reviewer. You may also mention me with `@` if you need my input. Tagging me as reviewer implies that you want me to do a formal review on this PR. FYI.

jamienoss:
> Tagging me as reviewer implies that you want me to do a formal review on this PR. FYI.

I know. I wanted you to review it but didn't expect that you actually would/could. Relying on the fact that you _might_ be watching things, kinda defaults on my responsibility. This is why I explicitly tagged you for review. You can always defer etc but if the future is for you to not be available to review, I just need to be given the word. Anyhow...

pllim:
>  just need to be given the word

Word is given. 😄

PR #235, opened 2017-10-02
Common up trlbuf.c into hstcal/lib

### LINKED ISSUE #191: Common up trlbuf.c into hstcal/lib

### LINKED ISSUE #198: Make trlGitInfo() call trlmessage() instead of just printf()
In relation to #194 and pending on #191 the following [code](https://github.com/spacetelescope/hstcal/blob/master/lib/hstcalversion.c#L41)

```c
...
printf("%s\n", gitInfo);//trlmessage(versionText); //waiting on #191
...
```

Needs to be just:
```c
...
trlmessage(versionText);
...
```

### Update:
Whilst there, add the following to ``include/hstcalversion.h``:

```c
...
#ifndef VERSION
#define VERSION "UNKNOWN"
#endif

#ifndef BRANCH
#define BRANCH "UNKNOWN"
#endif

#ifndef COMMIT
#define COMMIT "UNKNOWN"
#endif
...
```

### LINKED ISSUE #227: Add trivial const correctness upgrades to to trlbuf.c functions
_c.f._ #191

### LINKED ISSUE #228: fgets() has incorrect length passed in

### LINKED ISSUE #230: Rename asnmessage() to printfAndFlush and common to lib/trlbuf.c
This is needed for #191

### LINKED ISSUE #233: Missing dec in hstcalversion.h
[hstcalversion.h](https://github.com/spacetelescope/hstcal/blob/master/include/hstcalversion.h) is missing the declaration for ``char * sprintfGitInfo(char ** buffer)``, [source here](https://github.com/spacetelescope/hstcal/blob/master/lib/hstcalversion.c#L8-L34).

### LINKED ISSUE #243: Fix realloc bugs in trlbuf.c
The following [bug exists](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/lib/trlbuf.c#L355-L357) in ``pkg/acs/calacs/lib/trlbuf.c``.

```c
...
 oprefix = realloc (oprefix, (strlen(oprefix) + strlen(buf) +2 ));
 if (oprefix == NULL)
     //error
...
```

The correct pattern for using ``realloc()`` is to use a tmp pointer.
```c

...
char * oprefix = NULL;
...
void * ptr = realloc (oprefix, (strlen(oprefix) + strlen(buf) +2 ));
if (!ptr)
{
    //error message
    if (oprefix)
    {
         free(oprefix); // Only free if having ownership, which this func does.
         oprefix = NULL;
    }
    return; // or whatever
}
oprefix = ptr;
ptr = NULL;
...
```

The reason for this is that when ``realloc(ptr, length)`` fails it leaves ``ptr`` as is and only returns ``NULL``. In the original case, ``oprefix`` would be set to this ``NULL`` value, overwriting the original address thus leaking it.

@stsci-hack the following [comment](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/lib/trlbuf.c#L322-L323) is present for this particular function.
```
18 Jan 2002 (WJH): Removed extraneous, bug-ridden 'tmpptr' buffer used
        during reallocation of oprefix buffer size.
```
and @hbushouse in [calwf3/lib/trlbuf.c](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/lib/trlbuf.c#L267-L268)
```
21 June 2002 (HAB): Removed extraneous, bug-ridden 'tmpptr' buffer used
	during reallocation of oprefix buffer size (following CALACS changes).
```
Are you able to explain more?

This issue is true for many of the calls to ``realloc()`` in this file and its ``WFC3`` counterpart. 

There are also issues where ``realloc()`` seems to be oddly used instead of ``malloc()``. E.g. [here](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/lib/trlbuf.c#L432)
```c
...
if ( (trlbuf.buffer = realloc (NULL, INIT_LEN)) == NULL){
...
```
It is also an issue that length is used instead of explicitly size. This is only correct if ``trlbuf.buffer`` is of type ``char``. Which it is, however, it does make this code brittle to type changes to the underlying struct members. This is true for all calls to ``realloc()``.

These are commoned by #191 and fixed by its associated PR #235 

### Update

Warren pointed me to the original PR that removed the original use of a ``tmpptr`` from ``realloc()``.
[trac issue](https://trac.stsci.edu/ssb/stsci_python/changeset/4902/stsdas/trunk/stsdas/pkg/hst_calib/acs/calacs). The original code looked like this:
```c
tmpptr = realloc (oprefix, (strlen(oprefix) + strlen(buf) +2 )); 
if (tmpptr == NULL) {
    asnmessage ("Out of memory: Couldn't store trailer file comment.");
    status = OUT_OF_MEMORY; 
    /* Clean-up... */ 
     free (trlbuf.buffer); 
    free (trlbuf.preface);
    free (oprefix); 
    fclose (trlbuf.fp); 
    return (status); 
}
else
{ 
    free(oprefix);
    oprefix = tmpptr;
    tmpptr = NULL;
...
}
...
```
The bug wasn't the use of a temp ptr, this is the correct thing to do, the bug was the ``free(oprefix);`` upon the success of the reallocation. I.e. the ``else`` branch should have looked like this:
```c
...
else
{ 
    oprefix = tmpptr;
    tmpptr = NULL;
...
}
```
The correct fix was just deleting this one line. The reason for this is that ``realloc`` will expand the address space used if it can and in this case can return you the exact same pointer that existed before the call to ``realloc``.

jamienoss:
Re-opened from #231 

Resolves #191
Fixes #198 
Fixes #227 
Fixes #228
Fixes #230 
Fixes #233 
Fixes #243 

Note:
There were some slight differences between some of the trailer file functions in calacs and calwf3. They are listed below:

 * InitTrlFile() - port over temp file staging from calacs
 * CatTrlFile_NoEOF() - added from calacs to accommodate the above port
 * AppendTrlFile()
		   - ``char buf[CHAR_LINE_LENGTH];`` ->  ``char buf[CHAR_LINE_LENGTH+1];``
	           - ``fgets(buf, CHAR_LINE_LENGTH, trlbuf.fp);`` -> fgets(buf, CHAR_LINE_LENGTH+1, trlbuf.fp);`` Fixes #228
	           - Better pointer safety

The following functions had slight changes made to them during the commoning-up process:

 * writeTrlFile() - Better pointer safety
 * InitTrlFile() - ``trlbuf.init = 1;`` ported from calacs


jamienoss:
To review I would pay attention to trlbuf.c and the two older files that it replaces - pkg/acs/calacs/lib/trlbuf.c pkg/wf3/calwf3/lib/trlbuf.c. You'll have to checkout master and copy these two files somewhere to ref or inversely copy lib/trlbuf.c from my dev branch.

I would do a three-way compare with meld. Tip - put the new one in the middle.

FYI the commits will __not__ be squashed.

jamienoss:
@mdlpstsci thanks for the review. FYI if there are code dependent remarks, these are best made as inline comments so as to be explicit.

lib/trlbuf.c
```diff
@@ -566,51 +501,39 @@ void ResetTrlPreface (void) {
...
+    /*
+        Clear out messages from buffer.  Intended for use after
+        messages were already written to file.  Assumes trlbuf.buffer
+        already exists!
+    */
 
     extern int status;
-    void asnmessage (char *);
 
     free(trlbuf.buffer);
```
jamienoss: @mdlpstsci an inline comment looks like this. Hover the cursor over the line, a blue '+' tab should appear, click on it and enter comment. It is not available for greyed-out code.

jamienoss:
The ``CALACSBEG`` refereences are only to comments and in the history section which is why I didn't change them.

lib/trlbuf.c
jamienoss: Thanks for spotting this.

jamienoss: In fact, this is quite a bug, and the ``free()`` should be removed altogether due to the subsequent ``realloc()``.

jamienoss: Actually, this is odd. It frees and then subsequently re-allocates (mallocs) via a ``realloc(NULL, size)``. I'm not sure of what the point of this could be.

mdlpstsci:
Indeed - I focused on the wrong bug!

Michele D. De La Pena
Space Telescope Science Institute
mdelapena@stsci.edu
667.218.6501

lib/trlbuf.c
jamienoss: I think this is better if the ``free(trlbuf.buffer)`` is removed and just reallocated.
```ctrlbuf.buffer = realloc(trlbuf.buffer, initLength);```

File: lib/trlbuf.c [hunk 2]
```diff
@@ -566,51 +501,39 @@ void ResetTrlPreface (void) {
...
-    trlbuf.buffer = realloc(NULL, INIT_LEN);
+    trlbuf.buffer = realloc(NULL, initLength);
 
     if (trlbuf.buffer == NULL) {
-        asnmessage ("Out of memory: Couldn't resize buffer for trailer file.");
+        printfAndFlush ("Out of memory: Couldn't resize buffer for trailer file.");
         status = OUT_OF_MEMORY;
```
jamienoss: Also, this function is an example of very ugly error handling.

jamienoss:
@mdlpstsci I am unsure of your comment about ``ASNWARN``. This is ASN not ACS. I also didn't alter this func, I altered ``asnmessage()`` -> ``printAndFlush()``.

lib/trlbuf.c
jamienoss: Ok so this is why.
```c
/* This function will reset the preface to a null string, so nothing
        more gets written out to any other trailer files.  It also
        reallocates the space to preface so that it doesn't take any memory.
    */
```
COmment in ``void ResetTrlPreface (void)``

mdlpstsci:
It was just in a comment where ASNWARN was referenced, but it is not actually in this library.  It is in the CALACS library so it could be confusing to someone in the future.

Michele D. De La Pena
Space Telescope Science Institute
mdelapena@stsci.edu
667.218.6501

lib/trlbuf.c
jamienoss: Or at least a description of the intention. It still seems wrong to me.

jamienoss:
@mdlpstsci I haven't properly finished and therefore tested this yet but if you want to take a look, it might help. Thanks.

jamienoss:
>@mdlpstsci I haven't properly finished and therefore tested this yet but if you want to take a look, it might help. Thanks.

Tested and working OK. @mdlpstsci all yours, thanks.

PR #245, opened 2017-10-08
PtrRegister should not leak on assert()

### LINKED ISSUE #244: PtrRegister should not leak on assert()
These two sections of code leak:

[here](https://github.com/spacetelescope/hstcal/blob/master/hstio/hstio.c#L168-L169)
```c
...
assert(reg->ptrs = realloc(reg->ptrs, reg->length*sizeof(*reg->ptrs)));
assert(reg->freeFunctions = realloc(reg->freeFunctions, reg->length*sizeof(*reg->freeFunctions)));
...
```
and [here](https://github.com/spacetelescope/hstcal/blob/master/hstio/hstio.c#L190-L191)
```c
...
if (!found)
        assert(0); //internal error: the ptr trying to be freed was not added to the register
```

My original logic behind not caring about this was to simply let the OS tidy it up. Upon reflection this seems stupid, the entire point of this 'class' is to prevent leaks and therefore it should definitely not be leaking itself.

jamienoss:
Fixes #244

### DIFF
--- a/hstio/hstio.c
+++ b/hstio/hstio.c
@@ -165,8 +165,20 @@ void addPtr(PtrRegister * reg, void * ptr, void * freeFunc)
     if (++reg->cursor >= reg->length)
     {
         reg->length += PTR_REGISTER_LENGTH_INC;
-        assert(reg->ptrs = realloc(reg->ptrs, reg->length*sizeof(*reg->ptrs)));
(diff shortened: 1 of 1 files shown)


jamienoss:
@mdlpstsci rebased and now up to date.

PR #253, opened 2017-10-11
Resolve fclose() related bugs #226 & #252

### LINKED ISSUE #226: Calls to flcose() do not error check
It seems that no call to ``fclose(FILE *)`` checks for a returned error,  ``EOF``. This is really only a problem if the file is opened to write to and even then if all explicit writes aren't flushed. That being said, the last point, really just passes the buck to error checking on the flush instead, which probably is never checked either.

``int fclose(FILE *)`` returns 0​ on success, EOF otherwise. _c.f._ http://en.cppreference.com/w/c/io/fclose

stdio.h source comment:
```
Name: fclose
Prototype: int fclose (FILE *stream)
Description:
This function causes stream to be closed and the connection to the corresponding file to be broken. Any 
 buffered output is written and any buffered input is discarded. The fclose function returns a value of 0 if the 
 file was closed successfully, and EOF if an error was detected. 

It is important to check for errors when you call fclose to close an output stream, because real, everyday 
 errors can be detected at this time. For example, when fclose writes the remaining buffered output, it might 
 get an error because the disk is full. Even if you know the buffer is empty, errors can still occur when closing 
 a file if you are using NFS. 

The function fclose is declared in stdio.h. 
Header files:
stdio.h
```

### LINKED ISSUE #252: fclose(NULL) segfaults
From pandokia testing ([log-10-10-17](https://ssb.stsci.edu/pandokia/pandokia.cgi?query=treewalk&attn=%2A&compare=0&context=dev%3A3.5.4&host=nott&project=hstcal&status=F&test_name=acs%2Fcalacs_e%2F%2A&test_run=conda-conda_tests_py3_2017-10-10-10%3A30%3A02)) three tests failed. Testing [HSTCAL 748d4b00d5635dd53ca88398457134e19f10b146](https://github.com/spacetelescope/hstcal/tree/748d4b00d5635dd53ca88398457134e19f10b146). No stdout info was given, no trailer files either. Status returned !=0 aka None. Looking at the run log on the [10-11-17](https://ssb.stsci.edu/steuermann/steuermann.cgi?action=log&name=conda_tests_py3_2017-10-11-10:30:02/nott:conda_testing3/dev_tests) and searching for "hrc_single3_noniraf" (one of the failed tests) the following was discovered.
```
COMMAND : 'pdknose --pdk --with-doctest --doctest-tests hrc_single3_noniraf.py' (for file /srv/rt/hstcal/acs/calacs_e/hrc_single3_noniraf.py) 2017-10-11 10:41:15.655787
F
======================================================================
FAIL: hrc_single3_noniraf.test_hrc_single3_noniraf
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/srv/iraf/bldtmp/pvxk/p/envs/dev/lib/python3.5/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/srv/rt/hstcal/acs/calacs_e/hrc_single3_noniraf.py", line 78, in test_hrc_single3_noniraf
    assert status == 0 , 'run_process returned status %d' %status
AssertionError: run_process returned status -11
-------------------- >> begin captured stdout << ---------------------
/srv/iraf/bldtmp/pvxk/p/envs/dev/bin/calacs.e

--------------------- >> end captured stdout << ----------------------
```

This at least gave the signal ``-11``, a segfault on linux.

Debugging this found the cause to be a ``NULL`` ptr being passed to ``fclose()`` within [lib/trlbuf.c::CloseTrlBuf()](https://github.com/spacetelescope/hstcal/blob/748d4b00d5635dd53ca88398457134e19f10b146/lib/trlbuf.c#L630) from the final call to ``CloseTrlBuf()`` in [calacs/acsmain.c::main()](https://github.com/spacetelescope/hstcal/blob/748d4b00d5635dd53ca88398457134e19f10b146/pkg/acs/calacs/calacs/acsmain.c#L246).

Why this is ``NULL`` is under current investigation.

Poor concept of ownership. Internal functions closing, freeing, & nulling 'stuff' upon error when they probbaly shouldn't be.

jamienoss:
Set error status on failed fclose
    
Resolves #226
    
AND

Guard fclose(std::FILE* stream) against stream == NULL
    
Fixes #252
    
### DIFF
--- a/pkg/stis/calstis/cs1/doblev.c
+++ b/pkg/stis/calstis/cs1/doblev.c
@@ -183,8 +183,14 @@ int *driftcorr    o: true means correction for drift along lines was applied
 		fprintf (ofp, "%6d  %0.6g\n", j+1, biaslevel + averagedrift);
 	}
 
-	if (ofp != NULL)
-	    fclose (ofp);			/* done with output file */
+
+	if ((status = fcloseWithStatus(&ofp))) /* done with output file */
+	{
+	    freeSingleGroup (*x);
+	    if (*x)
+	        free (*x);
+	    return status;
+	}
 
 	/* This is the mean value of all the bias levels subtracted. */
 	*meanblev = sumbias / out->sci.data.ny;
@@ -223,7 +229,7 @@ static FILE *BlevOpen (char *fname, int extver, int *status) {
 
 	    } else {
 
-		fclose (fp);
+	        (void)fcloseWithStatus(&fp);
 		printf ("ERROR    File %s already exists.\n", fname);
 		*status = OPEN_FAILED;
 		fp = NULL;
--- a/pkg/stis/calstis/cs1/doblev.c
+++ b/pkg/stis/calstis/cs1/doblev.c
@@ -184,24 +184,19 @@ int *driftcorr    o: true means correction for drift along lines was applied
 	}
 
 
-	if ((status = fcloseWithStatus(&ofp))) /* done with output file */
-	{
-	    freeSingleGroup (*x);
-	    if (*x)
-	        free (*x);
-	    return status;
-	}
+	status = fcloseWithStatus(&ofp); // Don't return here and give caller opportunity to recover.
 
 	/* This is the mean value of all the bias levels subtracted. */
 	*meanblev = sumbias / out->sci.data.ny;
 
-	/* Free x, and copy out to x. */
+	// Free 'x', and move 'out' to 'x'.
 	freeSingleGroup (*x);
-	free (*x);
+	if (*x)
+	    free (*x);
 	*x = out;
 	*done = 1;			/* overscan was actually removed */
 
-	return (0);
+	return (status);
 }
 
 /* This routine opens an output text file if the name is non-null. */
@@ -232,7 +227,6 @@ static FILE *BlevOpen (char *fname, int extver, int *status) {
 	        (void)fcloseWithStatus(&fp);
 		printf ("ERROR    File %s already exists.\n", fname);
 		*status = OPEN_FAILED;
-		fp = NULL;
 	    }
 
 	} else {
--- a/hstio/hstio.c
+++ b/hstio/hstio.c
@@ -115,6 +115,26 @@
 # include "hstio.h"
 # include "hstcalerr.h"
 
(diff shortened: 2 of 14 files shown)


pkg/stis/calstis/cs1/doblev.c
```diff
@@ -223,7 +229,7 @@ static FILE *BlevOpen (char *fname, int extver, int *status) {
 
 	    } else {
 
-		fclose (fp);
+	        (void)fcloseWithStatus(&fp);
 		printf ("ERROR    File %s already exists.\n", fname);
 		*status = OPEN_FAILED;
 		fp = NULL;
```
mdlpstsci: fp has already been set to NULL in fcloseWithStatus

jamienoss: I did notice this, I forgot to delete it. Doesn't matter too much.

jamienoss:
@mdlpstsci Sorry. I took one further look at it before merging.

If it was to return where it did then really I needed to add make ``*x = NULL``.
Then I thought, actually, that's not be the funcs responsibility. The caller should be responsible for recovery. In which case I needed to make ``*x = out``. This also meant adding ``*meanblev = sumbias / out->sci.data.ny;`` to the returning ``IF``. At that stage, I decided it was simpler to just set ``status`` and carry on with final assignments and just return status.

What do you think?

eteq:
@jamienoss - workflow question: can you clarify what the "sign-off" means here?  In principle one should never "sign off" on their own code - instead a reviewer should "approve" it.  Or do you mean something else with the "sign-off" in the PR message?

jamienoss:
@eteq Hey Erik, sure thing. This is not related to review approvals.

It is nothing more than ``git commit -s``
https://git-scm.com/docs/git-commit#git-commit--s

As I have mentioned before in the following [comment](https://confluence.stsci.edu/pages/viewpage.action?pageId=94096360&focusedCommentId=101618671#comment-101618671) relating to our GitHub Policy.

eteq:
Oh, gotcha @jamienoss.  Didn't realize that was coming from the commit logs. Thanks for clearing that up.

PR #257, opened 2017-10-16
Remove extraneous Mac builds to lessen burden to Travis queue

pllim:
Mac builds on Travis CI are very susceptible to long queue compared to Linux builds. Given that HSTCAL does not depend on Python (outside of `waf`, which is a third-party package), there is really no need to test Mac builds across different Python versions. Here, I removed some duplicate builds to lessen the Travis queue time for testing this package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,18 +31,6 @@ matrix:
 
   - os: osx
     env:
-      - PYTHON_VERSION=2.7
-      - CFLAGS="-m64"
-      - LDFLAGS="-m64 -Wl,-rpath $EXE_PREFIX/envs/test/lib"
-
-  - os: osx
-    env:
-      - PYTHON_VERSION=3.5
-      - CFLAGS="-m64"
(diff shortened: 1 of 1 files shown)


jamienoss:
> Given that HSTCAL does not depend on Python (outside of waf, which is a third-party package), there is really no need to test Mac builds across different Python versions.

Whilst ``waf`` is a 3rd-party package, our ``wscripts`` are not. Since they constitute the entire package build mechanism, they __MUST__ be tested! That is the entire point.

pllim:
If `waf` can run in, say, Python 3.5 under Linux, there is no reason it won't run with same Python version on Mac. Anyway, this will be moot with Jenkins, I hope.

jamienoss:
> If waf can run in, say, Python 3.5 under Linux, there is no reason it won't run with same Python version on Mac.

I understand that but it just ins't _good_ testing practice. The rule of thumb for testing is to try and keep things as black-box as possible. As soon as you deviate from this you are reducing your testing integrity. You are also creating cross test dependencies which is not a good idea.

PR #258, opened 2017-10-17
ACS CTE: Use amp dependent RN from CCDTAB and not general RN from PCTETAB

### LINKED ISSUE #213: ACS CTE: Use amp dependent read-noise
My understanding of the acscte algorithm is that one of the first steps is to determine if a pixel has enough signal to noise to apply a decent correction. It does this by checking to see if the value of a pixel in the science array is above the readnoise of the detector which is a straight threshold. This is a good check, but it appears that acscte uses a readnoise value that is quoted in the header of the pctetab as the threshold value. However, @nmiles2718 discovered the value in the pctetab is wrong for ACS/WFC and the readnoise changes for each amplifier so a single value is not appropriate for determining this threshold. 

@nmiles2718 also discovered that this has a large effect in the number of corrected pixels when CTE correcting low background observations or dark images as there is very little signal. Also, because different amplifiers have different read noise properties it could correct amplifiers differently when it should not. While these pixels may not have much moved charge it does make a difference.

For a solution I think we could do one of three things:
- I *think* that at this point the blv_tmp file error arrays should have the readnoise and bias so we could use that as an indicator of initial signal to noise on a pixel by pixel basis. If the error array has the Poission noise from the science in there we may not want to do this, but we should consult Jay about that.
- Use the error array of the bias image appropriately scaled to determine signal to noise also on a pixel by pixel basis.
- Use an amplifier dependent readnoise by reading the values in from the ccdtab.

We will probably want to consult with Jay on the best solution.

jamienoss:
Resolves #213

### DIFF
--- a/ctegen2/ctehelpers.c
+++ b/ctegen2/ctehelpers.c
@@ -204,7 +204,7 @@ int loadPCTETAB (char *filename, CTEParamsFast *pars) {
        CTEDATE1 - reference date of CTE model pinning, in fractional years
 
        PCTETLEN - max length of CTE trail
-       PCTERNCL - readnoise amplitude and clipping level
+       PCTERNOI - readnoise amplitude and clipping level
        PCTESMIT - number of iterations used in CTE forward modeling
        PCTESHFT - number of iterations used in the parallel transfer
        PCTENSMD - readnoise mitigation algorithm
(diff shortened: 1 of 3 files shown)


jamienoss:
Moved to #271

PR #271, opened 2017-10-23
ACS CTE: Use amp dependent read-noise

### LINKED ISSUE #213: ACS CTE: Use amp dependent read-noise

### LINKED ISSUE #270: Update file header keys or add as history key if missing
The ``putKey<type>()`` functions in [hstio/keyword.c](https://github.com/spacetelescope/hstcal/blob/master/hstio/keyword.c#L645-L758) don't allow you to just update a keyword, i.e. if the key is mising do nothing. Instead they add the keyword if it is found missing.
```c
int putKeyF (Hdr *hdr, char *keyword, float value, char *comment) {

/* Arguments:
**      hdr     i: pointer to header to be updated
**      keyword i: name of keyword
**      value   i: value of keyword
**      comment i: comment to add with keyword if keyword doesn't exist
*/

        FitsKw kw;      /* keyword pointer */

        kw = findKw (hdr, keyword);
        if (kw == NotFound)
            addFloatKw (hdr, keyword, value, comment);
        else
            putFloatKw (kw, value);

        if (hstio_err())
            return (1);

        return (0);
}
```

These new functions are needed to facilitate with #213 so that the header keys can be updated if they exist and added as history keys if missing.

jamienoss:
ACS CTE: Use amp dependent RN from CCDTAB and not general RN from PCTETAB
    
Resolves #213

AND

Add updateKeyX() and updateKeyOrAddAsHistKeyX() functions
    
Resolves #270

### DIFF
--- a/pkg/acs/calacs/acscte/dopcte-gen2.c
+++ b/pkg/acs/calacs/acscte/dopcte-gen2.c
@@ -93,6 +93,10 @@ int doPCTEGen2 (ACSInfo *acs, CTEParamsFast * ctePars, SingleGroup * chipImage)
         amploc = strchr(AMPSORDER, ccdamp[nthAmp]);
         ampID = *amploc - AMPSORDER[0];
 
+        ctePars->rn_amp = acs->readnoise[ampID];
+        sprintf(MsgText, "(pctecorr) Read noise level from CCDTAB: %f.", ctePars->rn_amp);
+        trlmessage(MsgText);
+
         /* get amp array size */
         if (get_amp_array_size_acs_cte(acs, chipImage, ampID, amploc, ccdamp,
                                &amp_xsize, &amp_ysize, &amp_xbeg,
--- a/ctegen2/ctegen2.h
+++ b/ctegen2/ctegen2.h
@@ -101,7 +101,7 @@ void initCTEParamsFast(CTEParamsFast * pars, const unsigned _nTraps, const unsig
 int allocateCTEParamsFast(CTEParamsFast * pars);
 void freeCTEParamsFast(CTEParamsFast * pars);
 
-int populateHeaderWithCTEKeywordValues(SingleGroup *group, CTEParamsFast *pars);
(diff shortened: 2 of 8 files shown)


pkg/acs/calacs/acscte/dopcte-gen2.c
```diff
@@ -93,6 +93,10 @@ int doPCTEGen2 (ACSInfo *acs, CTEParamsFast * ctePars, SingleGroup * chipImage)
         amploc = strchr(AMPSORDER, ccdamp[nthAmp]);
         ampID = *amploc - AMPSORDER[0];
 
+        ctePars->rn_amp = acs->readnoise[ampID];
```
jamienoss: @mdlpstsci This single line is the actual fix. Everything else is to just reporting etc.

jamienoss:
@mdlpstsci just rebased so needs your approval again, cheers.

PR #275, opened 2017-10-31
Plug memory leaks in main.c code.

### LINKED ISSUE #239: Memory leak when exiting from cmd opt parsing
For command line options such as ``--gitinfo`` and ``--version``, the code is immediately exited from without freeing _some_ already allocated memory via ``FreeNames()`` or individually with ``free()``. The memory allocated is for the input and output filename lists.

``FreeNames()`` should check fro ``NULL`` before freeing.

This really shouldn't be explicit to ``FreeNames`` and should instead use ``PtrRegister`` srtuct to facilitate in this.

### Update:
Going to use ``PtrRegister`` instead of ``FreeNames()``. ``FreeNames()`` requires ``NULL`` checking and I can't be bothered to change all of these static funcs. Becuase of this, this issue is dependent on #236 

### Update
This is dependent on:
- [x] #236 Move PtrRegister from hstio/ to lib/ 
- [x] #249 CloseTrlBuf(void) -> CloseTrlBuf(struct TrlBuf * ptr)
- [x] #250 Move PtrRegister to lib/hstcal_memory.c and include/hstcal_memory.h
- [ ] ~~#246 Add PtrRegister::updatePtr(PtrRegister * reg, void * oldPtr, void * newPtr)~~

### LINKED ISSUE #274: Add --help to stis/cs7
This was missed in #238

jamienoss:
Resolves #239
Fixes #274 (smuggled)

### DIFF
--- a/pkg/stis/calstis/cs0/cs0.c
+++ b/pkg/stis/calstis/cs0/cs0.c
@@ -4,6 +4,7 @@
 # include <stdlib.h>		/* calloc */
 # include <string.h>
 
+#include "hstcal_memory.h"
 # include "c_iraf.h"		/* for c_irafinit */
 # include "ximio.h"
 
@@ -12,7 +13,6 @@
 # include "hstcalerr.h"
 # include "hstcalversion.h"
 
-static void FreeNames (char *, char *, char *, char *, char *, char *);
 static void printSyntax(void)
 {
     printf("syntax:  cs0.e [--help] [-t] [-s] [-v] [--version] [--gitinfo] input [outroot] [-w wavecal]\n");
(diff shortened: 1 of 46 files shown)


jamienoss:
@mdlpstsci The three commits can be reviewed independently, however, they will be squashed before being merged.

pkg/stis/calstis/cs0/cs0.c
```diff
@@ -130,41 +144,46 @@ int main (int argc, char **argv) {
...
 		strcpy (rawlist, argv[i]);
 	    } else if (outlist[0] == '\0') {
-		free (outlist);
+		freePtr(&ptrReg, outlist);
 		if ((outlist = calloc (strlen(argv[i])+1, sizeof(char)))
 			== NULL) {
 		    printf ("ERROR:  Out of memory.\n");
 		    exit (ERROR_RETURN);
 		}
+		addPtr(&ptrReg, rawlist, &free);
```
mdlpstsci: Did you mean this to be outlist instead of rawlist?

pkg/stis/calstis/cs2/wscript
```diff
@@ -5,7 +5,7 @@ def build(bld):
         name = 'cs2.e',
         source = 'cs2.c',
         target = 'cs2.e',
-        use = ['calstis', 'imphttab'] + bld.env.LOCAL_LIBS,
+        use = ['hstcallib', 'calstis', 'imphttab'] + bld.env.LOCAL_LIBS,
         lib = bld.env.EXTERNAL_LIBS,
```
mdlpstsci: Is hstcallib necessary here?

pkg/stis/calstis/cs0/cs0.c
jamienoss: Yes, good catch, thanks.

pkg/stis/calstis/cs2/wscript
jamienoss: No, because cs2.c doesn't use the PtrRegister. I probably added this when adding the rest as a blind list through the main files. I'll remove,  thanks.

jamienoss:
@mdlpstsci changes made and I squashed the 3 commits + rebased onto upstream/master.

To check:
```
git checkout jamienoss/issue239-mem-leak-from-main
git fetch jamienoss
git diff  jamienoss/issue239-mem-leak-from-main
```
From this, you should see that the only changes are those requested.

PR #283, opened 2017-11-27
Fix unguarded IFs introduced in #275

### LINKED ISSUE #280: Fix regression from #275
Woops, looks like I broke STIS.

The following test segfaulted both on `nott` (linux) and `pembry` (OSX):
On `nott` [stis/wavecal/wavecal_lev2](https://ssb.stsci.edu/pandokia/pandokia.cgi?query=detail&qid=7398)
On `pembry` [stis/wavecal/wavecal_lev2](https://ssb.stsci.edu/pandokia/pandokia.cgi?query=detail&qid=7399)

I ran this through the regression suite but with my own code which I've just noticed had all the STIS tests turned off, [here](https://github.com/jamienoss/regress/blob/3137606551d0e87aecc1ee232086d7d68302809d/regress.py#L453-L457).
```python
...
 # Start with ACS (have to start somewhere)
        exe = os.path.join(execPath, 'calacs.e')
        checkExeExists(exe)
        for test in acsInput:
            testItem = TestQueueItem(test, exe, outPath)
            testQueue.put(testItem)

        # Then STIS
        # cmd = os.path.join(execPath, 'calstis.e')
        # for test in stisInput:
        #    testItem = TestQueueItem(test, cmd, outPath)
        #    testQueue.put(testItem)

        # Then WFC3
        exe = os.path.join(execPath, 'calwf3.e')
        checkExeExists(exe)
        for test in wf3Input:
            testItem = TestQueueItem(test, exe, outPath)
            testQueue.put(testItem)
...
```

jamienoss:
Fixes #280

This also addresses the "-Wmisleading-indentation" compiler warnings generated
from the following files, also from #275

pkg/stis/calstis/lib/reffiles.c
pkg/stis/calstis/cs12/cs12.c
pkg/acs/calacs/lib/reffiles.c
pkg/wfc3/calwf3/lib/reffiles.c

### DIFF
--- a/pkg/acs/calacs/lib/reffiles.c
+++ b/pkg/acs/calacs/lib/reffiles.c
@@ -128,7 +128,7 @@ RefFileInfo *ref   io: the list of reference info
     if (!ref)
         return;
 
-	RefFileInfo *current, *next;
+    RefFileInfo *current, *next;
 
 	current = ref->next;		/* don't free the first record */
 	while (current) {
--- a/pkg/stis/calstis/cs12/cs12.c
+++ b/pkg/stis/calstis/cs12/cs12.c
@@ -147,9 +147,11 @@ int main (int argc, char **argv) {
 	    exit (ERROR_RETURN);
 	}
 	if (which_wavecal[0] != '\0') {
-	    if ((status = WavOption (which_wavecal, &w_option)))
-	        freeOnExit(&ptrReg);
-		exit (status);
+        if ((status = WavOption (which_wavecal, &w_option)))
+        {
+            freeOnExit(&ptrReg);
+            exit (status);
+        }
 	} else {
 	    w_option = STIS_LINEAR;
 	}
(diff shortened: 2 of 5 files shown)


pkg/acs/calacs/lib/reffiles.c
```diff
@@ -128,7 +128,7 @@ RefFileInfo *ref   io: the list of reference info
     if (!ref)
         return;
 
-	RefFileInfo *current, *next;
+    RefFileInfo *current, *next;
```
mdlpstsci: I agree to moving the declaration to the same indented position as the "if", but I am not sure why the code following this line was not also adjusted.  My statement is true for all of the reffiles.c programs.

jamienoss: They are both indented, the `if` with 4 ws and this dec with a tab. GitHub renders these differently for observation clarity. I could have changed the following lines of code for consistency but this would then link these lines to this PR via git blame. This is something I wanted to avoid. Changing the dec line was that minimally required to remove the compiler warning.

mdlpstsci: OK - understood.

jamienoss: I could have used a tab to indent the intended changed lines which would have also solved the issue, but then ws is better than a tab.

Issue #199, opened 2017-08-24
Call to CALACS somehow invokes ancient ACSDTH step


pllim:
**UPDATE**

Bugs to be fixed:
- [ ] Outdated message about PyDrizzle. This message should be updated.
- [ ] Why make me delete SPT files and then complains about the files not there?

**Earlier log**

@dborncamp is trying to process `jcqtb4020_asn.fits` but ran into an ancient and unexpected error message after ACS2D is completed:
```
<snipped>
*** ACS2D complete ***


CALACSBEG*** ACSDTH -- Version 9.2.0 (01-Jun-2017) ***
Begin    24-Aug-2017 14:22:24 EDT
The task PyDrizzle needs to be run in order to generate
a geometrically corrected, drizzle-combined product.
PyDrizzle requires PyRAF. See pyraf.stsci.edu for more details.
Warning    Output file `jcqtb4020_spt.fits' already exists.
ERROR:    CALACS processing NOT completed for jcqtb4020_asn.fits
ERROR:    CALACS processing NOT completed for jcqtb4020_asn.fits
ERROR:             status = 1021
```
ACSDTH is an ancient CALACS step that is no longer listed in ACS Data Handbook despite the existence of `pkg/acs/calacs/calacs/acsdth.c`.

To access the data:
```
ls /grp/hst/acs7/dborncamp/darktesting/ff/monthlong/jcqtb4*
```

To run:
```
calacs.e -v jcqtb4020_asn.fits
```

Content of ASN table:
```
<Table length=5>
   MEMNAME        MEMTYPE     MEMPRSNT
    str14          str14        bool  
-------------- -------------- --------
JCQTB4A5Q      EXP-DTH            True
JCQTB4BNQ      EXP-DTH            True
JCQTB4BWQ      EXP-DTH            True
JCQTB4DEQ      EXP-DTH            True
JCQTB4020      PROD-DTH           True
```

Trailer file with the failure is attached (because you need to delete the outputs before re-running, so this might get lost if not attached now):

[jcqtb4020.tra.txt](https://github.com/spacetelescope/hstcal/files/1249990/jcqtb4020.tra.txt)

pllim:
So... The reason for the error message actually had nothing to do with PyDrizzle but because SPT files already exists! When I delete the SPT files, it complains about SPT files not found (???) but completes successfully:
```
*** ACS2D complete ***
Finished CCD processing...
CALACS: Building DTH products


[...]

PR #285, opened 2017-12-11
Issue199 acsdth errors

mdlpstsci:
This request addresses Issue #199 (Call to CALACS somehow invokes ancient ACSDTH step).  The updates involve clarifying both Error and Warning messages for the ACSDTH step.  In addition, an error in producing the output association SPT file has been addressed.


mdlpstsci:
Moved to #286

PR #286, opened 2017-12-11
CALACS: Clarifies error/warning messages for SPT files.

### LINKED ISSUE #199: Call to CALACS somehow invokes ancient ACSDTH step

mdlpstsci:
Improves error message for existing output association SPT file and warning messages
for individual input SPT files.  Also, now writes all existing input SPT arrays into
output association SPT file.

This fix potentially changes the output association SPT file. CALACS now writes all
existing input SPT arrays (extensions) into the output association SPT file.  Formerly,
if the first input SPT file named in the association table did not exist, no output
association SPT file would be written at all.  Also, if an input SPT file contained
multiple SPT arrays, only one of the arrays would be written to the output association
SPT file.

Resolves #199
Signed-off-by: Michele De La Pena <mdelapena@stsci.edu>

### DIFF
--- a/pkg/acs/calacs/calacs/acsdth.c
+++ b/pkg/acs/calacs/calacs/acsdth.c
@@ -35,11 +35,11 @@ static int putEmptyDQ(char *, int, DQHdrData *, int);
  Warren J. Hack, 1998 Oct 14:
  Initial version.
  Warren J. Hack, 1999 Nov 11:
- Adds creation of SPT file for output product.
+ Adds creation of _spt.fits file for output product.
  Warren J. Hack, 2000 Jun 23:
  Modified trailer file creation to include ALL input files.
- Also modified SPT file creation to include all extensions
- from input SPT files.
+ Also modified _spt.fits file creation to include all extensions
+ from input _spt.fits files.
  Warren J. Hack, 2002 Feb 6:
  Moved trailer file initialization outside this so that a
  trailer file can be created/appended for all ASN tables regardless
@@ -47,6 +47,9 @@ static int putEmptyDQ(char *, int, DQHdrData *, int);
  This initialization is now handled by CalAcsRun in 'calacs.c'.
  Warren J. Hack, 2002 April 18:
  Eliminated creation of dummy '_dth.fits' product.
+ Michele D. De La Pena, 2017 September 26:
+ Updated the message regarding the creation of a drizzle-combined data product
+ and some cleanup.
 
  int AcsDth (char *input, char *output, int dthcorr, int printtime, int verbose) {
  */
@@ -80,12 +83,6 @@ int AcsDth (char *in_list, char *output, int dthcorr, int printtime, int verbose
 
   /* ----------------------- Start Code --------------------------------*/
 
-	/* Determine the names of the trailer files based on the input
-   and output file names, then initialize the trailer file buffer
-   with those names.
-   */
-
-  /*	InitDthTrl (in_list, output); */
 	root[0] = '\0';
 	sprintf(mtype,"PROD-DTH");
 
@@ -100,23 +97,18 @@ int AcsDth (char *in_list, char *output, int dthcorr, int printtime, int verbose
 	if (printtime)
     TimeStamp ("ACSDTH started", "");
 
-  sprintf(MsgText,"The task PyDrizzle needs to be run in order to generate");
+  sprintf(MsgText,"The DrizzlePac software package should be used to generate a drizzle-combined");
   trlmessage(MsgText);
-  sprintf(MsgText,"a geometrically corrected, drizzle-combined product.");
-  trlmessage(MsgText);
-  sprintf(MsgText,"PyDrizzle requires PyRAF. See pyraf.stsci.edu for more details.");
+  sprintf(MsgText,"data product.  Please see drizzlepac.stsci.edu for details.\n");
   trlmessage(MsgText);
 
-  /* create new SPT file for output product */
+  /* create new _spt.fits file for output product */
   if (mkNewSpt (in_list, mtype, output)) {
     return(status);
   }
 
   c_imtclose(tpin);
 	PrEnd ("ACSDTH");
-
-	/* Write out temp trailer file to final file */
-	WriteTrlFile ();
 
 	return (status);
 }
@@ -218,7 +210,7 @@ void InitDthTrl (char *inlist, char *output) {
 			trlmessage (MsgText);
 		}
 
-		/* Now, convert trailer filename extensions from '.fits' to '.trl' */
+		/* Now, convert trailer filename extensions from '.fits' to '.tra' */
 		if (MkNewExtn (out_name, TRL_EXTN) ) {
 			sprintf(MsgText, "Error creating input trailer filename %s", out_name);
 			trlerror (MsgText);
--- a/pkg/acs/calacs/calacs/calacs.c
+++ b/pkg/acs/calacs/calacs/calacs.c
@@ -53,6 +53,9 @@
  Moved FLSHCORR from ACSCCD to ACS2D.
  PLL, 2013 Aug 9:
  Separated PCTECORR from ACSCCD. Fixed indentation nightmare.
+ MDD, 2017 Sept 05:
+ Clarify WARNING and ERROR messages associated with _spt files in
+ the ACSDTH step.
 */
 
 void updateAsnTable (AsnInfo *, int, int);
@@ -203,55 +206,57 @@ int CalAcsRun (char *input, int printtime, int save_tmp, int verbose, int debug,
         trlmessage("Finished MAMA processing...");
     }
 
-
-    /* Add DTH processing here... */
-    /* For each DTH product... */
+    /* CALACS no longer generates a drizzled product. Drizzle-combined data is generated   
+       by post-CALACS processing. Instead, the "ACSDTH" pipeline step can create both combined 
+       _spt.fits and .tra files. In order to have a combined output _spt.fits file generated, some 
+       _spt.fits files must exist for the filenames listed in the association table.
+    */
+ 
+    /* Loop over each entry in the association table to create a combined _spt.fits and a 
+       combined .tra file.  The terminology of "dth" is leftover from the original CALACS 
+       implementation.
+    */
     if (asn.process == FULL){
-        if (asn.verbose) {
-            trlmessage ("CALACS: Building DTH products");
-        }
         acsdth_input = NULL;
         for (prod = 0; prod < asn.numprod; prod++) {
-            /* Create empty DTH product, header only */
-            /* This uses only one sub-product for the header template,
-               but later versions should use a function similar to
-               BuildSumInput to create list of subproducts as inputs...
-            */
+
+            /* Create the list of input files */
             acsdth_input = BuildDthInput (&asn, prod);
 
             /* We always want to create a final concatenated trailer file for
-               the entire association whether there is a product or not. So, we
-               set up the trailer file based on the association file name itself.
+               the entire association.  Set up the trailer file based on the 
+               association file name itself.
             */
 
-            /* If desired, we could optionally use the full _drz.tra filename
-               as the trailer filename, based on the output dither product name.
-            if (strcmp(asn.product[prod].prodname,"") != 0) {
-                InitDthTrl (acsdth_input, asn.product[prod].prodname);
-            } else { */
-
+            trlmessage ("CALACS: Building combined .tra product");
             InitDthTrl(acsdth_input, asn.rootname);
 
-            /* End brace for optional dither product name assignment...
-            } */
+            /* Write out temp trailer file to final file */
+            WriteTrlFile ();
 
-            /* Check if we have a PROD-DTH specified...*/
+            /* Check if we have a PROD-DTH specified in the association table.  Even
+               though CALACS will not produce the drizzled product, the MEMTYPE of the 
+               rootname of the association dataset can be PROD-DTH.
+            */
             if (strcmp(asn.product[prod].prodname, "") != 0) {
 
                 if ((asn.dthcorr == PERFORM || asn.dthcorr == DUMMY)) {
+                    trlmessage ("CALACS: Building combined _spt.fits product");
                     if (AcsDth (acsdth_input, asn.product[prod].prodname, asn.dthcorr, printtime, asn.verbose) )
                         return (status);
 
-                    /* Pass posid of 0 to indicate a PRODUCT is to be updated */
                     updateAsnTable(&asn, prod, NOPOSID);
                 }
             } else {
-                trlwarn ("No DTH product name specified. No product created.");
-                /*status = ACS_OK; */
+                trlwarn ("No combined product specified in the association table - no _spt.fits file to be generated.");
             }
         }
-        free (acsdth_input);
+
+        if (acsdth_input != NULL) {
+            free (acsdth_input);
+        }
     }
+
     if (asn.verbose) {
         trlmessage ("CALACS: Finished processing product ");
     }
--- a/pkg/acs/calacs/lib/mkspt.c
+++ b/pkg/acs/calacs/lib/mkspt.c
@@ -1,6 +1,6 @@
-# include <ctype.h>
 # include <stdio.h>
 # include <string.h>
+# include <ctype.h>
 #include "hstcal.h"
 # include "xtables.h"
 
@@ -9,18 +9,19 @@
 # include "acs.h"
 # include "hstcalerr.h"
 
-/* mkNewSpt -- Create a new SPT file for the output file.
+/* mkNewSpt -- Create a new _spt.fits file for the output file.
 
   Description:
   ------------
-    New file will be based on primary header of first input file
+    New file will be based on primary header of first found input file
     given and will not have any extensions.
 
   Date          Author      Description
   ----          ------      -----------
   11-10-1999    W.J. Hack   Initial version
                             based on InitRejTrl and n_mkSPT from CALNICB
-
+  11-18-2017    M.D. De La Pena  Clarified message when input *_spt.fits files not found.  
+                            Generalized creation of output association _spt.fits file.
 */
 
 int mkNewSpt (char *in_list, char *mtype, char *output) {
@@ -29,25 +30,26 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
     arguments:
     char    *in_list            i: input filename/list to copy SPT data
     char    *mtype              i: type of exposure in association
-    char    *output             o: rootname of output SPT file
+    char    *output             o: rootname of output _spt.fits file
 */
 
     extern int  status;
-    IRAFPointer tpin;
+    IRAFPointer tpin = NULL;
     int         n;
-	Hdr         header;		                /* SPT header */
-	FILE        *fp;		                /* file pointer */
-	IODescPtr   im;		           /* descriptor for input image */
+	Hdr         header;          /* _spt.fits header */
+	FILE        *fp = NULL;      /* file pointer */
+	IODescPtr   im = NULL;       /* descriptor for output image */
+	IODescPtr   imUpdate = NULL; /* new descriptor if output image has to be updated */
 
     char        in_name[CHAR_FNAME_LENGTH+1];       /* filename of input data */
-    char        in_spt[CHAR_FNAME_LENGTH+1];        /* filename of SPT source data */
-    char        out_spt[CHAR_FNAME_LENGTH+1];       /* output SPT filename */
-    char        rootname[CHAR_FNAME_LENGTH+1];       /* output SPT rootname */
-    char        obsnum[4];                  /* Observation number string */
+    char        in_spt[CHAR_FNAME_LENGTH+1];        /* filename of _spt.fits source data */
+    char        out_spt[CHAR_FNAME_LENGTH+1];       /* output _spt.fits filename */
+    char        rootname[CHAR_FNAME_LENGTH+1];      /* output _spt.fits rootname */
+    char        obsnum[4];                          /* Observation number string */
     ShortHdrData  stmp;
     int         nimgs, i;
     int         nx;
-	int 		extnum,nextn;
+	int 		extnum, nextn;
 
 
     char        *isuffix[] = {"_blv_tmp","_crj_tmp","_flt","_crj","_dth","_sfl"};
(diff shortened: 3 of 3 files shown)


pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,6 +214,27 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+    /* If necessary, re-open the primary header to update the NEXTEND keyword */
+    if ((extnum != numSPT) || (numSPT != nimgs)) {
+
+        im = openUpdateImage (out_spt, "", 0, &header);
+        if (hstio_err()) {
+            trlopenerr (out_spt);
+            return (status = OPEN_FAILED);
+        }
+
+        if (PutKeyInt (&header, "NEXTEND", extnum, ""))
```
jamienoss: I'm unsure of this, so am just going to throw it out there - is this correct? Should `NEXTEND` be updated with `extnum`?

pkg/acs/calacs/calacs/calacs.c
```diff
@@ -203,51 +206,49 @@ int CalAcsRun (char *input, int printtime, int save_tmp, int verbose, int debug,
...
+       _spt.fits files must exist for the filenames listed in the association table.
+    */
+ 
+    /* Loop over each entry in the association table to create a combined _spt.fits and a 
+       combined .tra file.  The terminology of "dth" is leftover from the original CALACS 
+       implementation.
+    */
     if (asn.process == FULL){
-        if (asn.verbose) {
-            trlmessage ("CALACS: Building DTH products");
```
jamienoss: Should this really be removed? What if the output is parsed for this marker?

pkg/acs/calacs/calacs/calacs.c
```diff
@@ -203,51 +206,49 @@ int CalAcsRun (char *input, int printtime, int save_tmp, int verbose, int debug,
...
+               association file name itself.
             */
 
-            /* If desired, we could optionally use the full _drz.tra filename
-               as the trailer filename, based on the output dither product name.
-            if (strcmp(asn.product[prod].prodname,"") != 0) {
-                InitDthTrl (acsdth_input, asn.product[prod].prodname);
-            } else { */
-
+            trlmessage ("CALACS: Building combined TRA product");
```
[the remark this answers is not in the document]
jamienoss: Is this the replacement/update for line 211, ``trlmessage ("CALACS: Building DTH products");``, if so shouldn't it be conditional on `asn.verbose`?

pkg/acs/calacs/calacs/calacs.c
```diff
@@ -203,51 +206,49 @@ int CalAcsRun (char *input, int printtime, int save_tmp, int verbose, int debug,
...
 
-            /* Check if we have a PROD-DTH specified...*/
+            /* Check if we have a PROD-DTH specified in the association table.  Even
+               though CALACS will not produce the drizzled product, the MEMTYPE of the 
+               rootname of the association dataset can be PROD-DTH.
+            */
             if (strcmp(asn.product[prod].prodname, "") != 0) {
 
                 if ((asn.dthcorr == PERFORM || asn.dthcorr == DUMMY)) {
+                    trlmessage ("CALACS: Building combined SPT product");
```
jamienoss: Similarly on the verbosity conditional?

pkg/acs/calacs/calacs/calacs.c
```diff
@@ -203,51 +206,49 @@ int CalAcsRun (char *input, int printtime, int save_tmp, int verbose, int debug,
...
                     if (AcsDth (acsdth_input, asn.product[prod].prodname, asn.dthcorr, printtime, asn.verbose) )
                         return (status);
 
-                    /* Pass posid of 0 to indicate a PRODUCT is to be updated */
                     updateAsnTable(&asn, prod, NOPOSID);
                 }
             } else {
-                trlwarn ("No DTH product name specified. No product created.");
-                /*status = ACS_OK; */
+                trlwarn ("No indicator in the association table for combined SPT product to be generated.");
```
jamienoss: Does "indicator" explicitly mean something or should this be less ambiguous?

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -47,7 +48,7 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
     ShortHdrData  stmp;
     int         nimgs, i;
     int         nx;
-	int 		extnum,nextn;
+	int 		extnum, nextn;
```
jamienoss: You touch it you own it ;) Could these be initialized here?

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -77,8 +82,11 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
         return (status);
     }
 
-    /* See if an output SPT file already exists */
+    // FUTURE WORK: What if there are no input files so CALACS would not make an output file anyway?
+    /* In this case it is not necessary to See if an output SPT file already exists */
     if (FileExists (out_spt)) {
+        sprintf (MsgText, "Output SPT filename for %s exists - cannot overwrite file.", out_spt);
+        trlerror (MsgText);
         return (status);
```
jamienoss: I know this was already here, but does this leak?

pkg/acs/calacs/calacs/acsdth.c
```diff
@@ -115,9 +110,6 @@ int AcsDth (char *in_list, char *output, int dthcorr, int printtime, int verbose
   c_imtclose(tpin);
 	PrEnd ("ACSDTH");
 
-	/* Write out temp trailer file to final file */
-	WriteTrlFile ();
```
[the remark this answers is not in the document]
mdlpstsci: The call to create the combined trailer file is done in calacs.c (though the actual routine resides in acsdth.c).  I moved the WriteTrlFile() to after the "creation" invocation in calacs.c so it is more obvious where the combined file is opened and closed (though additional messages are still appended).

pkg/acs/calacs/calacs/calacs.c
mdlpstsci: Good point although CALACS has not actually built any true DTH products since 2002.  Even at that time the original _dth.fits products were actually dummy or empty files.  The changes made thus far are on the path to removing *any* indication of an ACSDTH "step" in CALACS after some discussion with the ACS team.   Any mention of an ACSDTH step has already been removed from the Data Handbook.

pkg/acs/calacs/calacs/calacs.c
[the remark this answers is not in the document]
mdlpstsci: As to .tra - changed as it is clearer, particularly as the trailer files seem to have been .trl at some point in the past.

I could not determine a specific and consistent protocol as to when an informational (versus warning or error) message should be qualified by a conditional.  I applied my own judgement.

pkg/acs/calacs/calacs/calacs.c
mdlpstsci: Again, this is a judgement call on my part as to whether this should be a conditional output statement or not.

pkg/acs/calacs/calacs/calacs.c
mdlpstsci: You got me as the term "indicator" is vague.  I was trying to avoid the use of "DTH" terminology which was one of several triggers for invoking AcsDth().   I could update the message to be "No combined product specified in the association table so no _spt.fits file to be generated.'

pkg/acs/calacs/lib/mkspt.c
mdlpstsci: Both of these variables are initialized to zero in the source before they are used.  In particular, nextn is the count of the number of extensions in any single _spt.fits file so it is re-initialized for each new input file.  I slightly prefer to have my variables initialized in the place where I am about to use them, but if you think most folks like to see this done when they are declared, I can change this.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -77,8 +82,11 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
     }
 
-    /* See if an output SPT file already exists */
+    // FUTURE WORK: What if there are no input files so CALACS would not make an output file anyway?
+    /* In this case it is not necessary to See if an output SPT file already exists */
```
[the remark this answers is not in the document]
mdlpstsci: Is levity allowed? Come on and see see see Tennessee Tuxedo.
Seriousness: See -> see.

SPT -> _spt.fits in files I have had to modify for this update.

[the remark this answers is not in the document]
pllim: When ACS team discusses file formats, "RAW" is used in place of ".raw". No one liked the extra typing anyway, so leaving as "SPT" will not confuse anyone.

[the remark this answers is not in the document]
jamienoss: I'm cool with either as long as there's logical consistency. 
As a side note, RAW is not an acronym, I presume SPT is though? In which case maybe SPT is better. What made me ask this is the use of TRA which definitely isn't an acronym but an abbreviation for "trailer" -> .tra. I really don't think we should be using TRA, this being an example of the logical consistency I mention.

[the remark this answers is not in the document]
jamienoss: Also, I wouldn't consider this type of revision/review-comment a show stopper, BTW, so I ultimately wouldn't care either way. However, code reviews, in my professional opinion, are one of the most efficient forms of peer to peer learning and discussion of techniques. Also, there is always just the foundation as a review, to state for the record that "I've noticed this".

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,6 +214,27 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
         freeHdr (&header);
     }
 
+    /* If necessary, re-open the primary header to update the NEXTEND keyword */
+    if ((extnum != numSPT) || (numSPT != nimgs)) {
+
+        im = openUpdateImage (out_spt, "", 0, &header);
```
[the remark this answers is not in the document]
mdlpstsci: The first use of "im" is closed via closeImage(im) before the call to openUpdateImage().

pkg/acs/calacs/lib/mkspt.c
mdlpstsci: Yes.  The variable extnum represents the total number of SPT arrays in all the *existing* input _spt.fits files, accounting for the fact that an _spt.fits file could have multiple arrays (extensions).

pkg/acs/calacs/lib/mkspt.c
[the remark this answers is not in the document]
jamienoss: Sure, but you had to check if that was really so.

pkg/acs/calacs/lib/mkspt.c
jamienoss: K

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,6 +214,27 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+        if (hstio_err()) {
+            trlopenerr (out_spt);
+            return (status = OPEN_FAILED);
+        }
+
+        if (PutKeyInt (&header, "NEXTEND", extnum, ""))
+            return (status = 1);
```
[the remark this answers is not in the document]
mdlpstsci: I agree an enumeration is much better than an anonymous "1"!  Unfortunately, the existing enumerations do not have an entry for the issue at hand.  However, I upgraded all the "put keyword" calls in this routine to report the HSTIO message and error return code to exploit existing functionality.  I also free the header or image descriptor when appropriate.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,6 +214,27 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+
+        /* Write the updated header and clean up */
+        putHeader(im);
```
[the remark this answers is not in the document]
mdlpstsci: I upgraded this call to report the HSTIO message and error return code. I also free the header or image descriptor.

pkg/acs/calacs/lib/mkspt.c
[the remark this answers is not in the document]
mdlpstsci: Yes, I had to check!  Several options, but in the end I chose to create a new IODescPtr which I deemed to be the most recognizable solution for most folks.  Having done this (and after our conversation), scope power rocks.

pkg/acs/calacs/lib/mkspt.c
mdlpstsci: This seems to be OK.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,6 +214,27 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
         freeHdr (&header);
     }
 
+    /* If necessary, re-open the primary header to update the NEXTEND keyword */
+    if ((extnum != numSPT) || (numSPT != nimgs)) {
+
+        im = openUpdateImage (out_spt, "", 0, &header);
+        if (hstio_err()) {
+            trlopenerr (out_spt);
+            return (status = OPEN_FAILED);
```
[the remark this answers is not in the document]
mdlpstsci: Done -> freeHdr (&header).

The openUpdateImage (and similar functionality in the HSTIO source) should do a more thorough job of cleaning up after errors.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,6 +214,27 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+        }
+
+        if (PutKeyInt (&header, "NEXTEND", extnum, ""))
+            return (status = 1);
+
+        /* Write the updated header and clean up */
+        putHeader(im);
+        closeImage (im);
```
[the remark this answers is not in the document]
mdlpstsci: There is no clean up done in closeImage() after putHeader() or fits_close_file() failures.  The closeImage() is part of the HSTIO source which needs to be made more robust with regard to error handling.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,7 +227,50 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+    }
+
     c_imtclose (tpin);
 	/* Successful return */
 	return (status);
 }
+
+// Simple helper routine to report the HSTIO error to the .tra file and free 
+// the corresponding header 
+void
```
jamienoss: Would this better as ``static``?

pkg/acs/calacs/lib/mkspt.c
jamienoss: Cool

pkg/acs/calacs/lib/mkspt.c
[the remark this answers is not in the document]
jamienoss: This is cool as is but it's still better to get into the habit of scoping. Appart from controlling memory lifetimes, it conveys a more concrete message of intent to future developers - the message being that the scoped variable has no need to persist beyond its scoped block. This can also help constrain namespace.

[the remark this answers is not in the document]
jamienoss: ...in addition, it also prevents the next developer from stumbling at the same point & same manner by trying to reuse it. They just won't be able to and therefore won't try doing so, the scope guides them.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,7 +229,50 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
 	/* Successful return */
 	return (status);
 }
+
+// Simple helper routine to report the HSTIO error to the .tra file and free 
+// the corresponding header 
+static void 
+reportErrorAndFreeHeader (Hdr *hdr)
+{
+	sprintf (MsgText, hstio_errmsg());
```
jamienoss: Is this even correct? Using `hstio_errmsg()` instead of the `int status` returned from the offending `PutXXX()` calls? I think (haven't properly checked) you should be using `WhichError(int status)`.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,7 +229,50 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+        if (hstio_err()) {
+            trlopenerr (out_spt);
+            freeHdr (&header);
+            return (status = OPEN_FAILED);
+        }
+
+        if (PutKeyInt (&header, "NEXTEND", extnum, "")) {
+            reportErrorAndFreeHeader (&header);
+            closeImage (imUpdate);
+            return (status = hstio_err());
```
jamienoss: Why return `hstio_err()`? This may not be correct, e.g. are you sure that the caller is expecting an enumeration in this range, of this type ([HSTIOERROR_](https://github.com/spacetelescope/hstcal/blob/master/include/hstio.h#L387-L393)) rather than those from [hstcalerr.h](https://github.com/spacetelescope/hstcal/blob/master/include/hstcalerr.h)? Furthermore, I checked a couple of the calls and they just pass `status` back up, so it's not just the immediate callers that this affects.
 
Furthermore, this shouldn't be bypassing the error handling by the called func.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,7 +229,50 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+        /* Write the updated header and clean up */
+        /* Ignore the returned value from putHeader as an hstio error is set at the same time */
+        if (putHeader(imUpdate)) {
+            reportErrorAndFreeHeader (&header);
+            closeImage (imUpdate);
+            return (status = hstio_err());
+        }
+
+        closeImage (imUpdate);
+        freeHdr(&header);
```
jamienoss: Since this doesn't immediately return, I would nullify `header`. Similarly for `imUpdate`.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,7 +229,50 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
     }
 
+    /* If necessary, re-open the primary header to update the NEXTEND keyword */
+    if ((extnum != numSPT) || (numSPT != nimgs)) {
+
+        imUpdate = openUpdateImage (out_spt, "", 0, &header);
+        if (hstio_err()) {
+            trlopenerr (out_spt);
+            freeHdr (&header);
+            return (status = OPEN_FAILED);
```
jamienoss: _c.f._ previous review [comment](https://github.com/spacetelescope/hstcal/pull/286#discussion_r156191226):

>Also, openUpdateImage can actually set an error and still have opened the file, e.g. for bad dims. In this case, the file needs to be closed before the above return. This being said, responsibility & ownership needs to be considered, i.e. should openUpdateImage close before setting the error and returning?

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,7 +229,50 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+            closeImage (imUpdate);
+            return (status = hstio_err());
+        }
+
+        /* Write the updated header and clean up */
+        /* Ignore the returned value from putHeader as an hstio error is set at the same time */
+        if (putHeader(imUpdate)) {
+            reportErrorAndFreeHeader (&header);
+            closeImage (imUpdate);
+            return (status = hstio_err());
```
jamienoss: This shouldn't be bypassing the handling by `putHeader()` it should be using it, i.e. the returned value.

pkg/acs/calacs/lib/mkspt.c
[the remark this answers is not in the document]
jamienoss: I'd open an issue for this and link it here for reference.

jamienoss:
@mdlpstsci I think you may have created some new warnings, e.g. see that [Travis log](https://travis-ci.org/spacetelescope/hstcal/jobs/320278958)

```
../pkg/acs/calacs/lib/mkspt.c:271:20: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
        sprintf (MsgText, hstio_errmsg());
                          ^~~~~~~~~~~~~~
/usr/include/secure/_stdio.h:47:56: note: expanded from macro 'sprintf'
  __builtin___sprintf_chk (str, 0, __darwin_obsz(str), __VA_ARGS__)
                                                       ^~~~~~~~~~~
../pkg/acs/calacs/lib/mkspt.c:271:20: note: treat the string as an argument to avoid this
        sprintf (MsgText, hstio_errmsg());
                          ^
                          "%s", 
/usr/include/secure/_stdio.h:47:56: note: expanded from macro 'sprintf'
  __builtin___sprintf_chk (str, 0, __darwin_obsz(str), __VA_ARGS__)
```

pkg/acs/calacs/lib/mkspt.c
mdlpstsci: I neglected to make a comment here previously.  IMHO the openUpdateImage function should close the file before it returns an error code to the invoking routine.  This issue falls into the category of cleanup of HSTIO.

jamienoss: _c.f._ #289
Looking at the code in more detail, this is the only solution since `BADDIMS` is not persistent and thus there is no means to disambiguate this error from another returned `NULL` ptr.

pkg/acs/calacs/lib/mkspt.c
mdlpstsci: I double-checked this today to verify my original rationale. If one drills down through the PutKeyXXX functions, the lowest level code invokes the HSTIO error() function which sets an error_status and an error_msg.  This new code prints the fundamental reason for the error to the trailer file, and returns (as you note) the HSTIO error value. All the intermediate functions between PutKeyXXX and the low-level functions return 0, -1, or NULL.  Some of the intermediate functions also call HSTIO error().  At the level just below PutKeyXXX, the HSTIO putKeyXXX returns 0/1 based upon hstio_err().  I felt this information was more informative for the user than the current situation.  I can add the error code, for what it is worth, to the reportErrorAndFreeHeader function.

All existing functionality invoking mkNewSpt is just checking for 0/non-zero.  Having said this, I agree that I perhaps returning hstio_err() may be questionable in a general sense.  More tomorrow.

mdlpstsci: Ultimately, I agree with you so status is now saved from the PutKeyXXX/putXXX functions and returned to the calling function on exit.  Before exiting, status is passed to a helper function with checks for specific errors in order to provide informational messages to the user.

pkg/acs/calacs/lib/mkspt.c
mdlpstsci: Ditto the previous PutKeyXXX discussion.

pkg/acs/calacs/lib/mkspt.c
mdlpstsci: OK

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -191,7 +229,50 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+
     c_imtclose (tpin);
 	/* Successful return */
 	return (status);
 }
+
+// Simple helper routine to report the HSTIO error to the .tra file and free 
+// the corresponding header 
+static void 
+reportErrorAndFreeHeader (Hdr *hdr)
```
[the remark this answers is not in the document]
mdlpstsci: Changed to invoke WhichError(status).

pkg/acs/calacs/lib/mkspt.c
mdlpstsci: In order to avoid duplicate output trailer messages, the status of the PutKeyXXX functions is obtained and passed back to the calling function on error.  The "helper" function is not necessary any longer and was deleted.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -143,8 +141,10 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
         */
 	    /* Read the primary header of the input _spt.fits file */
         nextn = 0;
-		if (LoadHdr (in_spt, &header) )
-            	    return (status);
+        if (LoadHdr (in_spt, &header)) {
+            WhichError (status);
```
jamienoss: Will this be reported twice?

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -167,53 +167,48 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+            if ((status = PutKeyInt (&header, "NEXTEND", nimgs, ""))) {
+                return (status);
             }
 
             /* Write the new _spt.fits file */
             /* Open the image; this also writes the header */
             im = openOutputImage (out_spt, "", 0, &header, 0, 0, FITSBYTE);
-            if (hstio_err()) {
+            if (!im || hstio_err()) {
+                status = OPEN_FAILED;
```
jamienoss: Just an FYI, since we're not going to call `WhichError()`, `status` can be set within in the return as it was, i.e. `return (status= OPEN_FAILED);`.

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -225,36 +220,36 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
         	freeShortHdrData(&stmp);
         }
 
-	    /* Close the header here... */
+        /* Close the header here... */
         freeHdr (&header);
+        initHdr (&header);
```
jamienoss: `freeHdr()` does the init for you.

```c
void freeHdr(Hdr *h) {
# if defined (DEBUG)
        printf("freeHdr: %x %d %d %x\n",
                (int)h,h->nlines,h->nalloc,(int)(h->array));
# endif
        if (!h)
        	return;
        if (h->array)
            free(h->array);
        initHdr(h);
}
```

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -167,53 +167,48 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
                 trlmessage (MsgText);
 		    }
 
-	        /* Update the FILENAME header keyword */
-            if (PutKeyStr (&header, "FILENAME", out_spt, "")) {
-                reportErrorAndFreeHeader (&header);
```
jamienoss: You've removed the error but why have you also removed the free? Don't this still leak now?

pkg/acs/calacs/lib/mkspt.c
```diff
@@ -187,11 +220,43 @@ int mkNewSpt (char *in_list, char *mtype, char *output) {
...
+
+        imUpdate = openUpdateImage (out_spt, "", 0, &header);
+        if (!imUpdate || hstio_err()) {
+            status = OPEN_FAILED;
+            trlopenerr (out_spt);
+            return (status);
+        }
+
+        if ((status = PutKeyInt (&header, "NEXTEND", extnum, ""))) {
+            closeImage (imUpdate);
```
jamienoss: `header` leaks here.

Issue #278, opened 2017-11-16
Improve command line usage text for executables


mdlpstsci:
While the ACS executables have existing usage messages, they are incomplete in that the options have no explanation (i.e., what does "-t" mean?).  An example from ACSREJ:

    if (argc < 3) {
        syntax_error ("acsrej input output [-t] [-v]");
        printf("                    [-shadcorr] [-crmask] [-newbias] \n ");
        printf("                    [-table <filename>] [-scale #]\n");
        printf("                    [-init (med|min)] [-sky (none|mode)]\n");
        printf("                    [-sigmas #] [-radius #] [-thresh #]\n");
        printf("                    [-pdq #]\n");
        return(status);
    }

A thorough usage message will mitigate confusion and misuse.

See also:

* https://github.com/spacetelescope/hstcal/issues/572

pllim:
If it is the same as `calacs.e`, `-t` means include timestamp and `-v` means verbose.

pllim:
http://acstools.readthedocs.io/en/latest/acsrej.html is meant as an user-facing documentation of HSTCAL usage because there is no (easy) way to build Sphinx doc with pure C code, so I stuck the doc stuff in ACSTOOLS. You are welcome to move the doc somewhere more obvious if you wish.

While you are at it, there should be a `.e` suffix for executable.

p.s. On hindsight, option that uses a word like `-shadcorr` should have been `--shadcorr` to be more consistent with common usage out there but that boat has sailed.

PR #288, opened 2018-01-08
All BIAS files are processed by CALACS to use only read noise for ERR…

### LINKED ISSUE #277: Discrepancies in ERR extension when processing bias frames with CALACS
While verifying the changes made to `ACSREJ` in #2, I noticed that for bias calibration frames (exptime=0) there is an existing bug in the current release of `hstcal` concerning the computation of values in the ERR extensions. There are currently two ways to batch process and combine files with the `calacs` codebase; using an [association file](http://www.stsci.edu/hst/acs/documents/handbooks/currentDHB/acs_Ch22.html#149623), or using the python wrapper [acstools](http://acstools.readthedocs.io/en/latest/) for `calacs` to process the files and then using python wrapper for [ACSREJ.](http://acstools.readthedocs.io/en/latest/acsrej.html) 

If the raw bias frames are combined using an association with only `DQICORR`,`BLEVCORR`,`CRCORR` set to `PERFORM` the error arrays are correctly computed. However, if the bias frames are combined by first processing each image individually with only `DQICORR` and `BLEVCORR` set to `PERFORM`, and then using the `acstools` wrapper for acsrej there is an additional Poisson-like term that is included in the error which should not be present for images with EXPTIME=0. The wrapper only serves to facilitate user input and in the end is a direct call to the `acsrej.e` executable (defined at [L130](https://github.com/spacetelescope/acstools/blob/3025736de199951fcc346ba95385309dac9ba077/acstools/acsrej.py#L130), called at [L187](https://github.com/spacetelescope/acstools/blob/3025736de199951fcc346ba95385309dac9ba077/acstools/acsrej.py#L187)), which is the same executable that `calacs` uses.

error_v_signal_hstcal_bias

In the figure above the two populations of blue dots (one for the read noise of each amplifier used to readout chip 2) are the correct ones, whereas the red ones have an additional poisson-like term.

I double checked to ensure that the problem only exists for the bias frames by performing the same analysis on a set of dark frames with an exptime of 1000.5s. The two images produced are essentially identical with only 0.0025% of the pixels differing by more than 0.1e in a 1000's exposure where the typical pixel has a value less than 140 e-. The two plots below show the pixel values in the final images produced by each process; the first is the full spectrum, the second is zoomed in to show the ~97% of pixels with values less than 140e- which is the the hot pixel threshold in 1000.5s dark frames.
error_v_signal_hstcal_dark

error_v_signal_hstcal_dark_zoomed

mdlpstsci:
All BIAS files are now processed by CALACS to use only read noise (vs Poisson) for the computation of the ERR array as the date discriminant (pre- and post-SM4) has been removed. Also, when using ACSREJ standalone, the user now must use the "readnoise_only" flag (formerly "newbias" flag) to obtain the same results as CALACS for BIAS images.  NOTE: Modifying the flag to ACSREJ is expeditious rather than user-friendly.  The "newbias" flag has been removed (not just deprecated) as this functionality is used mostly internally for calibration purposes.  If the "newbias" flag is invoked, ACSREJ will issue an message regarding the change and exit with an error.

When updating ACSREJ for standalone use, improved the syntax or usage message which addresses a portion of #278 for ACSREJ only.

Resolves #277. (EDIT: Added a space for proper cross-linking.)
Addresses some components of #278 for ACSREJ only.

### DIFF
--- a/pkg/acs/calacs/acsrej/rej_command.c
+++ b/pkg/acs/calacs/acsrej/rej_command.c
@@ -47,12 +47,27 @@ int newpar[];       o: array of parameters set by the user
         List all options for user
     */
     if (argc < 3) {
-        syntax_error ("acsrej input output [-t] [-v]");
-        printf("                    [-shadcorr] [-crmask] [-newbias] \n ");
+        syntax_error ("acsrej.e input output [-t] [-v]");
+        printf("                    [-shadcorr] [-crmask] [-readnoise_only] \n");
         printf("                    [-table <filename>] [-scale #]\n");
         printf("                    [-init (med|min)] [-sky (none|mode)]\n");
         printf("                    [-sigmas #] [-radius #] [-thresh #]\n");
-        printf("                    [-pdq #]\n");
+        printf("                    [-pdq #]\n\n");
+		printf("input             comma-delimited string (e.g., filename or filename1,filename2,filename3)\n");
+		printf("output            string\n");
+		printf("-t                print timestamps\n");
+		printf("-v                turn on verbose mode\n");
+		printf("-shadcorr         turn on shutter shading correction (intended for CCD images only)\n");
+		printf("-crmask           flag CR-rejected pixels in the input files\n");
+		printf("-readnoise_only   use read noise and not Poission noise to create ERR array (intended for BIAS images)\n");
+		printf("-table filename   cosmic ray rejection table to use for default parameter values\n");
+		printf("-scale #          multiplicative factor (in percent) applied to noise\n");
+		printf("-init med|min     scheme for computing initial guess image (median or minimum)\n");
+		printf("-sky none|mode    scheme for computing sky levels to be subtracted (none:0.0 or mode:most frequently occurring value)\n");
+		printf("-sigmas #[,#...]  cosmic ray rejection thresholds, no. of thresholds are the no. of iterations\n");
+		printf("-radius #         radius (in pixels) to propagate the cosmic ray\n");
+		printf("-thresh #         cosmic ray rejection propagation threshold\n");
+		printf("-pdq #            data quality flag used for cosmic ray rejection\n\n");
         return(status);
     }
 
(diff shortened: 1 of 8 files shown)


pkg/acs/calacs/acsrej/rej_command.c
```diff
@@ -47,12 +47,27 @@ int newpar[];       o: array of parameters set by the user
         List all options for user
     */
     if (argc < 3) {
-        syntax_error ("acsrej input output [-t] [-v]");
-        printf("                    [-shadcorr] [-crmask] [-newbias] \n ");
+        syntax_error ("acsrej.e input output [-t] [-v]");
+        printf("                    [-shadcorr] [-crmask] [-readnoise_only] \n");
```
pllim: This needs a corresponding change over at ACSTOOLS (e.g., https://github.com/spacetelescope/acstools/blob/master/acstools/pars/acsrej.cfg).

jamienoss: My input: I would argue that this 1-1 API mapping creates an unnecessary dependency. I would state that it is better to use the, recently added, 'transparent window' `exe_args` and then just make this issue one of documentation only (since the ACS HSTCAL docs exist as doc strings in acstools). However, I understand that since the other parameters are mirrored the precedence is already set and it's already too late.

pkg/acs/calacs/acsrej/rej_command.c
```diff
@@ -89,8 +104,22 @@ int newpar[];       o: array of parameters set by the user
...
+            // This option/variable was previously named "newbias", and it has been renamed
+            // to "readnoise_only" which is more informative.  The variable controls
+            // the type of noise used in the computation of the error (ERR) array.  In particular,
+            // this variable is intended to control the noise used when processing BIAS data.  When
+            // using CALACS, readnoise_only will only be set to "1" (true) for BIAS data.  When using
+            // ACSREJ as a standalone utility, the readnoise_only can be set to "1" for any data.
             } else if (strcmp("newbias", argv[ctoken]+1) == 0) {
-                par->newbias = 1;
+                printf ("\n*************************************************************************************\n");
+                printf ("     As of HSTCAL Build 2018.1, the 'newbias' option for ACSREJ has been renamed \n");
```
jamienoss: I'm not sure of our official stance on this but we may need to refer to HSTCAL by something other than our internal build identifier (2018.1), i.e. the actual version number. This definitely needs to be done, I'm just not sure if we need both or just the latter. It may even need to also reference the version of CALACS.

pllim:
Corresponding Python wrapper changes are in spacetelescope/acstools#51. FYI.

Issue #168, opened 2017-07-05
CALACS: Enable bias shift for new subarrays


pllim:
Future planning mentioned in ACS meeting on July 5, 2017. Wait for official green light before implementation.

As of 2017.2a, bias shift is only applied to post-SM4 WFC fullframe exposures. ACS Team wants bias shift to also be applied to post-SM4 WFC "new" subarrays (the ones with same readout timing as fullframe; don't forget the polarizers). Currently, there are some hardcoded values in [blev_funcs_postsm4.c](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/acsccd/blev_funcs_postsm4.c); These need to be generalized somehow to support subarrays.

Also see #167 for another bias shift related feature request.

jamienoss:
@pllim Hey, did this get implemented? The acs team seem to think so but there are no PR links here?

mdlpstsci:
@nagrogin
This issue (GIT #168 CALACS: Enable bias shift for new subarrays) is what you believe has already been implemented in CALACS.  It is incorporated in CALACS, *but* it only applies to full frame exposures.  It was assigned to me two days after I started.  Unfortunately, it has been in limbo in the transition shuffle between Pey-Lian and me with regard to the priority of the implementation, as well as the official “proceed” to implement.   Having said this, I am now responsible for this issue and the lack of action.

Pey-Lian has now provided me with the background ISR (ACS 2012-02) on this topic.  Realistically, I do not believe this can be researched, implemented, and tested in time for 2018.1.   I first need to assess what needs to be done, if anything, beyond CALACS code changes.   The assessment is needed to determine if new keywords, or new entries in existing configuration files are necessary.  

I just spoke to DMS (Faith) regarding this situation.  She indicated if there are no new keywords or changes in configuration files, and only CALACS code changes which would affect the contents of the output files, this could be considered a “quick” fix (2018.1a).  Build 2018.2 (DMS database and infrastructure) will happen ~April.  As such, any 2018.1a and validation of the ACS data would need to happen prior to April.

PR #312, opened 2018-05-29
Update ACS BLEVCORR routines to accommodate new supported subarrays

mdlpstsci:
Addresses IT #168 (EDIT: Fixed linking)

The new supported subarrays (circa mid-2016) need to have the improved bias level processing applied during CALACS processing akin to what is done for the full frame data.  In detail only the bias shift correction is applicable to subarray data.  This is currently a work in progress in that only the new WFC 2K subarrays are handled.  The WFC 512 and 1K subarrays do not have virtual overscan, so there is no "magic square" data.  This issue will be addressed when an alternative to the "magic square" is defined.  Also, the polarizer and ramp subarrays are not yet handled.  

It should be noted the block of code handling the logic of which BLEV algorithm (original or new and improved) should be applied and the data appropriate for each algorithm has been re-written for clarity and at the request of the team lead.  While it is more verbose, the logic is cleaner.


pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -81,102 +117,137 @@ int bias_shift_corr(ACSInfo *acs, SingleGroup *chip2, SingleGroup *chip1) {
...
-    num = 0;
+      /* calculate "magic square mean" - this works for any single amp */
+      sum = 0.0;
+      num = 0;
     
-    for (j = 2057; j <= 2066; j++) {
-      for (k = 13; k <= 22; k++) {
-        sum += ampdata[arr_cols*j + k];
-        num++;
+      for (j = 2057; j <= 2066; j++) {
```
pllim: Even prior to this PR, I was uncomfortable with these magic numbers. Perhaps it is time to make them constant variables with meaningful names?

This comment also applies to other occurrences of magic numbers in this file.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -32,6 +35,8 @@ static void dqiMsg (ACSInfo *, int);
 static void BiasMsg (ACSInfo *, int);
 static void BlevMsg (ACSInfo *, int);
 static void SinkMsg (ACSInfo *, int);
+/* These are the subarray supported apertures for the bias shift correction */
+static const char *subApertures[] = {"WFC1A-2K", "WFC1B-2K", "WFC2C-2K", "WFC2D-2K"};
```
pllim: If you ever want to add polarizers, keep in mind that the aperture name never changed after flight software update, so same POL aperture is "old subarray" before that update, and "new subarray" after. Becareful with those.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -251,109 +257,291 @@ int DoCCD (ACSInfo *acs_info) {
...
+       apertures.  Use the aperture name and not the date as the discriminant 
+       as the date is fuzzy - some supported subarrays were acquired before the 
+       official change, and some old style subarrays were acquired after the change.
 
-            if (doBlev(&acs[i], &x[i], acs[i].chip, &meanblev, &done,
-                       &driftcorr))
-                return (status);
+       ***FIX MDD: What about polarizer and ramp apertures?
+    */
+    Bool supportedSubAperture = NO;
```
pllim: I feel that this should be inside the logic for WFC only, because this same code is also used to process HRC.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -251,109 +257,291 @@ int DoCCD (ACSInfo *acs_info) {
...
+                     driftcorr = NO;
+                  } /* End subarray processing */
+
+            } /* End of full frame/subarray data */
+
+          } /* End of Pre-/Post-SM4 data */
+
+       } /* End of only WFC detector */
+
+       /* All HRC and SBC data - use the original blev correction */
```
pllim: SBC data don't get processed here. This is only for CCD, not MAMA. (Please correct me if I misremember.)

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,13 +37,41 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+ *
+ * chip2 consists of amps C & D, chip1 consists of amps A & B. 
+ *
+ * Because full frame data uses both chips, but supported subarray data
+ * is only from a single amp (therefore, a portion of a single chip), this
+ * routine has a variable number of input SingleGroups.
+ *
+ * 16-May-2018 M.D. De La Pena: Generalized bias_shift_corr() to handle subarray data.
+ */
+int bias_shift_corr(ACSInfo *acs, int nGroups, ...) {
```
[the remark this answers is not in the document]
mdlpstsci: ...So much easier and more obvious with an OO language.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -81,102 +117,137 @@ int bias_shift_corr(ACSInfo *acs, SingleGroup *chip2, SingleGroup *chip1) {
...
-    if (i < 2) {
-      make_amp_array(arr_rows, arr_cols, chip1, i, ampdata);
-    } else {
-      make_amp_array(arr_rows, arr_cols, chip2, i, ampdata);
-    }
+
+  /* For full frame data, there are four amps in use.  For supported subarray
+   * data, there is only one amp in use.  The actual amp in use must be determined. 
+   */
+  int numAmpsInUse = (acs->subarray == YES) ? 1 : NAMPS;
```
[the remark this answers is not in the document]
mdlpstsci: NAMPS is hard-coded to 4 which nicely accommodates full frame data.  However, the new, "supported" WFC subarrays are single amp (512, 1K, or 2K).   This may have to be updated once the POL and RAMP data are incorporated into this new algorithm.  All other WFC subarrays will use the old blev algorithm.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
mdlpstsci: This is already in progress. An updated overscan table has been created with primary header keywords and a new binary table in the second extension to accommodate amp-dependent values.

pkg/acs/calacs/acsccd/doccd.c
mdlpstsci: Ooops.  I got a bit too inclusive in my comments - here there be no SBC.

pkg/acs/calacs/acsccd/doccd.c
mdlpstsci: Indeed!

mdlpstsci:
Polarized and ramp data are now handled by this upgrade.  The new WFC 512 and 1K subarrays are not yet accommodated by this upgrade as they do not have virtual overscan, so there is no "magic square" data to use.  The team needs to determine an alternative to the "magic square".  The additional work to handle the WFC 512 and 1K subarrays will be handled by a new issue.

mdlpstsci:
Nate (ACS Team) sent on 07 August 2018:
"TL;DR: I approve the updates for the bias shift as they properly handle full-frame Saturn data (Proposal ID: 12395) that was forced into the correct subarray format for the new bias-shift algorithm to be applied in BLEVCORR."

mdlpstsci:
@pllim Can you please re-review?  Basically, I added a lot of code to (1) handle all the cases for the full frame and subarray data with regard to the bias shift correction, and (2) free memory.  I will note the logic for application of the bias shift correction has been expanded at the request of the ACS Team lead.  The code is now rather verbose and repetitive, but it is more approachable to an astronomer.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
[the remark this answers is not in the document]
jamienoss: This may have been worth hiding within a func, in case the logic becomes more complicated in the future.

pkg/acs/calacs/acsccd/doccd.c
jamienoss: FYI: I've been trying to move stuff over to using ``stdbool.h::bool`` instead.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+  /* Code to handle variable number of input parameters.  The number of SingleGroup is 2 when 
+     processing a full frame (2 chips - all 4 amps) and only 1 when processing a supported subarray 
+     (1 chip from a single amp).
+  */
+  SingleGroup *sg[nGroups];
+  va_list arguments;
+
+  /* Initializing arguments to store all the values being passed to this routine via "..." */
+  va_start (arguments, nGroups);
+  {
```
jamienoss: Why the extra scope?

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
   const int ngap_pix = (int) (serial_freq * parallel_shift + 0.5);
   
-  const int arr_rows = chip2->sci.data.ny;
-  const int arr_cols = chip2->sci.data.nx/2;
+  /* There is always at least one SingleGroup - use sg[0] to get the image size of a single amp.
+   * If this is a full frame image, then "nx" represents all the columns across both amps plus
+   * the physical overscan and must be divided by 2.  If this is a subarray, then "nx" represents
+   * just the columns for a single amp plus physical overscan.
+   */
+  const int arr_rows = sg[0]->sci.data.ny;
```
jamienoss: ``sg[0]`` is being accessed without first being checked. This can be done here, or better still when first assigned to ``sg`` so as to shrink it to only valid ptrs (bookkeeping required of ``nGroups`` etc).

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+  SingleGroup *sg[nGroups];
+  va_list arguments;
+
+  /* Initializing arguments to store all the values being passed to this routine via "..." */
+  va_start (arguments, nGroups);
+  {
+       /* sg[0] is chip2 and sg[1] is chip 1 for a full frame image */
+       {unsigned int i;
+       for (i = 0; i < nGroups; i++) {
+           sg[i]= va_arg (arguments, SingleGroup *);
```
jamienoss: You may want to check that they're not ``NULL`` here, see comment below.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+ *
+ * chip2 consists of amps C & D, chip1 consists of amps A & B. 
+ *
+ * Because full frame data uses both chips, but supported subarray data
+ * is only from a single amp (therefore, a portion of a single chip), this
+ * routine has a variable number of input SingleGroups.
+ *
+ * 16-May-2018 M.D. De La Pena: Generalized bias_shift_corr() to handle subarray data.
+ */
+int bias_shift_corr(ACSInfo *acs, int nGroups, ...) {
```
jamienoss: Add ``assert(acs);``.

jamienoss: Also:
```c
if (nGroups < 1)
    return HSTCAL_OK; // or NOTHING_TO_DO, NO_GOOD_DATA, INTERNAL_ERROR
```

jamienoss: Actually, if going for ``INTERNAL_ERROR`` I would instead ``assert``. This is a much better way to handle internal errors as it gives the src line in question.

jamienoss: I tend to treat this kind of thing as a NOP and would return ``HSTCAL_OK``,  upto you to think about the function's semantics.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
jamienoss: Perhaps you want to scope ``arguments``, which is a very good idea, in which case this needs to be moved up just above `` va_list arguments;`` and the closing, just after ``va_end (arguments);``.

jamienoss: If that was not your intention, I would scope ``arguments``.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
   
-  const int arr_rows = chip2->sci.data.ny;
-  const int arr_cols = chip2->sci.data.nx/2;
+  /* There is always at least one SingleGroup - use sg[0] to get the image size of a single amp.
+   * If this is a full frame image, then "nx" represents all the columns across both amps plus
+   * the physical overscan and must be divided by 2.  If this is a subarray, then "nx" represents
+   * just the columns for a single amp plus physical overscan.
+   */
+  const int arr_rows = sg[0]->sci.data.ny;
+  const int arr_cols = (acs->subarray == NO) ? sg[0]->sci.data.nx/2 : sg[0]->sci.data.nx;
```
jamienoss: Rename ``arr_cols`` more semantically specific, e.g. ``ampArrCols`` etc.

jamienoss: FYI: (I know the mix already existed) but I don't like the mix of camel case and underscores.

jamienoss: Regarding the logic: I'm only now getting the impression that the ``singleGroups`` being passed in actually amps. Is this correct? So what does this func act on amps or...? If amps, then ``nx`` should already be correct, right? Then why the ``/2``?

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
   const int nquad_pix = (arr_cols + ngap_pix) * arr_rows;
   
-  /* array of true DC bias levels */
-  double * dc_bias_levels;
-  
   /* arrays below are in order of amp A, B, C, D */
   
-  /* time constant of AC high pass filter in external pre-amp. */
+  /* time constant of AC high pass filter in external pre-amp in microseconds */
   const double time_const[NAMPS] = {37.290251, 36.180001, 37.867770, 42.461249};
```
jamienoss: If in microsecs, reflect this in the param name, e.g. ``usecTimeConst``.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+  addPtr(&ptrReg, dc_bias_levels, &free);
     
-    for (j = 2057; j <= 2066; j++) {
-      for (k = 13; k <= 22; k++) {
-        sum += ampdata[arr_cols*j + k];
-        num++;
+  /* array for amp data and amp data + gap pixels */
+  double *ampdata = malloc(arr_rows * arr_cols * sizeof(double));
+  addPtr(&ptrReg, ampdata, &free);
+  double *ampdata_gap = malloc(nquad_pix * sizeof(double));
```
jamienoss: All of the above ``malloc``s need to be ``NULL`` checked handling ``OUT_OF_MEMORY`` errors.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
-  int num;
-  double magic_square_mean;
-  
-  /* allocate space for data arrays */
-  ampdata = malloc(arr_rows * arr_cols * sizeof(double));
-  ampdata_gap = malloc(nquad_pix * sizeof(double));
-  
-  /* allocate space for true DC bias levels array */
+
+  PtrRegister ptrReg;
```
jamienoss: 👍

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+   * data, there is only one amp in use.  The actual amp in use must be determined. 
+   */
+  int numAmpsInUse = (acs->subarray == YES) ? 1 : NAMPS;
+  
+  /* Loop over all the amps for full frame data (total of 4 amps, 0 to 3) or just use 
+   * the single amp value corresponding to the supported subarray (1 amp, value defined in acs.h).
+   */
+  int ampInUse = 0;
+  {unsigned int i;
+  for (i = 0; i < numAmpsInUse; i++) {
```
jamienoss: This should be a function and I believe it already is, ``parseWFCamps()`` ([src](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/lib/parseamps.c#L14-L44)) and should be used in conjunction with ``AMPSORDER`` ([src](https://github.com/spacetelescope/hstcal/blob/master/pkg/acs/calacs/include/acs.h#L75-L84)).

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+         else if (strcmp (acs->ccdamp, "B") == 0)
+            ampInUse = AMP_B;
+         else if (strcmp (acs->ccdamp, "C") == 0)
+            ampInUse = AMP_C;
+         else if (strcmp (acs->ccdamp, "D") == 0)
+            ampInUse = AMP_D;
+      } 
+
+      /* put a single amp's data into ampdata */
+      if (acs->subarray == YES) {
```
jamienoss: I'm getting the impression again that this func is designed to work on objects that _are_ amps. A better design would have been to have this function's intended use to be only on amps and have the chip to amp split outside, within a wrapper. The logic would be so much cleaner as it would remove the continuous subarray vs full frame checking/special-casing/handling etc.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
       }
-    }
+       
+      /* calculate "magic square mean" - this works for any single amp */
+      double sum = 0.0;
+      int num = 0;
+      double magic_square_mean = 0.0;
     
-    magic_square_mean = sum / (double) num;
+      {unsigned int j, k;
```
jamienoss: This would be better as a function, in which case the hardcoded boundaries may be less of an issue.

jamienoss: Is there a way to check that the amp being passed in even has this region? Should the check be in ``bias_shift_corr()``, I would have perhaps thought so.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
mdlpstsci: Indeed.  Went to great trouble to add these checks in other routines, but I neglected to add the checks here.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+            ampInUse = AMP_B;
+         else if (strcmp (acs->ccdamp, "C") == 0)
+            ampInUse = AMP_C;
+         else if (strcmp (acs->ccdamp, "D") == 0)
+            ampInUse = AMP_D;
+      } 
+
+      /* put a single amp's data into ampdata */
+      if (acs->subarray == YES) {
+         make_amp_array(arr_rows, arr_cols, sg[0], ampInUse, ampdata);
```
jamienoss: (@mdlpstsci & I discussed offline so will just jump to the point) The indexing here (and below) is hardcoded yet ``sg`` has length ``nGroups`` and is populated variadically.
If this func is limited to the number of ``SingleGroups`` it can take (especially if only a few, i.e. 2) it should not be variadic.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
-        } else {
-          ampdata_gap[(arr_cols + ngap_pix)*j + k] = magic_square_mean;
+      magic_square_mean = sum / (double) num;
+    
+      /* make amp + gap array */
+      {unsigned int j, k;
+      for (j = 0; j < arr_rows; j++) {
+          for (k = 0; k < (arr_cols + ngap_pix); k++) {
+              if (k < arr_cols) {
+                 ampdata_gap[(arr_cols + ngap_pix)*j + k] = ampdata[arr_cols*j + k];
```
jamienoss: I believe the preference (within HSTCAL) is to use the macro ``Pix(a,i,j)`` ([src](https://github.com/spacetelescope/hstcal/blob/master/include/hstio.h#L188-L196)) for readability, just make sure that when it's expanded you have the order correct.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
-      for (k = 0; k < (arr_cols + ngap_pix); k++) {
-        if (k < arr_cols) {
-          ampdata_gap[(arr_cols + ngap_pix)*j + k] = ampdata[arr_cols*j + k];
-        } else {
-          ampdata_gap[(arr_cols + ngap_pix)*j + k] = magic_square_mean;
+      magic_square_mean = sum / (double) num;
+    
+      /* make amp + gap array */
+      {unsigned int j, k;
+      for (j = 0; j < arr_rows; j++) {
```
jamienoss: Minor: if ``j`` & ``k`` are not an extension in dim to ``i`` (which is the iter for the outer loop) it is clearer to use another canonical set, e.g. ``m`` & ``n``.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+              }
         }
-      }
-    }
+      }}
 
-    /* calculate true DC bias levels */
-    factor = 1.0 - exp(-1.0 / (time_const[i] * serial_freq));
+      /* factor combining time constant and clocking frequency */
+      double factor = 1.0 - exp(-1.0 / (time_const[ampInUse] * serial_freq));
```
jamienoss: const

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +39,218 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
-    }
+      }}
     
-    /* copy corrected data back to ampdata */
-    for (j = 0; j < arr_rows; j++) {
-      for (k = 0; k < arr_cols; k++) {
-        ampdata[arr_cols*j + k] = ampdata_gap[(arr_cols + ngap_pix)*j + k];
-      }
-    }
+      /* copy corrected data back to ampdata */
```
jamienoss: Is it not possible to combine this op within the loop above (correction comp) and assign the correction directly to  ``ampdata``? This would be more efficient.

pkg/acs/calacs/acsccd/findover.c
```diff
@@ -129,33 +133,34 @@ static int CloseOverTab (TblInfo *);
...
-    int row;
     TblInfo tabinfo;
     TblRow tabrow;
-    int foundit;
 
     int SameInt (int, int);
     int SameString (char *, char *);
 
+    int foundit = NO;
+    *overscan = NO;
```
jamienoss: Godo job moving this to before the return.

pkg/acs/calacs/lib/getacskeys.c
```diff
@@ -46,14 +46,23 @@ int GetACSKeys (ACSInfo *acs, Hdr *phdr) {
...
 	if (GetKeyStr (phdr, "ROOTNAME", NO_DEFAULT, "",
                  acs->rootname, ACS_CBUF))
-    return (status);
+       return (status);
   
 	if (GetKeyStr (phdr, "APERTURE", USE_DEFAULT, "", acs->aperture, ACS_CBUF))
-    return (status);
+       return (status);
+    {unsigned int i;
+       for (i=0; i < strlen(acs->aperture); i++) {
```
jamienoss: indentation

pkg/acs/calacs/lib/getacskeys.c
```diff
@@ -46,14 +46,23 @@ int GetACSKeys (ACSInfo *acs, Hdr *phdr) {
   
 	if (GetKeyStr (phdr, "ROOTNAME", NO_DEFAULT, "",
                  acs->rootname, ACS_CBUF))
-    return (status);
+       return (status);
   
 	if (GetKeyStr (phdr, "APERTURE", USE_DEFAULT, "", acs->aperture, ACS_CBUF))
-    return (status);
+       return (status);
+    {unsigned int i;
```
jamienoss: This should be a func ``toUpper()`` that excepts a string. E.g. use this [utility in cfitsio](https://heasarc.gsfc.nasa.gov/docs/software/fitsio/c/c_user/node52.html) ``Convert a character string to uppercase (operates in place).
  void fits_uppercase / ffupch (char *string)``.

pkg/acs/calacs/lib/getacskeys.c
```diff
@@ -46,14 +46,23 @@ int GetACSKeys (ACSInfo *acs, Hdr *phdr) {
...
+       for (i=0; i < strlen(acs->aperture); i++) {
+           acs->aperture[i] = toupper (acs->aperture[i]);
```
jamienoss: This is only acceptable/valid if it is "illegal" for the char set to be [a-z], if not it must be removed. Even if it is I really don't think it is this function's responsibility to do so.

jamienoss: Imagine someone wanting to write a validity wrapper to ``GetACSKeys()``, they now can't because the functionality/semantics has been mangled.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -42,25 +64,13 @@ int DoCCD (ACSInfo *acs_info) {
...
-    int * blevcorr;
     char buff[ACS_FITS_REC+1];
-    Bool subarray;
 
     int to_electrons(ACSInfo *, SingleGroup *);
     int doBias (ACSInfo *, SingleGroup *);
     int biasHistory (ACSInfo *, Hdr *);
     int doBlev (ACSInfo *, SingleGroup *, int, float *, int *, int *);
-    int bias_shift_corr(ACSInfo *, SingleGroup *, SingleGroup *);
+    int bias_shift_corr(ACSInfo *, int, ...);
```
jamienoss: This should be in a header. It should also have a comment describing the variadic arg.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -81,22 +91,35 @@ int DoCCD (ACSInfo *acs_info) {
...
     int GetKeyBool (Hdr *, char *, int, Bool, Bool *);
 
     /*========================Start Code here =======================*/
 
+    PtrRegister ptrReg;
+    initPtrRegister(&ptrReg);
+
     /* set up an array of SingleGroups to hold all image extensions */
+    SingleGroup * x = NULL;   /* used for both input and output */
     x = (SingleGroup *) malloc(acs_info->nimsets * sizeof(SingleGroup));
```
jamienoss: Would have been good to have taken the opportunity and removed the cast.

jamienoss: and also changed ``sizeof(SingleGroup)`` -> ``sizeof(*x)``

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -81,22 +91,35 @@ int DoCCD (ACSInfo *acs_info) {
...
     int GetCCDTab (ACSInfo *, int, int);
     int GetKeyBool (Hdr *, char *, int, Bool, Bool *);
 
     /*========================Start Code here =======================*/
 
+    PtrRegister ptrReg;
+    initPtrRegister(&ptrReg);
+
     /* set up an array of SingleGroups to hold all image extensions */
+    SingleGroup * x = NULL;   /* used for both input and output */
```
jamienoss: This should be an actual object instance i.e. ``SingleGroup  x;``

pkg/acs/calacs/acsccd/doccd.c
jamienoss: See comment above about instantiation, you shouldn't be allocating for a struct like this.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -81,22 +91,35 @@ int DoCCD (ACSInfo *acs_info) {
...
     x = (SingleGroup *) malloc(acs_info->nimsets * sizeof(SingleGroup));
+    if (!x)
+       return (status = OUT_OF_MEMORY);
+    addPtr(&ptrReg, x, &free);
```
jamienoss: This is ok as is because the freeing happens in reverse order of that added. However, I'm wondering if this is "ok". Hmmm...

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -143,278 +171,582 @@ int DoCCD (ACSInfo *acs_info) {
...
+        blevcorr[i] = PERFORM;
+    }}
+
+    /* The logic here has been re-written (1) to accommodate the new subarrays
+       (Ref: ISR ACS 2017-03) which can be bias shift corrected ("*-2K" data only at 
+       this time), and (2) the head of the ACS team requested the logic be clarified.  
+       This code is verbose and potentially harder to maintain, but the logic is clearer
+       for the scientist reading the code.
+
+       NOTE: The variable "done" is overloaded.  When done = overscan[i], this means
```
jamienoss: Can't we un-overload and instead of any ``if (done)`` use if (blevDone || overscan[i])``? Not 100% certain of ``||`` vs ``&&`` as I haven't read any of the below logic.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -143,278 +171,582 @@ int DoCCD (ACSInfo *acs_info) {
...
+       from CCDTAB.
+    */
 
-                if (PutKeyFlt(&x[i].sci.hdr, "MEANBLEV", meanblev,
-                              "mean of bias levels subtracted in electrons")) {
+    int done = NO;       
+    int driftcorr = NO;     /* true means bias level was corrected for drift */
+    Bool isSupportedSubAperture = NO;
+    float meanblev = 0.0;   /* mean value of overscan bias (for history) */
+    if (acs_info->blevcorr == PERFORM) {
```
jamienoss: ``if (!OmitStep(acs_info->blevcorr))``

jamienoss: I would pull this entire block out into a sep func. This would help clear up the crowded logic. Same prob goes for all the other steps.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
mdlpstsci: Yes.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -143,278 +171,582 @@ int DoCCD (ACSInfo *acs_info) {
...
+                    if (virtOverscan[0]) {
+                       isSupportedSubAperture = YES;
+                       trlmessage("This subarray data is supported for the bias shift correction.");
+                    }
+                 }
+             }}
+          }
+
+          /* Only Pre-SM4 WFC data - this is the original bias level subtraction */
+          if (acs_info->expstart < SM4MJD) {
```
jamienoss: A this stage of me reading this huge logic block(s) I would reduce all of them to calls. This would make things infinitely clearer.
E.g.
```c
if (acs_info->expstart < SM4MJD)
    doSM4MJD(...);
else
    doPostSM4MJD(...);
```

jamienoss: The nesting could also be described as a comment here at the root level. Maybe even inline e.g.
```c
if (acs_info->expstart < SM4MJD)
    doSM4MJD(...); // -> if (subaray)
                                //          doSub();
else
    doPostSM4MJD(...);
```

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -143,278 +171,582 @@ int DoCCD (ACSInfo *acs_info) {
...
+                       for non-polarization or non-ramp data, but the comparison works for 
+                       this data too.  Only need to check one virtOverscan value in the array. 
+
+                       NOTE:Just because the aperture is supported does not mean the bias
+                       shift correction will be applied.  The DS_int and gain criteria also
+                       have to be satisfied.
+                    */
+                    if (virtOverscan[0]) {
+                       isSupportedSubAperture = YES;
+                       trlmessage("This subarray data is supported for the bias shift correction.");
```
jamienoss: Do you want a ``break`` here?

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
mdlpstsci: Keeping the arr_cols vs ampArrCols as the remaining functions in this file also use this variable name for a total of 87 name changes.  I want to keep the functions consistent in their terminology.

pkg/acs/calacs/acsccd/doccd.c
```diff
@@ -143,278 +171,582 @@ int DoCCD (ACSInfo *acs_info) {
...
+
+          /* Only Pre-SM4 WFC data - this is the original bias level subtraction */
+          if (acs_info->expstart < SM4MJD) {
+
+             {unsigned int i;
+             for (i = 0; i < acs_info->nimsets; i++) {
+
+                 /* The variable done is set to the results of FindOver indicating there is
+                    physical overscan. */
+                 done = overscan[i];
```
jamienoss: Making this a  note since you didn't edit ``doBlev()``. It's better to  not overload params, i.e. separate those that are input and those that are output. In this case it's ``done``. This should either be past in (As const) or a ptr so that it can be modified as output, but never both.

pkg/acs/calacs/acsccd/doccd.c
jamienoss: Actually, perhaps it might be possible (and better?) to use a switch and order such that the logic exists as case stmts dropping through?

mdlpstsci:
Addressing the mallocs and checking the sg[] pointers before actual use.  New IT #334 opened which refer to the remaining issues in this PR.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +40,243 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+ * *** This routine should be modified according to issues discussed in PR #312 and noted in IT #334. ***
+ *
+ * 16-May-2018 M.D. De La Pena: Generalized bias_shift_corr() to handle subarray data.
+ */
+int bias_shift_corr(ACSInfo *acs, int nGroups, ...) {
+
+  /* This function has built in assumptions regarding the data being passed */
+  assert (acs != NULL);  /* Make sure the ACSInfo structure is defined */
+  assert ((nGroups == 1) || (nGroups == 2)); /* Must be at least one and no more than two SingleGroups */
+  assert (((nGroups == 1) && (acs->subarray == YES)) || ((nGroups == 2) && (acs->subarray == NO)));  /* Subarray has one and full frame has two SingleGroups */
```
jamienoss: Just so I've said it - this and the line above are ugly. The variadic arg should have been removed instead and these asserts handled by the compiler (at compile time).

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
```diff
@@ -36,152 +40,243 @@ static int make_amp_array(const int arr_rows, const int arr_cols, SingleGroup *
...
+   * The nGroups value and the number of SingleGroups actually passed must match. 
+   */
+  va_start (arguments, nGroups);
+  {
+       /* sg[0] is chip2 and sg[1] is chip 1 for a full frame image */
+       {unsigned int i;
+       for (i = 0; i < nGroups; i++) {
+           sg[i] = NULL;
+           sg[i] = va_arg (arguments, SingleGroup *);
+           if (!sg[i]) {
```
jamienoss: This is not what I had in mind. I meant that if ``va_arg (arguments, SingleGroup *) == NULL`` it just wouldn't be added, i.e. 
```c
SingleGroup * tmpGroup = va_arg (arguments, SingleGroup *);
if (tmpGroup)
    sg[i] = tmpGroup;
else
    continue;
```
You could argue that in such a case ``nGroups`` is actually incorrect, it is unclear of the use of the function params now that they are variadic. I.e. are they explicit slots, e.g. ``SingleGroup * amp2, SingleGroup * amp1`` or is the order irrelevant and in the case that there is only one, it must always be the first, e.g. not passed as ``NULL, amp1``.

jamienoss: I think the rest of the code assumes a specific ordering.

pkg/acs/calacs/acsccd/blev_funcs_postsm4.c
mdlpstsci: This scheme is used in the acsccd and acscte modules where the specific WFC amp has been remapped into a 1D array.  The knowledge of the location of the physical and virtual overscan regions, as well as the use of the Pix() structure, is in the remapping routine.  In essence, regardless of the actual layout of the specific amp, the remapping into this 1D array allows all the amps to be treated in the same fashion by this routine.  Another motivator is that the coding of this routine is very closely based upon a description laid out in [ISR ACS 2012-02 (http://www.stsci.edu/hst/acs/documents/isrs/isr1202.pdf)

jamienoss: Breadcrumb: They're never _not_ 1D arrays, it's C. The re-mapping is only the rotation of the "amps" such that the position of the amplifiers all have the same logical position within the above array. This has nothing to do with my style comment about not having "naked" array equations.

PR #313, opened 2018-06-01
ACS CTE: Add forward modeling functionality

jamienoss:
The cmd line option ``--forwardModelOnly``
piggy-backs upon the normal CTE code and usage (gen2 only) however,
instead of correcting the CTE trails it actually simulates (adds) them.

All CTE flags and ref files must be defined in the exact same
manner needed to run the actual CTE correction.

``PCTECORR`` must be set to ``PERFORM``.

### DIFF
--- a/pkg/acs/calacs/acscte/dopcte-gen2.c
+++ b/pkg/acs/calacs/acscte/dopcte-gen2.c
@@ -2,6 +2,7 @@
 #include <string.h>
 #include <stdlib.h>
 #include <math.h>
+#include <stdbool.h>
 
 #include "hstcal_memory.h"
 #include "hstcal.h"
@@ -30,7 +31,7 @@ static int insertAmp(SingleGroup * amp, const SingleGroup * image, const unsigne
 static int alignAmpData(FloatTwoDArray * amp, const unsigned ampID);
 static int alignAmp(SingleGroup * amp, const unsigned ampID);
 
-int doPCTEGen2 (ACSInfo *acs, CTEParamsFast * ctePars, SingleGroup * chipImage)
+int doPCTEGen2 (ACSInfo *acs, CTEParamsFast * ctePars, SingleGroup * chipImage, const bool forwardModelOnly)
 {
 
     /* arguments:
@@ -158,13 +159,20 @@ int doPCTEGen2 (ACSInfo *acs, CTEParamsFast * ctePars, SingleGroup * chipImage)
             return (status = ALLOCATION_PROBLEM);
         }
 
-        //CALCULATE THE SMOOTH READNOISE IMAGE
-        trlmessage("(pctecorr) Calculating smooth readnoise image...");
-        if (ctePars->noise_mit != 0)
+        if (forwardModelOnly)
         {
-            trlerror("(pctecorr) Only noise model 0 implemented!");
-            freeOnExit(&ptrReg);
-            return (status = ERROR_RETURN);
+            // Not needed as we won't be smoothing.
+        }
+        else
+        {
+            //CALCULATE THE SMOOTH READNOISE IMAGE
+            trlmessage("(pctecorr) Calculating smooth readnoise image...");
+            if (ctePars->noise_mit != 0)
+            {
+                trlerror("(pctecorr) Only noise model 0 implemented!");
+                freeOnExit(&ptrReg);
+                return (status = ERROR_RETURN);
+            }
         }
         SingleGroup smoothedImage;
         initSingleGroup(&smoothedImage);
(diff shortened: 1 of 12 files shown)


jamienoss:
@mdlpstsci & @pllim This is still a little hacky and there's duplicated code that _could_ be factored, but anyhow here it is.

jamienoss:
I decided to revert back to requiring ``PCTECORR=PERFORM`` on the basis that if no action flags are set to ``PERFORM``, hstcal complains (at some early and generic stage) and exits. So if all ``XXXCORR`` equal either ``OMIT`` or, more relevant ``COMPLETE``, something needs to be set to ``PERFORM`` and this _should_ be ``PCTECORR``.

jamienoss:
An error may need to be added for when using ``--forwardModelOnly`` and the gen 1 algorithm. If not, gen1 use will ignore the flag do the actual CTE without the user knowing. Actually, I've just thought about where this could be added, so will quickly do now.

pkg/acs/calacs/calacs/acsmain.c
```diff
@@ -25,7 +26,7 @@ char MsgText[MSG_BUFF_LENGTH]; // Global char auto initialized to '\0'
 
 static void printSyntax(void)
 {
-    printf("syntax:  calacs.e [--help] [-t] [-s] [-v] [-q] [-r] [--version] [--gitinfo] [-1|--nthreads <N>] [--ctegen <1|2>] [--pctetab <path>] input \n");
+    printf("syntax:  calacs.e [--help] [-t] [-s] [-v] [-q] [-r] [--version] [--gitinfo] [-1|--nthreads <N>] [--ctegen <1|2>] [--pctetab <path>] [--forwardModelOnly] input \n");
```
pllim: What is the use case of running the entire pipeline on RAW using forward model? I don't think this needs to be an option for `calacs.e`?

jamienoss: Good point, I'll remove.

pllim:
> there's duplicated code that could be factored

Would be nice to address the code duplication sooner than later, IMHO.

jamienoss:
> I know how you and Michele feels about those Python wrappers but to be consistent, this might also need a wrapper over at ACSTOOLS.

I'll wrap away.

> I was thinking more like a new, say, acsctefowardmodel.e command, as not to confuse users too much. acscte.e turns blv_tmp into blc_tmp

I'll create a wrapper for this too.

>turns blv_tmp into blc_tmp.

This still creates a blc_tmp. which would need to be addressed before merge. Not sure what the team will want.

mdlpstsci:
I will be able to do a full review next week.

To clarify: I am not against Python wrappers per se, I am against duplication of code.

I would agree with @pllim that this is not really a calacs function.  It should be a standalone routine as it is for modelling purposes.  I can hopefully say more intelligent things once I review the code.

I would say this overall project was the biggest success story of the Hack Day! Kudos.

jamienoss:
@pllim 
>>I was thinking more like a new, say, acsctefowardmodel.e command, as not to confuse users too much. acscte.e turns blv_tmp into blc_tmp
>
>I'll create a wrapper for this too.

Actually, nope, maybe I'll let someone else do that.

jamienoss:
>>there's duplicated code that could be factored
>
>Would be nice to address the code duplication sooner than later, IMHO.

Ok, I lied about the _could_. The code isn't easy to dedup (if at all) and prob. isn't worth it or even trying. The main argument is the diff in memory allocations.

pkg/acs/calacs/acscte/dopcte-gen2.c
```diff
@@ -201,13 +221,26 @@ int doPCTEGen2 (ACSInfo *acs, CTEParamsFast * ctePars, SingleGroup * chipImage)
...
-       trlmessage("(pctecorr) Running correction algorithm...");
-       //perform CTE correction
        SingleGroup * cteCorrectedImage = &columnMajorImage;
-       if ((status = inverseCTEBlur(&smoothedImage, cteCorrectedImage, &trapPixelMap, ctePars)))
+       if (forwardModelOnly)
        {
-           freeOnExit(&ptrReg);
-           return status;
+           trlmessage("(pctecorr) Running forward model simulation...");
+           //perform CTE correction
```
mdlpstsci: Minor issue: Comment should be updated.

jamienoss:
I had forgotten to ``git add`` the new ``wscript``. @mdlpstsci I wouldn't bother re-reviewing it till we hear from the team about the file name.

pkg/acs/calacs/acscte/forward_model/wscript
```diff
@@ -0,0 +1,25 @@
...
+                  '../dopcte.c',
+                  '../dopcte-gen2.c',
+                  '../getcteflag.c',
+                  '../getctesw.c',
+                  '../pcte_fixycte.c',
+                  '../pcte_funcs.c',
+                  '../maincte.c',
+                  'main.c'
+                  ],
+        target = 'acscteforwardmodel.e',
```
jamienoss: @mdlpstsci This is where the new exec is named.

jamienoss:
The ACS team decided on ``_ctefmod.fits``.

pkg/acs/calacs/calacs/calacs.c
```diff
@@ -611,7 +612,7 @@ int ProcessACSCCD (AsnInfo *asn, ACSInfo *acshdr, int *save_tmp, int printtime,
                 if (acshdr->sci_basic_cte == PERFORM) {
                     if (ACScte(acshdr->blv_tmp, acshdr->blc_tmp,
                                &acscte_sci_sw, &sciref, printtime,
-                               asn->verbose, nThreads, cteAlgorithmGen, pcteTabNameFromCmd)) { //this is the line
+                               asn->verbose, nThreads, cteAlgorithmGen, pcteTabNameFromCmd, false)) { //this is the line
```
mdlpstsci: I would like to see an explicit comment here indicating the forwardModelOnly is not a feature that was designed to be used during pipeline processing (hence the false).  Rather this option is for standalone processing when running the CTE component of the pipeline for calibration purposes.

jamienoss:
@mdlpstsci back to you for final review.

Issue #314, opened 2018-06-11
Modify dark correction algorithm to work in 2D


mdlpstsci:
The dark correction algorithm needs to be modified such that it works on the data in a two-dimensional sense rather than on a line-by-line basis.  This is to accommodate application of a new dark correction algorithm (TBD by ACS Team) and high-dynamic range reference file which needs to address the science data in a two-dimensional way.

This work was started during the ACS Hack Day.

mdlpstsci:
Implemented and cleaned up Hack Day code to support the dark correction being applied on the full two-dimensional image rather than corrections being done on a line-by-line basis. Additional modifications or new ancillary routines were necessary to correct processing subarray data.  Tested on WFC data (full frame, subarray data, polarization, and ramp), as well as data from the other detectors, HRC and SBC. The manual regression testing on my workstation was successful on this small sampling (18 images) of data.

Issue #88, opened 2017-04-14
Allow use of C99 standard


jamienoss:
For those older compilers that do not already default to C99 but are capable of, we should allow the use of the C99 standard for our build. We should not be supporting those that are too old to do so, in fact, there are infinitely greater reasons for doing this other than just language standards.

jamienoss:
@stsci-hack @sosey @pllim @jhunkeler

jamienoss:
Spawned from failing Travis builds in #82

pllim:
While I have no objection to this, I am just going to slowly back away and walk out the door...

jamienoss:
https://gcc.gnu.org/onlinedocs/gcc/Standards.html
https://gcc.gnu.org/c99status.html

jamienoss:
What are our min c compiler version requirements, I couldn't see anything in the readmes etc? 

BTW we should state our support levels/requirements within the README/INSTILLATION including that which ops uses for the actual pipeline build.

jhunkeler:
Give or take a minor/patch revision:
- GCC 4.1.2
- GLIBC 2.5

jamienoss:
It's not quite 4.5 but close enough perhaps...

What do we build the pipeline with again? I would have thought that it only made sense to make that version the min version requirement, for validation purposes?

jhunkeler:
Linux HSTDP pipeline builds are compiled on RHEL5 (using GCC 4.1.2), however the build coming up in a ~week will be built using RHEL6 (GCC 4.4.7). DMS just got an OS upgrade, so our requirement for RHEL5 binaries has gone away.

PS - Generic HSTCAL in AstroConda is already being built under RHEL6.  And yes, it's the build that gets tested daily with pandokia. ;)

jamienoss:
@stsci-hack what are your thoughts?

stsci-hack:
We should not mandate the language used for this code.  However, I do support finding ways to allow the use of C99 language in the code where it is appropriate.  This will allow us to identify how to support newer language standards throughout our builds going forward.

sosey:
[...]

PR #315, opened 2018-06-13
Modify the dark correction to work directly on the 2D image rather than line-by-line

mdlpstsci:
Addresses IT #314.

The dark correction algorithm has been modified to work directly on the full 2D data rather than on a line-by-line basis.  This is in preparation for the implementation of a new dark correction algorithm which accommodates a high-dynamic range dark.  The new algorithm is TBD by the ACS team.  At this stage, the known 2D modifications have been done and regression testing with a varied sample of 18 ACS datasets.  Please note there are still some print statements in the code being utilized for testing.  Also, there is an open question as to the exact computation wanted for the mean dark value.  This code only exists on my origin/hackday branch at this time.

### DIFF
--- a/pkg/acs/calacs/acs2d/dodark.c
+++ b/pkg/acs/calacs/acs2d/dodark.c
@@ -8,6 +8,7 @@
 # include <stdlib.h>		/* calloc */
 # include <math.h>		/* fabs */
 
+#include "hstcal_memory.h"
 #include "hstcal.h"
 # include "hstio.h"
 # include "acs.h"
@@ -28,10 +29,9 @@
  assume ratio of bin factors to be 1.
  
  The value of MEANDARK is calculated based on the weighted average
- of each lines' dark value.  The weighting is based on the percent of 
- good pixels in each line, so only pixels not flagged BAD (in some way)
- will contribute to the average, and each line will contribute only 
- as much as the line has good pixels. 
+ of the image where the weighting is based on the percent of 
+ good pixels in the image.  Only pixels not flagged BAD (in some way)
+ will contribute to the average.
  
  Warren Hack, 1998 June 11:
  Initial ACS Version.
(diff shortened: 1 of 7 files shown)


mdlpstsci:
This branch will be merged **after** Nate has had the opportunity to evaluate the changes.

pkg/acs/calacs/acs2d/dodark.c
```diff
@@ -53,152 +52,166 @@
...
-    
-    multk1d(&z, darktime);
-    
-    AvgSciValLine (&z, acs2d->sdqflags, &dark, &wdark);
+     update = NO = do not update the header of the scratch image
+  */
+  update = NO;
+  if (trim2d (&y, x0, y0, rx, ry, update, &z)) {
+     trlerror ("(darkcorr) size mismatch.");
+     freeSingleGroup (&y);
```
jamienoss: For future ref, I would have taken the opportunity to use the ``PtrRegister`` so as to remove the burden of knowing what is alive to be freed upon return.

Struct is in [include/hstcal_memory.h](https://github.com/spacetelescope/hstcal/blob/master/include/hstcal_memory.h) and the "methods" in [lib/hstcal_memory.c](https://github.com/spacetelescope/hstcal/blob/master/lib/hstcal_memory.c). [Here's an example](https://github.com/spacetelescope/hstcal/blob/master/include/hstcal_memory.h#L4-L32) - for ``SingleGroup``s you pass ``&freeSingleGroup`` as the free func.

pkg/acs/calacs/lib/multk2d.c
```diff
@@ -13,25 +18,57 @@ float k          i: multiply a by this constant
...
 	/* science data */
-    dimx = a->sci.data.nx;
-    dimy = a->sci.data.ny;
+    int dimx = a->sci.data.nx;
+    int dimy = a->sci.data.ny;
     
-	for (j = 0;  j < dimy;  j++)
-	    for (i = 0;  i < dimx;  i++)
-		Pix (a->sci.data, i, j) = k * Pix (a->sci.data, i, j);
+    int i, j;
```
jamienoss: I would have scoped ``i`` & ``j`` and these should be ``unsigned``. Really I would do:
```c
for (unsigned int j = 0; j < dimy; ++j)
{
...
```
However, this is C99 which we have not yet got to (_c.f._ #88) so
```c
{unsigned j;
for (j = 0; ...}
{
...
}}
```
 Also, just small poke, ``++i`` is more performant than ``i++``, even though pretty much all compilers will do the conversion for you.

jamienoss: Also, it looks like you have an auto formatter set up, as there seem to be unnecessary indent changes that will pollute the blame history.

pkg/acs/calacs/lib/multk2d.c
```diff
@@ -13,25 +18,57 @@ float k          i: multiply a by this constant
...
+    for (j = 0;  j < dimy;  j++) {
```
jamienoss: These could have been trivially threaded with OpenMP.

pkg/acs/calacs/lib/sub2d.c
```diff
@@ -0,0 +1,57 @@
...
+
+   The science data arrays are subtracted; the error arrays are combined;
+   the data quality arrays are combined.
+
+   M.D. De La Pena: 05 June 2018
+   Created this file/routine from sub1d.c.
+   the data quality arrays are combined.  For all extensions, the entire 2D
+   arrays are used.
+*/
+int sub2d (SingleGroup *a, SingleGroup *b) {
```
jamienoss: I would have made a generic 2d vector func that had the op func passed into it, in this instance the op is the sub.

jamienoss: Multi flavours or wrappers may be wanted/needed to deal with the DQ masks.

pkg/acs/calacs/lib/trim2d.c
```diff
@@ -0,0 +1,99 @@
...
+		    DQSetPix (b->dq.data, m, n, DQPix(a->dq.data, i, j));
+        }
+    }
+
+	/* Copy header info only if requested... */
+	if (update == YES) {
+		/* Copy the headers. */
+		copyHdr (b->globalhdr, a->globalhdr);
+		if (hstio_err())
+	    	return (status = 1001);
```
jamienoss: Missing indentation, unguarded IF warnings may have been produced for these.

jamienoss: Never use the naked numbers, use the enum ids in [include/hstcalerr.h](https://github.com/spacetelescope/hstcal/blob/master/include/hstcalerr.h#L40). In this instance it's ``INVALID_EXPTIME`` which I don't think makes sense here. Also, I think you should be returning the status (or it's mapping) actually tested within ``hstio_err()``.

jamienoss: These are missing from a lot of your IFs, check whether you have tabs or ws for indents.

pkg/acs/calacs/acs2d/dodark.c
```diff
@@ -53,152 +52,166 @@
...
-	mean = 0.0;
-  weight = 0.0;
-  
-	/* Bin the dark image down to the actual size of x. */
   
-	initSingleGroupLine (&z);
-	allocSingleGroupLine (&z, x->sci.data.nx);
-	for (i=0, j=y0; i < scilines; i++,j++) { 
+  /* Trim the dark image (y->z) down to the actual size of the science image (x). */
+  initSingleGroup (&z);
```
jamienoss: I would move the dec for ``z`` to here or the ``initSingleGroup`` to just after the dec. Pref the former, either way, the uninitialized gap needs to be closed.

pkg/acs/calacs/lib/trim2d.c
jamienoss: @mdlpstsci  This looks like a regression/bug to me.

pkg/acs/calacs/lib/multk2d.c
mdlpstsci: Made unsigned ints.  Wanted the C99 version - boo hoo.
No autoformatter, but code whacked out previously.  Regrettable regarding the spaces for blame, but "a woman's gotta do what a woman's gotta do".  At least use "?w=1" at end of URL or "diff settings" to view nicely.

pkg/acs/calacs/lib/sub2d.c
mdlpstsci: No go.  This issue balloons into a lot of variations.

pkg/acs/calacs/lib/trim2d.c
mdlpstsci: Updated.  Cruft resident in historic code.

mdlpstsci: Updated.  Cruft resident in historic code.  FYI - copyHdr()  can return raw status, and depending upon the error, hstio_err() may or may not be set.  Grrr.  I now check a return and hstio_err() and will ultimately return the more informative error.

pkg/acs/calacs/acs2d/dodark.c
mdlpstsci: Updated for y and z.

pkg/acs/calacs/acs2d/dodark.c
mdlpstsci: Excellent utility and implementing in routines I have modified.

pkg/acs/calacs/lib/findbin.c
```diff
@@ -295,3 +300,105 @@ int *x0, *y0      o: location of start of subimage in ref image
...
+        int *rx, int *ry, int *x0, int *y0) {
+
+    /* arguments:
+       SingleGroup *x    i: science image
+       SingleGroup *y    i: reference image
+       int *same_size    o: true if zero offset and same size and binning
+       int *rx, *ry      o: ratio of bin sizes
+       int *x0, *y0      o: location of start of subimage in ref image
+    */
+
```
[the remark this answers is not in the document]
mdlpstsci: If there are no error conditions, the variables are all set by this function.  If there is error, the program exits.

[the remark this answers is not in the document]
jamienoss: I don't understand. One of the first things this func does is access a passed ptr.
```c
/* Get bin sizes of science and reference images from headers. */
    if (GetCorner (&x->sci.hdr, rsize, sci_bin, sci_corner))
        return (status);
```
i.e. the ``&x->sci.hdr``. There currently exists no guard against this access.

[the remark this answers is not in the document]
mdlpstsci: This code is full of this type of issue.  Will do as an ensemble fix in different commit.

mdlpstsci:
Email from Nate regarding this IT/PR: Hey @mdlpstsci, everything here looks great. The only changes that I find are in the value of the computed mean dark current that is reported in the header, but the changes are at the level of a part in million (1e-6) and that is expected. Feel free to proceed from the process of including these changes into 2018.3

PR #322, opened 2018-07-16
The Breakup [POC]

jhunkeler:
Replaces WAF with CMAKE.

* Source files are explicitly tracked per project
* HSTCAL is now 99% circular dependency-free (i.e.`ctegen2` still requires symbols from each instrument package, but some instruments require `ctegen2` so I couldn't fix it)
* Patched up a few inter-library linkage issues (i.e. `calstis0` requires symbols from `cs2` but those symbols were not defined/exported correctly)
* Dropped nested directories for each instrument
* Refactored instrument directories to use a `lib` and `src` design.
  * `PKGNAME/lib` contains all of the source code for the instrument
  * `PKGNAME/src` contains the main program(s)
* Rolled a few libraries into `hstcalib` because it saves time and decreases the burden of tracking common symbols different sub-projects (i.e. `imphtab` and `ncarfft`)

## How do I play with this?

1. Install CMake (with or without conda)
2. Install GCC and GFortran (with or without conda)
3. Install cfitsio (with or without conda)
4. Install pkgconfig (with or without conda)
5. Add remote target to existing clone
```
git remote add jhunkeler https://github.com/jhunkeler/hstcal.git
git fetch jhunkeler
```

6. Checkout the branch
```
git checkout thebreakup
```

7. Build
```
mkdir _build
cd _build
cmake ..
make
```

To set an installation prefix, do:
```
cmake -DCMAKE_INSTALL_PREFIX=/some/path ..
make
make install
```

This supports `DESTDIR` as well for testing of installations:
```
cmake ..
make
make install DESTDIR=/some/path
```


jhunkeler:
Superseded by #412

PR #330, opened 2018-08-03
Use findTotalNumberOfImsets() to compute ACSInfo::nimsets

### LINKED ISSUE #323: ACS CTE forward model fails when input image has >6 extensions
The most likely science use case for the ACS CTE forward model is for users to add artificial stars to an FLC/FLT image, and often these contain >6 extensions (15 in the case of full-frame data). The extra extensions contain distortion-related information, and shouldn't affect the forward model. 

When testing this functionality, I get the following error (jd9w05toq_sm2.fits is a renamed FLT):

> WARNING: running CTE forward model onlygit tag: 2.0.0-14-g31c96c4
> git branch: master
> HEAD @: 31c96c4af35d5c12aa097199fb3455217447901d
> Setting max threads to 32 out of 32 available
> Trying to open /grp/hst/acs3/ryon/ctecorr/forward_mod/jd9w05toq_sm2.fits...
> Read in Primary header from /grp/hst/acs3/ryon/ctecorr/forward_mod/jd9w05toq_sm2.fits...
> 
> 
> CALBEG*** ACSCTE -- Version 10.0.0 (19-Jan-2018) ***
> Begin    19-Jul-2018 13:13:46 EDT
> Input    /grp/hst/acs3/ryon/ctecorr/forward_mod/jd9w05toq_sm2.fits
> Output   /grp/hst/acs3/ryon/ctecorr/forward_mod/jd9w05toq_sm2_ctefmod.fits
> Trying to open /grp/hst/acs3/ryon/ctecorr/forward_mod/jd9w05toq_sm2.fits...
> Read in Primary header from /grp/hst/acs3/ryon/ctecorr/forward_mod/jd9w05toq_sm2.fits...
> APERTURE WFC
> FILTER1 F606W
> FILTER2 CLEAR2L
> DETECTOR WFC
> DESCRIP read in as: 
> DESCRIP read in as: 
> DESCRIP read in as: 
> ERROR:    Error processing /grp/hst/acs3/ryon/ctecorr/forward_mod/jd9w05toq_sm2.fits.
> ERROR:             HSTIO error 114:  Filename /grp/hst/acs3/ryon/ctecorr/forward_mod/jd9w05toq_sm2.fits EXTNAME SCI EXTVER 3 CFITSIO status 301
> ffopen could not move to the specified extension: extension with EXTNAME = SCI, and with EXTVERS = 3, doesn't exist or couldn't be opened.
> Error opening image array.
> *** stack smashing detected ***: /grp/hst/acs3/ryon/ctecorr/hstcal/bindev/acscteforwardmodel.e terminated

@jamienoss determined that when such a file is opened by CALACS, it finds that there are 5 imsets, and fails when it reaches the 3rd because no SCI extension is found (EXTNAME=SCI and EXTVER=3 is never true for ACS data). He also determined that the stack smashing warning comes from the cfitsio error being longer than the buffer allowed for it.

Potentially relevant parts of CALACS:
[https://github.com/spacetelescope/hstcal/blob/2.0.0/pkg/acs/calacs/include/acs.h#L33-L34](https://github.com/spacetelescope/hstcal/blob/2.0.0/pkg/acs/calacs/include/acs.h#L33-L34)
[https://github.com/spacetelescope/hstcal/blob/2.0.0/pkg/acs/calacs/lib/getacskeys.c#L94-L95](https://github.com/spacetelescope/hstcal/blob/2.0.0/pkg/acs/calacs/lib/getacskeys.c#L94-L95)

jamienoss:
Fixes #323

### DIFF
--- a/hstio/hstio.c
+++ b/hstio/hstio.c
@@ -208,6 +208,11 @@ int getNumHDUs(const char * fileName, int * hduNum)
 
 int findTotalNumberOfImsets(const char * fileName, const char * setContainsExtName, int * total)
 {
+    return findTotalNumberOfHDUSets(fileName, setContainsExtName, IMAGE_HDU, total);
+}
+
+int findTotalNumberOfHDUSets(const char * fileName, const char * setContainsExtName, const int hduType, int * total)
+{
     *total = 0;
     assert(fileName && *fileName!='\0');
 
@@ -240,9 +245,9 @@ int findTotalNumberOfImsets(const char * fileName, const char * setContainsExtNa
     {unsigned i;
     for (i = 1; i <= hduNum; ++i) // HDUs are 1 based
     {
-        int loopStatus = HSTCAL_OK; // dec here to auto reset
-        int hduType = ANY_HDU;
-        if (fits_movabs_hdu(fptr, i, &hduType, &loopStatus))
+        int loopStatus = HSTCAL_OK; // decl here to auto reset
+        int extHDUType = ANY_HDU; // This is populated by fits_movabs_hdu() but init anyhow
+        if (fits_movabs_hdu(fptr, i, &extHDUType, &loopStatus))
         {
             // Since we already know the total number of HDUs, if we can't
             // move through all of them, a real IO error has occurred.
@@ -250,6 +255,10 @@ int findTotalNumberOfImsets(const char * fileName, const char * setContainsExtNa
             fits_close_file(fptr, &closeStatus);
             return loopStatus;
         }
+
+        // Check for HDU type correctness
+        if (hduType != ANY_HDU && extHDUType != hduType)
+            continue;
 
         // Get keyword value
         char keyValue[FLEN_VALUE];
--- a/hstio/hstio.c
+++ b/hstio/hstio.c
@@ -111,6 +111,7 @@
 # include <unistd.h>
 # include <assert.h>
 # include <stdlib.h>
+# include <stdbool.h>
 
 # include "hstio.h"
 # include "hstcalerr.h"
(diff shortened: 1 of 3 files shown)


jamienoss:
@mdlpstsci I still need to do a little more testing, maybe @jryon you can take it for a spin also?

FYI this fix is across all of ``ACS`` not just for ``acscteforwardmodel.e``.  I believe this is how it should be but I might be wrong.

jamienoss:
This implements two methods for computing the number of ``imsets`` in a file.
```python
nimsets = [hdu.name for hdu in hduList].count(setContainsExtName)
```
or
```python
nimsets = len(set([hdu.ver for hdu in hduList]))
```

jryon:
I ran a quick test of the forward model on an FLC file from MAST with 15 extensions. The forward model does not crash and updates the SCI extensions as expected. However, the non-imset extensions were not present in the output FLC_CTEFMOD file. Was this intended? It would be great if those extensions could be retained, if it wouldn't require too much extra effort.

jamienoss:
@jryon 
>However, the non-imset extensions were not present in the output FLC_CTEFMOD file. Was this intended?

Nope, no that was not intended. So close...

jamienoss:
@jryon 

>It would be great if those extensions could be retained, if it wouldn't require too much extra effort.

Should they be retained? I mean, they're not the output from the CTE sim. I think for correctness, there would then have to be some way to indicate that these extra extensions were not forward modelled, right? The user may wonder why they have not changed - something I didn't think about before.

jryon:
The extra extensions are entirely WCS and distortion information. The user wouldn't expect them to be affected by the forward model. Also, only the SCI extension within each imset is affected by the forward model, but the ERR and DQ extensions aren't removed from the output file. Are they retained simply because they're part of the input imset?

jamienoss:
> Are they retained simply because they're part of the input imset?

FYI: none of this can be said to be intentional as it all belongs on the other side of my horizon when implementing it, i.e. I didn't think about it.

Doing so now, however, my previous statement of retention belongs to the imset and not the extension. The ``DQ`` is retained as it is used as is in the CTE code (if at all). The ``ERR`` ext should be altered by the CTE sim if it is not already, right? We should open an issue for this in itself. The CTE correction simply adds 10% of the correction for the error induced, what this should be for the forward model I have no idea.

jamienoss:
BTW, regarding:
>It would be great if those extensions could be retained, if it wouldn't require too much extra effort.

Maybe we should leave this aside for now and open an issue for it after this PR is merged (assuming it is).

@jryon would this be ok or is it critical for these other extensions to persist?

jryon:
It isn't critical for these extensions to be present, simply the most preferred result.

jamienoss:
@mdlpstsci 
>Determining the number of Imsets via EXTVER does give me some cause for concern

This is the implementation of the [ST definition](http://www.stsci.edu/hst/HST_overview/documents/datahandbook/intro_ch23.html#1913215) of an imset:

HDU Keyword | Description | Values
-- | -- | --
EXTVER | Imset number (see Table 2.3) | Integer

>Strictly speaking, the findTotalNumberOfImsets() really finds the total number of extensions named "someString" or extensions which have the same EXTVER.

Are these not imsets? Finding them by ``EXTVER`` _is_ by definition, as stated above. However, since each instrument wants to deal with imsets containing certain extensions, as you stated, I added the ``setContainsExtName`` to account for specificity. Since extensions with the same name _have_ to have unique ``EXTVER`` values (FITS std), and there seems to be no description excluding an imset of size 1, finding imsets by ``EXTNAME`` is also by definition a counting of imsets - just not necessarily all of them but perhaps a subset.

>It seems to me one would get the same answer using EXTNAME=SCI if the file contained only SCI extensions

As is to be expected, right? Sorry, I'm not sure of what your point is here?

>Determining the number of Imsets via EXTVER does give me some cause for concern as this "descriptor" is less bound to a particular type of data.

Perhaps you are referring to this description in the [ST definition](http://www.stsci.edu/hst/HST_overview/documents/datahandbook/intro_ch23.html#1913215):
```
A set of FITS extension images which are logically related to one another is called an imset
```
In which case, I can make the search explicit to those of only ``IMAGE`` type (rather than ``ANY_HDU``). I just wasn't 100% sure if this was literal, is it?

jamienoss:
>It seems to me one would get the same answer using EXTNAME=SCI if the file contained only SCI extensions
>
>>As is to be expected, right? Sorry, I'm not sure of what your point is here?

This is one of the side effects of overloading ``EXTVER`` to define an imset by. Perhaps your point was that if only ``SCI`` extensions existed then technically they're not sets, because they're not associated with anything else, their cardinality is one. 

The issue is that if multiple extensions with identical ``EXTNAME`` exist they have to have unique ``EXTVER`` defined. They are then auto defined as "imsets" by the ST def, which doesn't (and can't because of this point) exclude imsets of size 1. You also can't have them with undefined ``EXTVER``s as they have the same ``EXTNAME``.

mdlpstsci:
I am fine with your fix.  My comments about the EXTNAME and EXTVER I realize are really concerns with the original definition of Imset.   My concern was a general statement regarding the best way IMHO to search for an Imset. When the search is on EXTNAME=SCI, for example, you know what you have in the result - at least for the extension explicitly named SCI.  When the search is on EXTVER=1, you do not actually know the content of the extension.  However, giving the user flexibility on their search via EXTNAME or EXTVER, in conjunction with the assumption HST data is in the form of official HST datasets so the conventions imposed by HST apply, is valid functionality to have.

PR #345, opened 2018-08-27
Adjustments for use on the RT Jenkins instance.

rendinam:
build_mode -> name
Use utility library to manage conda
Force python 3.6


README.md
```diff
@@ -5,6 +5,9 @@
 
 Calibration software for HST/WFC3, HST/ACS, and HST/STIS.
 
+Nightly regression test results are available only from within the STScI network at this time.
```
pllim: The endline in file does not really translate to endline in rendered markdown. FYI. I don't care personally but in case you do.

rendinam:
That failure helpfully drew my attention to the log where I noticed the need to change an invalid property name in the `Jenkinsfile` to allow it to run.

PR #348, opened 2018-08-29
Do not add an error contribution to the ERR array for the CTE forward model

### LINKED ISSUE #347: Do not add extra error to ERR extensions in ACS CTE forward model
According to @jamienoss in [#330](https://github.com/spacetelescope/hstcal/pull/330#issuecomment-410352332),

> The CTE correction simply adds 10% of the correction for the error induced, what this should be for the forward model I have no idea.

Ideally, the ERR extensions should be the same before and after forward modeling, so no extra error should be added by the CTE forward model. This is because the input data into the forward model may vary in terms of how many calibration steps have been performed, and so it may not be accurate to add more noise. We'd like to simply simulate the effects of readout on the SCI extensions, and leave any ERR extension updates to the user.

jamienoss:
Resolves #347

__WARNING!__ I did not test this beyound a build.


jamienoss:
FYI: I don't like having the conditional within the loop but at the same time it's a trivial enough computation not to matter,  and is probably cleaner than recomputing ``delta``.

pllim:
Re: testing -- It is now possible to add a test case to go with the PR.

jamienoss:
Closing due to https://github.com/spacetelescope/hstcal/issues/347#issuecomment-417386658

PR #353, opened 2018-09-13
Enable TEST_BIGDATA to be used across multiple repos

pllim:
As discussed with @stsci-hack . For example, what if a developer wishes to run HSTCAL and CALCOS big data tests back to back without wanting to change `TEST_BIGDATA` in their bash startup script?


pllim:
Hmm... I get this even from internal network. Not sure why.

```python
>>> import requests
>>> url =  'https://bytesalad.stsci.edu/artifactory'
>>> r = requests.head(url, allow_redirects=True)
>>> print(r.status_code, r.reason)
(500, 'Internal Server Error')
```

PR #363, opened 2018-10-05
Code clean up: add explicit scope to crit-sec in WFC3 CTE

### LINKED ISSUE #362: Add explicit scope to OpenMP crit sec in wf3cte.c
[src](https://github.com/spacetelescope/hstcal/blob/master/pkg/wfc3/calwf3/wf3cte/wf3cte.c#L1359-L1364)
```c
#pragma omp critical (cte)
        for (j=0; j< RAZ_ROWS; j++){
            if (Pix(rz.dq.data,i,j)){
                Pix(rc.sci.data,i,j)= pix_modl[j];
            }
        }
```

should be:

```c
#pragma omp critical (cte)
{ // <-----ADD
        for (j=0; j< RAZ_ROWS; j++){
            if (Pix(rz.dq.data,i,j)){
                Pix(rc.sci.data,i,j)= pix_modl[j];
            }
        }
}  // <-----ADD
```

I tried looking through the manual and couldn't find anything concrete that the ``for`` scope would not suffice. Neither in the [4.5 std, sec 2.13.2](https://www.openmp.org/wp-content/uploads/openmp-4.5.pdf) nor the [2.0 std, sec 2.6.2](https://www.openmp.org/wp-content/uploads/cspec20.pdf).

~~I'm not sure if this is or isn't an actual bug~~ (it isn't a bug, but only because the crit-sec isn't even needed - it may be a bug if it were, I don't know). If not, we should still fix for clarity and not having to revisit this question.

@mdlpstsci & @stscicrawford ~~it may be worth testing this as a sanity check for #358.~~ _c.f._ https://github.com/spacetelescope/hstcal/issues/362#issuecomment-427366616

jamienoss:
Resolves #362

Adding explicit scope for crit-sec, however, it isn't actually needed (the crit-sec that is)
since the array access isn't thread overlapping. If it were, the
entire core parallel CTE computation above would not be possible as
it too would need to be within in a crit-sec.

### DIFF
--- a/pkg/wfc3/calwf3/wf3cte/wf3cte.c
+++ b/pkg/wfc3/calwf3/wf3cte/wf3cte.c
@@ -1357,9 +1357,14 @@ int inverse_cte_blur(SingleGroup *rsz, SingleGroup *rsc, SingleGroup *fff, CTEPa
         } /*totflux > 1, catch for subarrays*/
 
 #pragma omp critical (cte)
-        for (j=0; j< RAZ_ROWS; j++){
-            if (Pix(rz.dq.data,i,j)){
-                Pix(rc.sci.data,i,j)= pix_modl[j];
+        {    // Adding explicit scope for crit-sec, however, it isn't actually needed
+             // since the array access isn't thread overlapping. If it were, the
+             // entire core parallel CTE computation above would not be possible as
+             // it too would need to be within in a crit-sec.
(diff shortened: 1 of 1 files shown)


jamienoss:
I would __not_ merge this and instead merge its counterpart,  #364

jamienoss:
Closing as merging #364 instead.

PR #370, opened 2018-10-23
Pin pytest to 3.8.2 to avoid  problem in 3.9.1; normalize package list.

rendinam:
Behavior of `--basetemp` flag in `pytest` appears to have changed in recently published `pytest 3.9.1`, breaking RT tests. Reverting to `3.8.2` works around the issue until the pytest situation is sorted out.

### DIFF
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -24,7 +24,7 @@ bc.conda_channels = ['http://ssb.stsci.edu/astroconda']
 bc.conda_packages = ['python=3.6',
                      'cfitsio',
                      'pkg-config',
-                     'pytest=3.8.2',
+                     'pytest<=3.8.2',
                      'requests',
                      'astropy']
 
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -24,7 +24,7 @@ bc.conda_channels = ['http://ssb.stsci.edu/astroconda']
 bc.conda_packages = ['python=3.6',
                      'cfitsio',
                      'pkg-config',
-                     'pytest',
+                     'pytest=3.8.2',
                      'requests',
                      'astropy']
 
--- a/JenkinsfileRT
+++ b/JenkinsfileRT
@@ -24,7 +24,7 @@ bc.conda_channels = ['http://ssb.stsci.edu/astroconda']
 bc.conda_packages = ['python=3.6',
                      'cfitsio',
                      'pkg-config',
-                     'pytest<=3.8.2',
+                     'pytest=3.8.2',
                      'requests',
                      'astropy']
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,10 +27,15 @@ bc1.name = "release"
 bc1.env_vars = ['PATH=./_install/bin:$PATH',
                 'OMP_NUM_THREADS=8',
                 'TEST_BIGDATA=https://bytesalad.stsci.edu/artifactory']
-bc1.build_cmds = ["conda config --add channels http://ssb.stsci.edu/astroconda",
-                  "conda install -q -y cfitsio pkg-config pytest requests astropy",
-                  "pip install ci-watson",
-                  "${configure_cmd} --release-with-symbols",
+bc.conda_channels = ['http://ssb.stsci.edu/astroconda']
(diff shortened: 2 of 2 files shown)


pllim:
You need to replay that Jenkins first to ensure your fix works.

Also, I would like to see the log of reported failure first before approving. I don't see anything shocking in the pytest doc: https://docs.pytest.org/en/latest/tmpdir.html#the-default-base-temporary-directory

rendinam:
The fix was verified as part of the JWST RT run investigation affected by the same issue.
https://boyle.stsci.edu:8081/job/RT/job/JWST/1/console
The equivalent in `hstcal`'s run:
https://boyle.stsci.edu:8081/job/RT/job/hstcal/60/console
```
[jwst/3.6] self = PosixPath('test_outputs')
[jwst/3.6] 
[jwst/3.6]     def iterdir(self):
[jwst/3.6]         """Iterate over the files in this directory.  Does not yield any
[jwst/3.6]         result for the special paths '.' and '..'.
[jwst/3.6]         """
[jwst/3.6]         if self._closed:
[jwst/3.6]             self._raise_closed()
[jwst/3.6] >       for name in self._accessor.listdir(self):
[jwst/3.6] E       FileNotFoundError: [Errno 2] No such file or directory: 'test_outputs'
[jwst/3.6] 
[jwst/3.6] ../../miniconda-bconf0/envs/tmp_env0/lib/python3.7/pathlib.py:1063: FileNotFoundError
```

Both appared on the same day with the only change being the update of `pytest` package version provided by conda the afternoon before.
```
$ firewatch -t 2d | grep pytest
2018-10-22T13:43:02 : main/osx-64 : pytest-3.9.1-py27_0.tar.bz2             
2018-10-22T13:46:17 : main/osx-64 : pytest-3.9.1-py37_0.tar.bz2             
2018-10-22T13:46:47 : main/osx-64 : pytest-3.9.1-py36_0.tar.bz2
```
Pinning `pytest` to the previously available release version corrects the error as seen in the lastest JWST RT job run. https://boyle.stsci.edu:8081/job/RT/job/JWST/6/console

Some new functionality was added to pytest in the problematic version as indicated here: https://docs.pytest.org/en/latest/tmpdir.html It's plausibly related to the breakage.
Possibly due or related to this: https://github.com/pytest-dev/pytest/issues/4199

rendinam:
After fixing a couple gaffes, the test run proceeds but has unrelated failures.

pllim:
Hmm. Related or not, what does this mean?

```
equests.exceptions.SSLError: HTTPSConnectionPool(host='bytesalad.stsci.edu', port=443): Max retries exceeded with url: /artifactory/scsb-hstcal/dev/wfc3/ir/input/iabg21020_asn.fits (Caused by SSLError(SSLError("bad handshake: SysCallError(-1, 'Unexpected EOF')",),))
```

JenkinsfileRT
```diff
@@ -24,7 +24,7 @@ bc.conda_channels = ['http://ssb.stsci.edu/astroconda']
 bc.conda_packages = ['python=3.6',
                      'cfitsio',
                      'pkg-config',
-                     'pytest',
+                     'pytest=3.8.2',
```
pllim: Can we be more flexible and do `pytest<=3.8.2` here? I usually avoid `=` unless absolutely necessary.

rendinam: > Hmm. Related or not, what does this mean?
> 
> ```
> equests.exceptions.SSLError: HTTPSConnectionPool(host='bytesalad.stsci.edu', port=443):
> Max retries exceeded with url: /artifactory/scsb-hstcal/dev/wfc3/ir/input/iabg21020_asn.fits
> (Caused by SSLError(SSLError("bad handshake: SysCallError(-1, 'Unexpected EOF')",),))
> ```
Unlikley to be related, but I don't know what's going on here other than an obvious problem with communicating to Artifactory.

We know `pytest 3.8.2 works. I cannot necessarily say the same for other versions.
Unless pytest 3.8.2 suddenly goes missing from the conda main channel, the `<` won't have an effect. Ultimately, this is temporary, so it's not critical either way. Adjusted.

rendinam:
Thanks!

PR #395, opened 2018-12-14
Jenkinsfile to use only 4 threads

pllim:
And this, Joe? 

(Yes, I went the lazy route and used GitHub web interface, hence 2 separate PRs.)


pllim:
Superseded by #449

PR #399, opened 2019-02-05
Use released ci-watson from conda, rather than master commit via pip.

rendinam:
This matches the approach taken in `Jenkinsfile`.


jhunkeler:
Makes sense. Good catch. 

This script was likely created before `ci_watson` was in the channel or up on pypi.

pllim:
Yes, HSTCAL was the first adopter of `ci-watson`. Thanks for cleaning it up!

Issue #373, opened 2018-10-30
ACSREJ: Refactoring the treatment of pixels marked in the BPIXTAB


nmiles2718:
After the discovery of [flecks on the ACS/WFC CCD](http://www.stsci.edu/hst/acs/documents/isrs/isr1803.pdf), the ACS Team created a new bad pixel table (BPIXTAB) to flag pixels contaminated by the fleck during processing. In the midst of creating the new BPIXTAB, @samanthalh discovered that the values for the flags in the BPIXTAB were incorrect (they were supposed to have a value of 4, but they were marked with 1024). 

The information stored in the BPIXTAB is propagated to the DQ arrays of the input file in `acsccd` during the doDQI step. This means they are present before the call to ACSREJ. After updating the BPIXTAB to use the correct value of 4, @jryon noticed that the imprint of the flecks began appearing in the SCI and ERR extensions of superbias files that are created by stacking a month of bias frames together. The flecks are caused by particulates on the surface of the CCD and therefore only affect data where light (from external sources, the post-flash LED, or lamps) is incident on the surface of the CCD. Hence, they do not and should not appear in the SCI and ERR extensions of individual bias frames, which contain only signal from the detector and electronics. This suggests that ACSREJ has internal logic to treat pixels flagged with DQ=4 differently in the SCI and ERR arrays, which went unnoticed because the BPIXTAB was previously using the incorrect values.

An example of the effects can be seen in the figure below:
![bpixtab_issue](https://user-images.githubusercontent.com/20194427/47741026-abf09d00-dc4f-11e8-9160-96d35ff7dbc0.png)

The desired behavior would be for ACSREJ to propagate the BPIXTAB values from the individual DQ arrays to the final CRJ product, while ignoring them in the SCI and ERR arrays during the combination process (i.e. treat them as normal pixels).

nmiles2718:
FYI:
For the pixels marked in the BPIXTAB, the values are replaced by the sky value that is computed by `ASCREJ`. If you stack a set of images with `SKYSUB=‘none’`, then all the pixels marked in the BPIXTAB will have a value of 0. If you set it to `SKYSUB=‘mode’`, then they will have whatever value is computed by ACSREJ. Given this, I would guess that `ACSREJ` sniffs the DQ during some initial step and gathers information about pixels marked in the BPIXTAB before it does anything else. Hopefully this helps to narrow things down some, but if not, it at least provides us with an easy way to test the new behavior after updates are made.

[...]

PR #402, opened 2019-02-27
WIP - BPIXTAB: Do not use DQ=4 when combining BIAS or DARK images in ACSREJ.

mdlpstsci:
Addresses IT #373.

The value of IMAGETYP is used to determine whether or not the data to be processed via ACSREJ is a set of BIAS or DARK images.  If BIAS or DARK images are processed, the DQ=4 value (Bad pixel detector) is ignored during the combination of the SCI and ERR extensions, though the DQ=4 value is present in the DQ extension of the final *_crj.fits file.  Results based upon the updates to the CALACS pipeline can be found in /user/mdelapena/acs where the asn, raw, and spt input files were provided by @nmiles2718.  Please see these files for correctness.  Also, my branch (issue373_bpixtab) can be accessed to allow you to process datasets yourself.  If you have any issues, please let me know.

### DIFF
--- a/pkg/acs/calacs/acsrej/acsrej_loop.c
+++ b/pkg/acs/calacs/acsrej/acsrej_loop.c
@@ -3,13 +3,14 @@
 # include   <stdlib.h>
 # include   <math.h>
 
-#include "hstcal.h"
+# include   "hstcal.h"
 # include   "hstio.h"
 
 # include   "acs.h"
 # include   "acsrej.h"
 # include   "rej.h"
 # include   "hstcalerr.h"
+# include   "str_util.h"
 
 /* local mask values */
 # define    OK          (short)0
@@ -137,6 +138,8 @@ Code Outline:
   12-Jan-2016   P.L. Lim    Calculations now entirely in electrons.
   14-Jan-2016   P.L. Lim    Replace threshold formula with ERR. Cleaned up
                             threshold calculation function.
+  26-Feb-2019   M.D. DeLaPena Check the IMAGETYP and adjust the maskdq
+                            for BIAS and DARK images.
 */
 int acsrej_loop (IODescPtr ipsci[], IODescPtr iperr[], IODescPtr ipdq[],
                  char imgname[][CHAR_FNAME_LENGTH], int grp [], int nimgs,
@@ -144,7 +147,7 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr iperr[], IODescPtr ipdq[],
                  float sigma[], multiamp noise, multiamp gain,
                  float efac[], float skyval[], FloatTwoDArray *ave,
                  FloatTwoDArray *avevar, float *efacsum,
-                 ShortTwoDArray *dq, int *nrej, char *shadfile) {
+                 ShortTwoDArray *dq, int *nrej, char *shadfile, char *imagetyp) {
     /*
       Parameters:
 
@@ -186,6 +189,7 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr iperr[], IODescPtr ipdq[],
                   CR flags in the last iteration. This is used to
                   calculate REJ_RATE keyword value in CRJ/CRC output image.
       shadfile i: SHADFILE filename. This is used to apply SHADCORR.
+      imagetyp i: Image type or exposure identifier (e.g., bias, dark).
     */
 
     extern int status;
@@ -199,7 +203,7 @@ int acsrej_loop (IODescPtr ipsci[], IODescPtr iperr[], IODescPtr ipdq[],
     float   sig2, psig2, rej2;  /* square of something */
     float   *exp2;              /* square of exptime per image*/
     float   scale, val, dum;
-    short   sval, crflag, nocr, nospill, dqpat;
+    short   sval, crflag, nocr, nospill, dqpat, nobadpix;
     short   maskdq;
 
     float   efacn, skyvaln;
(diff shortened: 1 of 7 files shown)


nmiles2718:
@jryon and I are currently in the process of reviewing this PR. For completeness, I am listing the following cases that we are testing below and their expected outcome:
1) For BIAS and DARK frames 
    - Ignore the presence of BPIXTAB values in the individual imaged DQ arrays and simply combine the images as if all the pixels are good. We noticed that the BPIXTAB information is still propagated into the DQ array of the final CRJ product, but for the reference files this is fine. After combining the images, the DQ array of the CRJ is wiped clean and new flags are introduced specific to the bias and dark artifacts we flag (eg bias structure and hot/warm pixels).
2) For external there are two scenarios:
     1)  If the pixel is marked bad in *all* images, replace it with the computed sky value and
          propagate the BPIXTAB DQ information from the individual frames into the final CRJ product. 
          This will help to notify the users that these pixels were marked as bad in all images and that their value may not be representative of the true value of the source when observed with a good pixel.
    2)  If the pixel is marked as bad in some, but not all images then use the pixels where it is not
           marked as bad to compute the final CRJ pixel value and do not propagate the BPIXTAB DQ information into the final CRJ product.

mdlpstsci:
I certainly will make the change requested.  Jenna elaborated on "Test Case 2" in the Issue, but I will note for you this is also a change to the current ACSREJ functionality (i.e., the DQ=4 flags are retained in "external" datasets currently).  Stay tuned!

mdlpstsci:
@nmiles2718 @jryon Do you have a favorite test dataset for Case 2.ii above?  I have the jd4e01040_asn dataset.

mdlpstsci:
@nmiles2718 @jryon Please find the output CRJ for bias and dark data in /home/mdelapena/ACS/BPIXTAB/[bias|dark].  I am still fiddling with the source for EXT data.  The code (WIP) has been updated on my branch `mdlpstsci:issue373_bpixtab`.

tddesjardins:
From offline discussions, the ACS team has decided to move the flecks in BPIXTAB, which triggered this request for an ACSREJ update, to the DQ arrays of the flat files using a DQ flag of 512 (bad pixel in reference file) rather than 4 (bad pixel). Since no external data bridging the appearance of the flecks will be combined with ACSREJ in the archive, and such data should not be forcibly combined, only internal observations may be affected. Thus, only test case 1 needs to be implemented at this time and test case 2 can be dropped.

mdlpstsci:
@nmiles2718 @jryon The results of the updates to `calacs.e` for BIAS and DARK image can be found in  the subdirectories of `/user/mdelapena/ACS29May/`. These files are essentially the same as the ones you have examined ~28 February 2019, except I value of "4" has been removed from the DQ arrays.  I am not sure this is what you still want, so please let me know.  This is an easy and quick fix (really).   SCSB also has an new policy regarding testing of code.  If these results seem OK, then the code will be reviewed and pushed to master.  You can then retrieve the code from master.

jryon:
@mdlpstsci  I reviewed the BIAS and DARK images you provided, and they look as expected. Thanks!

mdlpstsci:
@nmiles2718 @jryon The results of the updates to `calacs.e` for BIAS and DARK image can be found in  the subdirectories of `/user/mdelapena/ACS29May/`. These files are essentially the same as the ones you have examined ~28 February 2019, except I value of "4" has been removed from the DQ arrays.  I am not sure this is what you still want, so please let me know.  This is an easy and quick fix (really).   SCSB also has an new policy regarding testing of code.  If these results seem OK, then the code will be reviewed and pushed to master.  You can then retrieve the code from master.

mdlpstsci:
@nmiles2718 @jryon 
I have been thinking about this issue (perhaps overthinking this issue), and I want to verify something with you.  I have been studying `calacs` in more detail.  If the BPIXTAB value really does not apply to bias and dark images, the value currently will not be applied to the DQ array by setting DQICORR to OMIT.  The downside of this is the two saturation values (256 and 2048) which are also set during the DQICORR step will also not be applied.  If you want the saturation values to be set, but you agree that you do not want the BPIXTAB value set for bias and dark images, I can modify the DQICORR step as it applies _only_ to bias and dark images.   This is actually a cleaner way to create the bias or dark CRJ file without BPIXTAB interference.  Having said this, I may not understand the full extend of what you want, so perhaps what I am suggesting is just crazy talk.  I have made a quickie version of files processed only with BIXTAB values not applied to the bias and dark data.  You can see the results in `/user/mdelapena/ACS05June`.  I realize you want to close out this issue, but I thought it was worth investigating this alternate solution.

jryon:
@mdlpstsci 
@nmiles2718 and I discussed this idea, and we agreed that we do not want to change DQICORR. We still want individual darks and biases to contain bad pixel flags. It's only in the case of combining stacks of biases or darks with ACSREJ that we want the bad pixel-flagged pixels to be treated as normal pixels. We actually don't really care what ends up in the DQ arrays of a bias or dark CRJ file, since during reference file creation, the DQ arrays are wiped and repopulated with other flags. Let me know if this doesn't make sense.

mdlpstsci:
@nmiles2718 @jryon Please know I need one of you to review and approve before I can merge the software onto master.

jryon:
I approve of these changes.

pkg/acs/calacs/acsrej/acsrej_loop.c
```diff
@@ -137,14 +138,16 @@ Code Outline:
   12-Jan-2016   P.L. Lim    Calculations now entirely in electrons.
   14-Jan-2016   P.L. Lim    Replace threshold formula with ERR. Cleaned up
                             threshold calculation function.
+  26-Feb-2019   M.D. DeLaPena Check the IMAGETYP - can ignore cosmic ray flag (4)
+                            for BIAS and DARK images.
```
nmiles2718: I would switch this to "can ignore BPIXTAB flag (4) for bias and dark". This could be confusing to future people because since the cosmic ray flag is 8192

mdlpstsci: Updated comments.

mdlpstsci:
@nmiles2718 Sorry, you have to approve again as I pushed new comment changes.

PR #404, opened 2019-04-17
Update RT link

pllim:
Ununwished.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+junit_family = xunit2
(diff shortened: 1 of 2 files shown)


pllim:
Does the `pytest.ini` look right, Matt?

pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
```
pllim: Is this supposed to be `[pytest]` or `[tool:pytest]`?

pllim: Ah, `tool` is only for `setup.cfg`...

Issue #409, opened 2019-09-04
TST: calacs.e not found, relative path broken in Jenkins


pllim:
Due to spacetelescope/jenkins_shared_ci_utils#61

cc @rendinam and @jhunkeler

pllim:
Fixed by #410

PR #411, opened 2019-09-04
TST: Fix gcc not found in Travis CI

pllim:
Example log: https://travis-ci.org/spacetelescope/hstcal/builds/580815630

```
PackagesNotFoundError: The following packages are not available from current channels:

  - gcc

Current channels:
  - http://ssb.stsci.edu/astroconda/osx-64
  - http://ssb.stsci.edu/astroconda/noarch
  - https://repo.anaconda.com/pkgs/main/osx-64
  - https://repo.anaconda.com/pkgs/main/noarch
  - https://repo.anaconda.com/pkgs/r/osx-64
  - https://repo.anaconda.com/pkgs/r/noarch
```

Note for reviewer: This PR only attempts to fix Travis CI. Jenkins failure is out of scope (see #409).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
   global:
     # Set defaults to avoid repeating in most cases
     - ASTROCONDA=http://ssb.stsci.edu/astroconda
-    - CONDA_DEPS="cfitsio curl gcc pkg-config"
+    - CONDA_DEPS="cfitsio curl pkg-config"
     - EXE_PREFIX="/tmp/miniconda3"
     - PYTHON_VERSION=3.6
     - WAF_ARGS_CONFIGURE="-v"
@@ -32,6 +32,7 @@ before_install:
 
 install:
   - conda install --yes -c $ASTROCONDA $CONDA_DEPS
+  - conda install --yes -c conda-forge gcc
   - ./waf configure --prefix=$CONDA_PREFIX $WAF_ARGS_CONFIGURE
   - ./waf build $WAF_ARGS_BUILD
   - ./waf install $WAF_ARGS_INSTALL
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
   global:
     # Set defaults to avoid repeating in most cases
     - ASTROCONDA=http://ssb.stsci.edu/astroconda
-    - CONDA_DEPS="cfitsio curl pkg-config"
+    - CONDA_DEPS="cfitsio curl gcc pkg-config"
     - EXE_PREFIX="/tmp/miniconda3"
     - PYTHON_VERSION=3.6
     - WAF_ARGS_CONFIGURE="-v"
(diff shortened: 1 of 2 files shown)


jamienoss:
Hey, ``gcc`` is available from conda-forge.

pllim:
Yup... trying to figure out how to make the YML add conda-forge.

.travis.yml
```diff
@@ -5,7 +5,7 @@ env:
   global:
     # Set defaults to avoid repeating in most cases
     - ASTROCONDA=http://ssb.stsci.edu/astroconda
-    - CONDA_DEPS="cfitsio curl gcc pkg-config"
+    - CONDA_DEPS="cfitsio curl pkg-config"
```
jamienoss: You could split this out into two, e.g. ``ASTROCONDA_DEPS`` and ``CONDA_FORGE_DEPS``.

.travis.yml
```diff
@@ -32,6 +32,7 @@ before_install:
 
 install:
   - conda install --yes -c $ASTROCONDA $CONDA_DEPS
+  - conda install --yes -c conda-forge gcc
```
jamienoss: And then... exactly like you've done here but ``gcc`` -> ``CONDA_FORGE_DEPS``

jamienoss:
@pllim Perhaps we _should_ just dump it though, it's pretty old (4.8.5) and, if I remember correctly,  astroconda have their own and newer.

pllim:
@jamienoss , what do you mean by "dump it"? You mean, to discontinue OSX tests on Travis CI?

If `astroconda` had `gcc`, it is not finding it now. :woman_shrugging:

jamienoss:
>You mean, to discontinue OSX tests on Travis CI?

No, sorry, I meant that perhaps we shouldn't be explicitly using gcc as a conda dependency but instead whether it is now implicit. Let me try and find the conda doc I'm thinking off and maybe I'm misremembering but what I remember is anaconda trying to shift away from gcc (GNU) and distributing their own.

jamienoss:
Not sure if this is the definitive doc, but I at least remember reading this one. (rereading it now).
https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html

pllim:
Well, I got Travis to be green again at least. As for where to grab the compilers, I would have to defer to @jhunkeler or @rendinam . I would want the CI to mirror as closely to what we ship to end-users as possible.

jamienoss:
> I would want the CI to mirror as closely to what we ship to end-users as possible.

I agree. I'm just not sure if this is actually happening.

Someone might want to take a look at https://github.com/astroconda/astroconda-dev/blob/master/hstcal/meta.yaml#L25

waf_patches/g77.py
```diff
@@ -57,7 +57,7 @@ def get_g77_version(conf, fc):
         conf.fatal('Could not determine the compiler type')
 
     # --- now get more detailed info -- see c_config.get_cc_version
-    cmd = fc + ['-dM', '-E', '-']
+    cmd = fc  # + ['-dM', '-E', '-']
```
pllim: Nope... Need to patch https://gitlab.com/ita1024/waf/blob/master/waflib/Tools/gfortran.py#L61 but I don't want to waste time on this until we have a decision from @jhunkeler.

pllim:
After some offline discussions, this is superseded by #412

Issue #372, opened 2018-10-30
TST: Unpin pytest


pllim:
A reminder for the future to unpin pytest that was pinned in #370. Maybe `pytest` 3.9.3 is peachy again.

pllim:
Fixed by #425

PR #425, opened 2019-09-16
Unpin pytest

rendinam:
`pytest >= 4.0` is required to allow `junit_family` to be specified in the pytest configuration so that test report XML will be produced in way that's compatible with Jenkins. Without specifying `junit_family = xunit2`, test results fail to be imported.

Addresses #372


pllim:
p.s. Ignore the Travis CI failure. @jhunkeler is on that one and it is unrelated to this PR.

rendinam:
Merging as tests failing is an improvement in this case, since they were not being imported at all previously.

Issue #435, opened 2019-09-25
Change CALACS to support time-dependent sensitivity for SBC


rjavila:
I am in the process of implementing the time-dependent sensitivity for the SBC. I’m now testing the IMPHTTAB and I’m finding that the PHOTFLAM values are not being updated correctly. CALACS is not properly updating the PHOTMODE keyword (EXT=1) of FLT images. Currently it looks like this for SBC:

PHOTMODE= 'ACS SBC F115LP'     / obser  

In WFC images the PHOTMODE includes an MJD, and that’s how CALACS knows to apply TDS. It looks like the culprit is the code block at line 103 in the photmode.c file.

mdlpstsci:
The modification is to remove the "if" statements surrounding the update of the PHOTMODE keyword which currently only allow the update to HRC and WFC.  @rjavila do you have a IMPHTTAB file I can use for testing as CALACS fails 

Found parameterized variable 1.
NUMPAR=1, N=1
Allocated 1 parnames
Adding parameter mjd#56392.6597 as parnames[0]
==>ERROR: Matching row not found in IMPHTTAB `acs,sbc,f150lp,mjd#'.
==>ERROR: OBSMODE acs,sbc,f150lp,mjd#
ERROR:    Error return from GetPhotTab.

mdlpstsci:
The hstdp-2019.4 build containing this fix has been generated:
https://github.com/astroconda/astroconda-releases/tree/master/hstdp/2019.4

HSTCAL Release Notes:
https://github.com/spacetelescope/hstcal/releases

mdlpstsci:
Fixed in PR #436.

PR #436, opened 2019-09-25
PHOTMODE keyword updated with MJD for SBC 

mdlpstsci:
CALACS has been modified to allow the PHOTMODE keyword to be appended with the MJD for the SBC, as is currently done for HRC and WFC. This will allow time-dependent sensitivity to be applied properly to SBC.

Testing was done before and after the fix.  Before the fix PHOTMODE would contain the following string in the first extension of the FLT:
jc3003qnq_flt.fits[1],PHOTMODE = "ACS SBC F150LP"

After the fix (testing was accomplished using a test IMPHTTAB provided by @rjavila):
jc3003qnq_flt.fits[1],PHOTMODE = "ACS SBC F150LP MJD#56392.6597"


mdlpstsci:
This PR addresses Issue #435.

Issue #458, opened 2019-12-01
Streamline the extensive WFC3 tests


mdlpstsci:
The WFC3 tests need to be streamlined in that repetitive testing of the same paths (many datasets of the same type) in the calibration code is not necessary and takes a very long time to execute.  Further, there have been instances where tests failed due to a lack of disk space.

mdlpstsci:
Fixed by PR#459.

PR #459, opened 2019-12-01
Streamlined regression tests

mdlpstsci:
These changes address Issue #458.

### DIFF
--- a/tests/wfc3/test_uvis_12single.py
+++ b/tests/wfc3/test_uvis_12single.py
@@ -28,16 +28,14 @@ class TestUVIS12Single(BaseWFC3):
     # Ported from ``calwf3_uv_12``.
     @pytest.mark.parametrize(
         'rootname', ['iaao09l0q'])
-        """
-        'rootname', ['iaao09l0q', 
-                     'iaao09l1q', 
-                     'iaao11odq', 
-                     'iaao11oeq', 
-                     'ibbq01n4q', 
-                     'ibbq01n5q', 
-                     'iblk57bzq', 
-                     'iblk57c0q', 
-                     'iblk57c3q'])
-        """
+        # 'rootname', ['iaao09l0q', 
+        #             'iaao09l1q', 
+        #             'iaao11odq', 
+        #             'iaao11oeq', 
+        #             'ibbq01n4q', 
+        #             'ibbq01n5q', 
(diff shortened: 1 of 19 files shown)


tests/wfc3/test_uvis_12single.py
```diff
@@ -27,6 +27,8 @@ def _single_raw_calib(self, rootname):
 
     # Ported from ``calwf3_uv_12``.
     @pytest.mark.parametrize(
+        'rootname', ['iaao09l0q'])
+        """
```
pllim: I'd rather this is commented out than being turned into a docstring. Same comment applies to all other similar occurrences.

tests/wfc3/test_uvis_15single.py
```diff
@@ -3,7 +3,7 @@
 
 from ..helpers import BaseWFC3
 
-@pytest.mark.slow
+@pytest.mark.skip
```
pllim: Maybe add a reason why this is skipped?

tests/wfc3/test_uvis_12single.py
mdlpstsci: Understood.  I was saving myself some typing.  I actually want to remove many of these datasets, but I would rather do the actual removal when I have time to make sure I am absolutely not deleting something that is useful.

mdlpstsci: Fixed.

tests/wfc3/test_uvis_15single.py
mdlpstsci: Added comment.

tests/wfc3/test_uvis_15single.py
```diff
@@ -3,7 +3,10 @@
 
 from ..helpers import BaseWFC3
 
-@pytest.mark.slow
+# This test is curretnly skipped as it contains 100 (!) files to
```
pllim: I was thinking more like `skip(reason='...')` but this works too. For reference: http://doc.pytest.org/en/latest/skipping.html#skipping-test-functions

```suggestion
# This test is currently skipped as it contains 100 (!) files to 
```

mdlpstsci: Fixed typo and use pytest facilities when appropriate.

Issue #326, opened 2018-08-01
CALACS to use the value of the DARKTIME keyword, moderated by the type of observation


mdlpstsci:
This issue is related to and may well supersede IT #1.  In the DARKCORR step of CALACS, the value of the DARKTIME keyword is used to account for additional time that dark current accumulates.  It has been determined that depending upon the type of observation (i.e., full frame or subarray, post-flash or no post-flash) there is an additional correction needed (ref. ACS ISR 2018-03).  The additional correction values need to be read from the CCDTAB reference file, and the DARKCORR algorithm updated to accommodate new logic for applying the correction.

nmiles2718:
**Long overdue status update**

First things first, it helps to have a working definition of the commanding overheads. The "theoretical" value for the commanding overheads (&delta;t<sub>theo</sub>) is contained in the header keyword, DARKTIME, and it is computed using information stored in the unique data log (UDL) for the observation. For a given observation, the commanding overheads may be computed from the header keywords as follows:
&delta;t<sub>theo</sub> = DARKTIME - (FLASHDUR + EXPTIME)
where,
DARKTIME := total time for dark current to accrue in units of seconds
EXPTIME := exposure time in units of seconds
FLASHDUR := flash duration in units of seconds

To empirically verify the commanding overheads, we analyze stable, hot pixels in dark frames that are taken as part of the CCD Daily Monitor program. For an arbitray hot pixel in one of these dark frames, we can represent the observed pixel value as a linear combination of the contributing sources (dark current, post-flash, and bias). 

p(x,y) = (FLASHDUR+ &delta;t<sub>obs</sub> + EXPTIME ) * dc(x,y) + FLASHDUR * fr(x,y) + bias(x,y)

p(x,y) := pixel value at (x, y) in units of electrons
dc(x, y) := dark current at (x, y) in units of electrons/second
fr(x, y) := flashrate at (x, y) in units of electrons/second
bias(x, y) := bias level at (x, y)
&delta;t<sub>obs</sub> :=  observed commanding overheads
EXPTIME := same as above
FLASHDUR := same as above

Note that since dark current is an inherent property of the detector, it continually accrues the entire time the detector is operating in ACCUM mode which includes the post-flashing. If the DARKTIME keyword accurately encapsulates all the additional offsets associated with commanding, then &delta;t<sub>obs</sub> = &delta;t<sub>theo</sub>.  

[...]

Issue #217, opened 2017-09-12
CALACS: Does CRJ result change when PCTECORR is turn on/off?


pllim:
While working on #71, long story short, I suspect that CRJ result might change depending on whether `PCTECORR=OMIT` or `PCTECORR=PERFORM` with Gen 2 algorithm. Note that CRJ is the CR-SPLIT version of FLT, and hence should not be affected by status of `PCTECORR`. Suggestion to check:

1. Create two directories with the same input files. You can probably take the files from `wfc_asn1_noniraf-gen2cte.py`.
2. In the first directory, run CALACS as usual.
3. In the second directory, first set `PCTECORR=OMIT` in all the input RAW files. Then, run CALACS.
4. Compare `*_crj.fits` from both directories.

If they are the same, then phew, it was just me messing up locally. If they are different, more investigation is needed on why `PCTECORR` would affect a non-CTE corrected file.

I don't have time to look into this, but hopefully @mdlpstsci or @jamienoss can? Thank you!

jamienoss:
@pllim what version of hstcal was this?
I'll run tests now using 1.3.0.

pllim:
It was actually my feature branch in #71, which is equivalent to `dev` as that PR only adds Python tests.

jamienoss:
> It was actually my feature branch in #71, which is equivalent to dev as that PR only adds Python tests.

If it was master, which commit?

pllim:
I just rebased yesterday, so it has to be the latest.

jamienoss:
> I just rebased yesterday, so it has to be the latest.

For future doc, this would be 2b8fe41af30c2d70dca721e7a0afde0ea4665ebc

jamienoss:
Ok, so... (1.3.0)

The temp CRJ files, j6lq01011_crj_tmp.fits, are identicle (other than the header keyword ``PCTECORR``, obviously), howewer, the resultant ASN CRJ files are not, they differ.

The files, j6lq01011_crj.fits differ in data values but also differ in the following header values:

```
Headers contain differences:
     Keyword HISTORY [23] has different values:
        a>   reference image jref$n261315hj_drk.fits
         ?                        ^^^^   ^   -
        b>   reference image jref$w3g15525j_dkc.fits
         ?                        ^ +  ^^^    +
     Keyword HISTORY [25] has different values:
        a>     Superdark created by Doug Van Orsow from proposal 9558
        b>     CTE corrected dark for data taken after Aug 13 2002----------------
[...]

PR #467, opened 2020-03-16
Issue 326 darktime updates

mdlpstsci:
This PR addresses Issue #326.  (FYI - @nmiles2718 @mmcdonald123)

Removed hard-coded darktime scaling value and read new post-flashed and unflashed columns from updated CCDTAB reference files (WFC and HRC) to use for the offset to the DARKTIME FITS keyword value under appropropriate date and supported subarray criteria.  The DARKTIME keyword value is now the default scaling factor for the superdarks (WFC, HRC, and SBC), with the offset being an additive correction to DARKTIME under appropriate circumstances.  The offset only applies to WFC and HRC.

This software update changes data and requires new reference files for WFC and HRC.

A new PyTest was created for regression testing of the new algorithm.

### DIFF
--- a/pkg/acs/calacs/acs2d/dodark.c
+++ b/pkg/acs/calacs/acs2d/dodark.c
@@ -56,6 +56,10 @@
  M.D. De La Pena, 2018 June 05:
  Dark correction processing is now done on the entire image instead of 
  line by line. Also, performed some clean up and new commentary.
+ M.D. De La Pena, 2019 November 26:
+ Remove hard-coded darkscaling value and read new post-flashed and
+ unflashed columns from updated CCDTAB reference file. 
+ 
  */
 
 int doDark (ACSInfo *acs2d, SingleGroup *x, float *meandark) {
@@ -68,7 +72,7 @@ int doDark (ACSInfo *acs2d, SingleGroup *x, float *meandark) {
   
   extern int status;
 
-  const float darkscaling = 3.0;  /* Extra idle time */
+  //const float darkscaling = 3.0;  /* Extra idle time */
   
   int extver = 1;	/* get this imset from dark image */
 
--- a/pkg/acs/calacs/acs2d/dodark.c
+++ b/pkg/acs/calacs/acs2d/dodark.c
@@ -149,8 +149,9 @@ int doDark (ACSInfo *acs2d, SingleGroup *x, float *meandark) {
 
      /* Full-frame data */
      if (acs2d->subarray == NO) {
-         if (acs2d->expstart >= SM4MJD)
+         if (acs2d->expstart >= SM4MJD) {
               darktime = darktimeFromHeader + darktimeOffset;
+         }
          sprintf(MsgText, "Full Frame adjusted Darktime: %f\n", darktime);
          trlmessage(MsgText);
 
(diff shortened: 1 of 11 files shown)


tests/acs/test_wfc_darktime.py
```diff
@@ -0,0 +1,74 @@
...
+    Full-frame, flashed and unflashed, post-SM4
+    Subarrays, flashed and unflashed, pre-Cycle 24
+    Subarrays, flashed and unflashed, post-Cycle 24
+    """
+    detector = 'wfc'
+    ignore_keywords = ['filename', 'date', 'iraf-tlm', 'fitsdate',
+                       'opus_ver', 'cal_ver', 'proctime', 'history',
+                       'bitpix', 'naxis', 'extend', 'simple']
+
+    def _single_raw_cte(self, rootname):
```
pllim: I don't think CTE is necessary to test DARKTIME. And it takes too long.

mdlpstsci: Of course - DOH.

nmiles2718: I know that we've run into issues in the past with the dark subtraction of non-CTE v. CTE corrected data (see e.g. #217). So while this make take a long time to run, I think it's a good idea in the end. Where are these tests executed and are we in danger of timing out on the platform if we leave the CTE correction in?

mdlpstsci: We do need to be mindful of the time it takes to run the as the whole set of HSTCAL tests gets kicked off on the PR.  In the past I have added boundary pushing tests which took a lot of time (will not delete until I have time to evaluate) to the SKIPPED category so they are not run under normal circumstances.  I used these particular datasets to test for this issue so I could compare against Nate's output as a baseline for "correctness".  I just checked the time it took to run these tests to previous Jenkins runs which did not include these tests, and the processing time is just about the  same (?!).  I will leave the tests as is for now , and I can adjust as necessary.

pkg/acs/calacs/acs2d/dodark.c
```diff
@@ -97,19 +116,59 @@ int doDark (ACSInfo *acs2d, SingleGroup *x, float *meandark) {
...
+
+     Effectively the additive factor only applies to ACS/WFC as the HRC was no longer
+     operational by SM4MJD or CYCLE24, and SBC is a MAMA detector.
+  */
+  darktime = darktimeFromHeader;  /* Default */
+  if (acs2d->detector != MAMA_DETECTOR) {
+
+     /* Full-frame data */
+     if (acs2d->subarray == NO) {
+         if (acs2d->expstart >= SM4MJD)
```
stsci-hack: Standard best practice for if statements in C is to use the {} around the conditional code.

tests/acs/test_wfc_darktime.py
pllim: If CTE needs to be tested, at least on a 512 subarray? If this is already a small subarray, then maybe it is okay.

nmiles2718: I like the idea of using a subarray for this, but to test all the possible combinations for the commanding overheads, then full frame data will be required. Also, I think leaving the test as is and adding it to the SKIPPED category is fine.

mdlpstsci:
I made the changes to dodark.c with regard to the curly braces, and the full frame cte darktime tests are now characterized as SKIP.  I am not sure why my  comments  do not show up (resolve before comment?), but you can see the changes in the code.

pllim:
Yes, I can see your replies when I click "show resolved". Thanks!

mdlpstsci:
This PR fixes Issue #326.

PR #470, opened 2020-04-01
BUG: main2d not setting suffix correctly

pllim:
`acs2d.e` on `crj_tmp` input does not produce `crj` as expected. Without this patch, it produces `crj_tmp_flt`. I think I forgot to increase the counter all those years ago back in 510b2184682ede529e9a5.

For comparison, look at the correct counter deeper in the code here:

https://github.com/spacetelescope/hstcal/blob/f5680b59b8a1eeab3d92de708ea4a44c2dfe8550/pkg/acs/calacs/acs2d/acs2d.c#L190

Disclaimer: The code has changed enough since I last tried build it from source that now I am no longer able to build it successfully on my local Linux machine, so I am not sure how to test this patch. 🤷‍♀  (Also see #472)


mdlpstsci:
This certainly looks like the fix.  I will test the change myself this morning.

pllim:
@mdlpstsci , any input with `crj_tmp` or `crc_tmp` should do. But the trick is run `acs2d.e` as standalone, not the entire `calacs.e`.

PR #471, opened 2020-04-02
Fix travis

jhunkeler:


.travis.yml
```diff
@@ -38,4 +46,4 @@ install:
 
 script:
   # Without this dead call, "language: c" runs its own WAF-unaware script
-  - python --version
+  - echo
```
pllim, line 49: How about a call to make sure the executable can actually do something?
```suggestion
  - echo
  - calacs.e --version
```

jhunkeler: This NOP is specific to the `script:` when `language: c` is used... But I can add some runtime tests under `after_install:`

pllim: OK, thanks!

Issue #460, opened 2019-12-05
A-to-D Saturation Flagging in ACS CCD Data


tddesjardins:
The ACS team would like to add to `calacs` a line that checks for pixels with >= 2^16 (65,536) DN during DQICORR and flags these with DQ flags 256 + 2048 (=2304). The ACS data handbook currently describes that this is happening, but it is not (it also says it happens during ATODOCORR, but that step has always been set to OMIT). 

If possible @mdlpstsci could we work this in to coincide with the darktime fix?

tddesjardins:
Actually, speaking with Norman, the limit should be 2^16 - 1 (65,635) DN.

tddesjardins:
Update: we've found a new issue where post-SM4 observations don't seem to follow the 65,535 limit rule. The value appears to be a little lower and it is not a single value, though the range in values is small. We would request that all pre-SM4 (before 2009-01-01) CCD observations use the 65,535 value, while AFTER SM4 pixels with 64,000 are flagged.

mdlpstsci:
@tddesjardins 

1. Can you please provide a dataset to use for testing?
2. I have done a quick check and the code is in place for the ACS CCDs to be flagged with (256 + 2048).  At this time, if the science pixel value is > 65534, then the current dq value is set to the current_value + 2048.  If the science pixel value is > 65534 or the CCD full well saturation level, then the current dq value is set to current_value + 256. The CCD full well saturation level is defined in the CCDTAB.

mdlpstsci:
Discussed with the ACS team that "calibration" values need to be incorporated into the calibration reference files. This is so the information is documented and readily accessible to the Instrument teams for future updates. The CCDTAB currently has a "full well saturation" column, SATURATE.  It is suggested another column be added to the CCDTAB which contains the A-to-D saturation value. Also, there should be a version of this file for pre- and post-SM4 data.  In this way the calibration of the data is reference-file driven and not hard-coded.

Just as FYI - the current code uses an A-to-D threshold of 65534, and the logic is such that values will be flagged with 2048 if the science pixel > 65534.

mdlpstsci:
Sent an email to interested parties on the ACS team for information on this ticket.  Response was:
From: Tyler Desjardins <desjard@stsci.edu>

> Date: Friday, March 27, 2020 at 1:50 PM
> To: Michele De La Pena <mdelapena@stsci.edu>
[...]

PR #475, opened 2020-04-30
Modified "dodqi" to accommodate the A-to-D saturation threshold

mdlpstsci:
The new A-to-D saturation threshold is no longer a constant value and has different behavior pre- and post-SM4.  The CCDTAB files have been updated by the ACS team to contain a new column, ATODSAT, which contains the saturation threshold value.

Please be aware the old code compared a science pixel value to the saturation threshold as only **_greater than_**.  In contrast, the new code is **_greater than or equal to_** as specified in the issue.

This PR fixes Git Issue #460.

### DIFF
--- a/pkg/acs/calacs/Dates
+++ b/pkg/acs/calacs/Dates
@@ -1,3 +1,6 @@
+ 29-Apr-2020   CALACS 10.2.3 Modified "dodqi" to accommodate the A-to-D saturation threshold 
+                             which is now defined in the CCDTAB. The threshold is no longer 
+                             a constant and has different behavior pre- and post-SM4.
  10-Jan-2020   CALACS 10.2.2 Removed hard-coded darktime scaling value and read new 
                              post-flashed and unflashed columns from updated CCDTAB reference 
                              file to use for the offset to the DARKTIME FITS keyword value 
--- a/pkg/acs/calacs/History
+++ b/pkg/acs/calacs/History
@@ -2,7 +2,7 @@
     Modified "dodqi" to accommodate the A-to-D saturation threshold 
     which is now defined in the CCDTAB. The threshold is no longer 
     a constant and has different behavior pre- and post-SM4.
-    Updated CCDTAB files for HRC and WFC.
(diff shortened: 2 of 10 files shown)


mdlpstsci:
From: Meaghan McDonald <mmcdonald@stsci.edu>
Date: Wednesday, May 6, 2020 at 3:27 PM
To: Tyler Desjardins <desjard@stsci.edu>, Michele De La Pena <mdelapena@stsci.edu>
Subject: Re: ACS A-to-D saturation

Hi there,
Everything looks good to me! Thanks for putting that together Tyler. Michele, feel free to proceed with whatever comes next. I know there was talk today at the DMS meeting of possibly implementing this build before the astrometry fixes are finalized (because wfc3 needed some more time I think?), so just let us know if this is the case. Thanks!
Best,
Meaghan
 
From: Tyler Desjardins <desjard@stsci.edu>
Date: Wednesday, May 6, 2020 at 12:14 PM
To: Michele De La Pena <mdelapena@stsci.edu>, Meaghan McDonald <mmcdonald@stsci.edu>
Subject: Re: ACS A-to-D saturation
 
Hi all,
Attaching a jupyter notebook that demonstrates the test I did, but suffice it to say that it looks good to me. When I check the pixels that should be above the A-to-D saturation threshold from the CCDTAB and then check how many pixels are flagged, I get identical numbers both pre- and post-SM4. If Meaghan concurs, then I say we give the thumbs up on this change and can progress to the next step in the workflow.
-Tyler

mdlpstsci:
Regression tests were failing due to use of -std=C99 in wscript.  The fix implemented in PR #485 fixes these regression test failures, as well as the initial reason -std=C99 was added in the first place.   For completeness, the -std=C99 (=gnu99 better) was needed because "for loop initial declaration" was used in this ticket.

mdlpstsci:
This PR was previously failing due to the "-std=C99" option on the compile, and this was resolved in the latest update by pulling in Joe's change to use "-std=gnu99".  However, now three different tests (as compared to the previous failures) are failing due to socket timeouts.

PR #518, opened 2020-08-12
WFC3 Time-dependent photometry update

mdlpstsci:
imphttab
- Fixed an existing bug in getphttab.c by ensuring a string was set to
  the empty string.  The bug would only be manifested for multiple Imsets
  in an input file, in conjunction with a parameterized value.
wfc3
- Added MJD as a parameterized variable for the PHOTMODE keyword to
  enable a time-dependent photometric correction for the UVIS only.
- Removed appending "CAL" to the PHOTMODE keyword.

### DIFF
--- a/pkg/imphttab/getphttab.c
+++ b/pkg/imphttab/getphttab.c
@@ -15,6 +15,10 @@ PLL, 12/2013 Allow MJD extrapolation using simple straight line from
 MLS: 07/2015 Cleand up for unused variables and warning
 MLS: 08/2015 Added some initializations the clang complained about
 
+MDD: 03/2020 Ensure obs->obsmode is an empty string at the beginning of
+             this routine - this was a lurking bug raised in the case of 
+             multiple Imsets and parameterized value.
+
 */
 # include <stdio.h>
 # include <string.h>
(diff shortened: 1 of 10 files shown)


mdlpstsci:
The photometry algorithm has changed in a manner which is **_not_** backwards compatible to previous results.  Once the algorithm has been vetted by the WFC3 team, new truth files will have to be generated.

catherine-martlin:
@mdlpstsci - Just a note that I sent the UVIS IMPHTTAB that you needed in an email. Our testing of the updates went well and we're ready for the next steps.

mdlpstsci:
On 18 September 2020 Catherine Martlin emailed:
Hi Matt and Michele,

Attached you’ll find the new UVIS IMPHTTAB to use with testing with the WFC3 code updates by Michele. We’ve finished our review of her updates and are happy with them and ready for the regression testing to start.

Please let me know if there is anything else you need from us to produce the regression test data – I’ll be the one looking at it once it’s ready for us to inspect.

Thanks! Best wishes,

Catherine

stsci-hack (approved):
The changes look fine to me.

PR #523, opened 2020-09-10
EXP: Multi arch build

pllim:
@sean-lockwood asked and now I am curious.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: c
 env:
   global:
     # Set defaults to avoid repeating in most cases
+    - EXE_PREFIX="/tmp/miniconda3"
     - PYTHON_VERSION=3.7
     - WAF_ARGS_CONFIGURE="-v"
     - WAF_ARGS_BUILD="-v"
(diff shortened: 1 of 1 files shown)


sean-lockwood:
I was able to manually build HSTCAL for ARM32 on my Raspberry Pi 4B (Raspberry Pi OS, Buster; 32-bit OS).
An initial reduction of a STIS dataset gave identical results to MAST.

pllim:
Sean, great! Now we don't need AWS no more. An army of Raspberry Pi probably cheaper. 😉

pllim:
Hmm. How did you install cfitsio there? Apparently there is no conda available for ARM64.

sean-lockwood:
Built from source.  There were some library dependencies I had to track down and "apt install".

sean-lockwood:
And apparently while the 4B is a 64-bit architecture, the OS is still 32-bit for now.  They're testing a 64-bit version...

pllim:
Interesting... You should try to install PyRAF there next while it is still 32-bit.

pllim:
Re: https://github.com/spacetelescope/hstcal/pull/523#issuecomment-690328854

Ouch. Definitely a bug. Do you want to open PR to fix them? We don't support Python 2 anymore.

pllim:
p.s. No idea where that file came from, the git blame makes no sense to me. It wasn't modified for 10 years.

pllim:
So, looks like it is possible to build HSTCAL on all the architectures supported by Travis CI on Linux. You just can't use `conda`. Case closed.

pllim:
p.s. Building it is one thing, making sure the numbers are still correct is another...

PR #539, opened 2021-01-12
Issue453 acs satmap

### LINKED ISSUE #453: ACS Saturation Map
The ACS Branch would like to update `calacs` to do full-well saturation flagging (currently part of DQICORR) using an image rather than the current single value in CCDTAB. The CCDTAB uses a single value of full-well saturation for data quality flagging for ALL amplifiers, whereas studies have shown that the full-well spatially varies across each amplifier region. Work on generating these maps is ongoing, and currently there are some open questions on the exact implementation. There are two possible scenarios being discussed:

1. The image (which we'll call "SATUFILE" for the moment) would be checked against during DQICORR. The image would be in units of DN and have two extensions of sizes 4144 x 2068 for WFC and a single 1062 x 1044 extension for HRC. The check would be controlled by the DQICORR header keyword. No check would be made for SBC.

2. Similar to 1 above, but the check would happen at the end of BIASCORR. The image would be in units of electrons and would have calibrated image dimensions after overscan trimming has occurred. 

The ACS Branch will work with CRDS/ReDCaT on implementing the new reference file.

**Aiming to implement this before end of FY2020 Q2 (March 2020) if possible.**

mdlpstsci:
The CALACS pipeline has been updated to incorporate a full two-dimensional image to be used for full-well saturation flagging.  Essentially, the code which performed the full-well flagging using a single value in dodqi.c has been removed.  A new "step" has been added to the pipeline after BIAS correction, BLEV correction, and conversion to electrons - doFullWellSat().

This PR fixes #453.


mdlpstsci:
Note: The regression tests will fail as the new input and truth files, as well as the new calibration file, have not yet been put into Artifactory.

pkg/acs/calacs/Dates
```diff
@@ -1,3 +1,8 @@
+ 31-Dec-2020   CALACS 10.2.5 Implementation to use an image to detect and flag full-well
```
stsci-hack, line 1: It seems to me that a new step would warrant an update from 10.2 to 10.3, whereas a bug fix in a current routine would warrant an update from 10.2.4 to 10.2.5.

stsci-hack (dismissed):
From what I can remember of my C, and the logic in CALACS, this PR looks fine to me.

mdlpstsci:
Code has already been reviewed though changes were made to read the entire image into memory versus reading the image into memory line-by-line.  The four failing regression tests are WFC3, and they are failing as the Artifactory input and truth files have already been updated, but the code is still pending.  This ticket (code) is going to be merged first.

PR #541, opened 2021-02-11
Added missing import for pytest to the STIS test

mdlpstsci:
The import statement for pytest was added to the test in order to allow the test to be marked properly for skipping (pytest.mark.skip).


pllim (approved):
That makes sense.

stsci-hack (approved):
Seems logical...

PR #548, opened 2021-04-28
Fix the name of the standalone executable acscteforwardmodel.e

mdlpstsci:
The name of a standalone executable inadvertently was changed when code was re-arranged by the build manager to address circular references. Changed acsforwardmodel.e back to acscteforwardmodel.e
to aid documentation and script issues.


stsci-hack (dismissed):
Simple enough.

Issue #489, opened 2020-05-18
Post-flashed ACS CRJ/CRC images are not fully flash-corrected


jryon:
Post-flashed CRJ/CRC images show residual flash background after flash correction. The FLASHDUR keyword is the same in the FLT of a single CR-SPLIT as in the CRJ of the sum of CR-SPLITs.

The solution should be to sum the FLASHDUR keywords of the input images to acsrej, and save the sum to the FLASHDUR keyword of the output CRJ/CRC. The FLASHDUR keyword is used directly in acs2d to scale the flash reference file for flash correction.

mdlpstsci:
@jryon Do you have a dataset that would be good for testing this change?

jryon:
@mdlpstsci  jdzn01b6q and jdzn01b7q could be used to test. They are post-flashed, CR-split images, and also 1k subarrays. I can look for some full-frame data if you'd prefer that. Their association ID is jdzn01010, if you need it.

mdlpstsci:
@jryon Thanks.  Let me start with this association.

mdlpstsci:
@jryon Before I make the software available for you for any testing you might like to do, I have put a tar file on plhstins1 in /home/mdelapena/FOR_JENNA/imp_flash_test.tar.  This file contains the "jdz" data processed with the current release of CALACS (BADFLASH subdirectory) and the modified CALACS (NEWFLASH subdirectory).  Please see if this newly processed data meets your expectations.

jryon:
@mdlpstsci Thanks! The NEWFLASH data meets expectations. The FLASHDUR keyword in the NEWFLASH CRJ is the sum of the individual exposure FLASHDURs, and there is no obvious post-flash signal remaining in the NEWFLASH CRJ (such signal is clearly visible in the BADFLASH CRJ). I also measured a sigma-clipped median of the background with the bright star masked out for both CRJs: BADFLASH median = 33.8 e-, NEWFLASH median = 1.7 e-. The remaining signal in the NEWFLASH CRJ is likely to be sky background, but I can investigate that more closely when the software is available.

mdlpstsci:
@jryon I have put the "test" executable for calacs.e which incorporates the flash duration fix on plhstins1.  You should just be able to invoke my executable (!) and process data.  In the past I have had folks do a bit more setup, but I am not sure this was really necessary.  My executable is where CALACS is Version 10.3.3 (03-May-2021):
[...]

PR #550, opened 2021-05-06
Issue489 fixflash

mdlpstsci:
Fixed a bug associated with the FLASHDUR computation.  The FLASHDUR keyword was updated in the CRJ/CRC headers to be the sum of the individual input images as this keyword is used in ACS2D to scale the flash reference file for flash correction.

### DIFF
--- a/pkg/acs/calacs/acsrej/acsrej_do.c
+++ b/pkg/acs/calacs/acsrej/acsrej_do.c
@@ -41,6 +41,8 @@ static void closeSciDq (int, IODescPtr [], IODescPtr [], IODescPtr [], clpar *);
                                 Added doc. Cleaned up codes.
   27-Feb-2019   M.D. DeLaPena   Read the IMAGETYP keyword from Primary of first image
                                 as this info is needed by acsrej_loop.
+  30-Apr-2021   M.D. DeLaPena   Read the FLASHDUR keyword from Primary of all the input
+                                images in order to compute the cumulative value.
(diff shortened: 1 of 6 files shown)


pkg/acs/calacs/acsrej/acsrej_do.c
```diff
@@ -395,6 +399,7 @@ int acsrej_do (IRAFPointer tpin, char *outfile, char *mtype, clpar *par,
                    "Total sky level (electrons)");
         PutKeyFlt (sg.globalhdr, "REJ_RATE", (float)nrej / texpt,
                    "Cosmic ray impact rate (pixels/sec)");
+        PutKeyFlt (sg.globalhdr, "FLASHDUR", cumFlashDur, "");
```
stsci-hack, line 402: This new keyword should have a comment to explain what it is; specifically, total/cumulative flash duration in seconds (or minutes or hours or mjd? ).

stsci-hack (changes requested):
The new keyword should have some description when added to the header, just to be nice to the user.

pkg/acs/calacs/acsrej/acsrej_do.c
stsci-hack: That's fine.

stsci-hack (approved):
LGTM

mdlpstsci:
Resolves Issue #489.

PR #551, opened 2021-05-24
Fixed a bug associated with the DARKTIME computation and utilization

mdlpstsci:
Fixed a bug associated with the DARKTIME computation Code written for CALACS 10.2.2 was moved from ACS2D to ACSCCD so every BLV_TMP file has the correct DARKTIME keyword value.  Also, the DARKTIME keyword was updated in the CRJ/CRC
headers to be the sum of the individual input images. Only the DARKTIME value in the input file header is used to scale the dark reference file for the dark correction.

This PR fixes Git Issue#490.


mdlpstsci:
Jenna Ryon of ACS has approved of this bug fix (Git Issue#490) in terms of its scientific validity.  She has also indicated this fix is fine to be part of the same build with Git Issue#489 (PR#550).  New truth files will be submitted to Artifactory so the regression tests can succeed.

stsci-hack (approved):
This looks fine to me.

PR #556, opened 2021-07-21
Use strncpy for inlist in acsccd.e, better error

### LINKED ISSUE #554: acsccd.e crashes when input is too long
As discovered by @samanthalh when running it with `acstools.acsccd.acsccd('*raw.fits')` on too many files. That command expands to `acsccd.e file1,file2,file3,...` . Also see INC0169541 . `acsccd.e` crashes ungracefully with no helpful error message.

Offending lines:

https://github.com/spacetelescope/hstcal/blob/d395b8b88c2f4ccad65c6a3f176bf1907066c12d/pkg/acs/calacs/acsccd/mainccd.c#L106

https://github.com/spacetelescope/hstcal/blob/e0f20b7810862ed790e9c4989798eb51a8ec5bc1/include/hstcal.h#L11

pllim:
Fix #554 by exiting "gracefully" and giving more useful hint on why it failed. With this patch:

```
ERROR: Input path exceeds maximum supported length (255)
syntax:  acsccd [--help] [-t] [-v] [-q] [--version] [--gitinfo] input output
```

Out of scope: I have no WBS to charge to, so I cannot apply this to CALACS globally. This PR is only meant to solve the immediate problem reported originally over at ACSTOOLS.

### DIFF
--- a/pkg/acs/calacs/acsccd/mainccd.c
+++ b/pkg/acs/calacs/acsccd/mainccd.c
@@ -117,10 +117,6 @@ int main (int argc, char **argv) {
         freeOnExit(&ptrReg);
         exit (ERROR_RETURN);
     }
-    inlist[0] = '\0';
-    outlist[0] = '\0';
-    input[0] = '\0';
-    output[0] = '\0';
 
     /* Initialize the lists of reference file keywords and names. */
     InitRefFile (&refnames);
@@ -180,7 +176,7 @@ int main (int argc, char **argv) {
                 }
             }
         } else if (inlist[0] == '\0') {
-            strcpy (inlist, argv[i]);
+            strncpy (inlist, argv[i], sizeof(inlist));
         } else if (outlist[0] == '\0') {
             strcpy (outlist, argv[i]);
         } else {
--- a/pkg/acs/calacs/acsccd/mainccd.c
+++ b/pkg/acs/calacs/acsccd/mainccd.c
@@ -62,6 +62,7 @@ int main (int argc, char **argv) {
     int verbose = NO;	/* print additional info? */
     int quiet = NO;	/* print additional info? */
     int too_many = 0;	/* too many command-line arguments? */
+    int too_long = 0; /* command-line argument too long? */
     int i, j;		/* loop indexes */
     int k;
 
@@ -142,6 +143,7 @@ int main (int argc, char **argv) {
             switch_on = 1;
         } else if (argv[i][0] == '-') {
         **********/
+        too_long = strlen(argv[i]) > CHAR_LINE_MAX;
         if (argv[i][0] == '-') {
             if (!(strcmp(argv[i],"--version")))
             {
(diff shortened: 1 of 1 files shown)


pkg/acs/calacs/acsccd/mainccd.c
```diff
@@ -117,10 +117,6 @@ int main (int argc, char **argv) {
         freeOnExit(&ptrReg);
         exit (ERROR_RETURN);
     }
-    inlist[0] = '\0';
```
pllim, line 120: This should not be needed because `calloc` was used. -- Joe

jhunkeler:
I couldn't let it go... The comma delimited argument list you sent me earlier today was 455 bytes long.

https://github.com/spacetelescope/hstcal/blob/d395b8b88c2f4ccad65c6a3f176bf1907066c12d/include/hstcal.h#L11

This is only allocating 256 bytes to hold your 455 byte string. 

https://github.com/spacetelescope/hstcal/blob/d395b8b88c2f4ccad65c6a3f176bf1907066c12d/pkg/acs/calacs/acsccd/mainccd.c#L106

Someone should go through a change these arbitrary values to use `PATH_MAX` from `limits.h`, or define a reasonable (large) default value. `PATH_MAX` on MacOS is 1024 and on Linux 4096. And yeah the POSIX standard sets the maximum path size at 256 because it's old, however the _kernel_ should be guaranteed to honor whatever `PATH_MAX` is set to.

```c
#include <stdlib.h>  // calloc
#include <string.h>  // strncpy
#include <limits.h>  // PATH_MAX

// in hstcal.h
#define CHAR_LINE_LENGTH PATH_MAX

// and everywhere paths are accepted as input
char *path;
path = calloc(CHAR_LINE_LENGTH, sizeof(char));
if (!path) {
    // die
}
strncpy(path, argv[i], CHAR_LINE_LENGTH - 1);
```

pllim:
Changing that to PATH_MAX affects the whole HSTCAL, including variable that has nothing to do with paths. Also would increase the memory footprint overall.

https://github.com/spacetelescope/hstcal/blob/e0f20b7810862ed790e9c4989798eb51a8ec5bc1/pkg/wfc3/calwf3/lib/imghistory.c#L21

I think it is a known (but perhaps not well documented) limitation that the input length cannot handle too long of an input. All I am trying to do here is to have it crash with a slightly more sane error. Now you can see that the input is truncated and perhaps have an idea that input list is maybe too long. 😬

pllim:
For my own record:

```
$ grep PATH_MAX /usr/include/linux/limits.h
```
```c
#define PATH_MAX        4096    /* # chars in a path name including nul */
```

jhunkeler:
I see. Then how about this?

```c
#include <stdlib.h>  // calloc
#include <string.h>  // strncpy
#include <limits.h>  // PATH_MAX

// and everywhere paths are accepted as input
char *path;
path = calloc(PATH_MAX + 1, sizeof(char));
if (!path) {
    // die
}
strncpy(path, argv[i], PATH_MAX);```

pllim:
> https://github.com/spacetelescope/hstcal/pull/556#issuecomment-884868967

Sure, I do not disagree, but it is also not my call to make. Let's see what @mdlpstsci or @stsci-hack say...

mdlpstsci (approved):
Straightforward fix for a straightforward issue.

pllim:
Since this is approved, I am taking the liberty to merge. Thanks for the review!

PR #560, opened 2022-01-11
Fixed bug triggered ONLY when executing the condaforge on MacOS

mdlpstsci:
Fixed a bug which was triggered ONLY when executing the condaforge
distribution on MacOS where failure was due to a missing return statement.
Only call computeDarktime() once to compute the updated
DARKTIME primary keyword and function return is now void. Update
the keyword in main to avoid passing extra data by reference.


stsci-hack (approved):
Certainly simplifies that call.  LGTM.

mdlpstsci:
Waiting for actual PyTests on my machine to finish.

PR #561, opened 2022-02-08
ACSREJ algorithm modified for ERR array

### LINKED ISSUE #371: ACSREJ: ERR extension is computed, rather than propagated
In the process of reviewing some previous updates regarding the treatment of error by the `ACSREJ `algorithm, I realized that the errors aren't being propagated. Instead at the end of the combination process, the errors are computed based off a model for what they should look like. While this isn't a bad approximation, it is incorrect and will wipeout contributions from the superbias. 

As a test case, I took two `FLT` images and replaced their ERR arrays with all 0's. I then combined those two images using `acstools` to create jd5702091_no_err_crj.fits and then examined the output SCI and ERR extension. If `ACSREJ` were propagating the error from the input images, then you would expect the final ERR extension of the combined product to be all zeros. In the figure below, you can see that this is not the case.
example

mdlpstsci:
Update to the cosmic ray rejection algorithm as to the way the output ERR extension is computed for the CRJ/CRC file. The output ERR is now propagated from the usable input ERR extensions versus being computed by a model.

Resolves #371.

### DIFF
--- a/pkg/acs/calacs/Dates
+++ b/pkg/acs/calacs/Dates
@@ -1,9 +1,8 @@
- 11-Jan-2022   CALACS 10.3.4 Update to the cosmic ray rejection algorithm as to the way
+ 08-Feb-2022   CALACS 10.3.5 Update to the cosmic ray rejection algorithm as to the way
                              the output ERR extension is computed for the CRJ file. The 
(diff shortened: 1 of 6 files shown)


stsci-hack (approved):
This makes sense to me and looks good.

PR #567, opened 2022-05-09
Fix dist tag resolution

jhunkeler:
This change covers an edge case for when `./waf dist` is executed from a branch outside of `master` and no tag is associated with the current commit it returns `UNKNOWN`. All this does is add fallback behavior to ensure `VERSION` gets populated with the closest parent tag rather than `UNKNOWN`.

### DIFF
--- a/wscript
+++ b/wscript
@@ -218,10 +218,10 @@ def _get_git_details(ctx):
 
     # Report a shadowed release tag if one exists, otherwise fallback to
     # the generic description method. The generic method does NOT
-    # account for shadows. This might retun a release candidate tag if the point
+    # account for shadows. This might return a release candidate tag if the point
     # release tag shares the same commit hash. This is purely aesthetic.
     #
-    # To avoid this scenario one must commit a change to teh repository prior to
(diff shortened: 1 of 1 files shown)


mdlpstsci (approved):

wscript
```diff
@@ -212,15 +212,27 @@ def _get_git_details(ctx):
...
+    cmd_branch = 'git rev-parse --abbrev-ref HEAD'
+
+    # Report a shadowed release tag if one exists, otherwise fallback to
+    # the generic description method. The generic method does NOT
+    # account for shadows. This might return a release candidate tag if the point
+    # release tag shares the same commit hash. This is purely aesthetic.
+    #
+    # To avoid this scenario one must commit a change to the repository prior to
+    # tagging a final point release.
+    tmp = call(cmd_describe + ' --contains') or \
```
pllim, line 226: Tsk tsk, Joe. Do you really have to use `\` continuation? 😉

jhunkeler: Yes. `\`
:laughing:

PR #570, opened 2022-09-16
Modify wscript as OS no longer returns patch version

mdlpstsci:
MacOS Monterey no longer returns a patch version for 
`/usr/bin/sw_vers -productVersion`.  Solution was to remove the obsolete code.

PR #571, opened 2022-09-16
wscript: Fix _check_mac_osx_version() on modern MacOS systems

jhunkeler:
If there's no PATCH version available then handle it.

Close #570

### DIFF
--- a/wscript
+++ b/wscript
@@ -317,7 +317,13 @@ def _check_mac_osx_version(floor_version):
         return None
 
     # Extract the integer values between the '.'s
-    osx_version_major, osx_version_minor, osx_version_patch = tuple(int(x) for x in s.strip().split('.'))
+    osx_version_data = tuple(int(x) for x in s.strip().split('.'))
+    osx_version_major = osx_version_data[0]
+    osx_version_minor = osx_version_data[1]
+    osx_version_patch = 0
+
+    if len(osx_version_data) > 2:
+        osx_version_patch = osx_version_data[2]
 
     # Convert major/minor/patch values into a single 24-bit integer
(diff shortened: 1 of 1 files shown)


jhunkeler:
#570

wscript
```diff
@@ -317,7 +317,13 @@ def _check_mac_osx_version(floor_version):
...
 
     # Extract the integer values between the '.'s
-    osx_version_major, osx_version_minor, osx_version_patch = tuple(int(x) for x in s.strip().split('.'))
+    osx_version_data = tuple(int(x) for x in s.strip().split('.'))
+    osx_version_major = osx_version_data[0]
+    osx_version_minor = osx_version_data[1]
+    osx_version_patch = 0
+
+    if len(osx_version_data) > 2:
+        osx_version_patch = osx_version_data[2]
```
pllim, line 326:
```suggestion
    if len(osx_version_data) > 2:
        osx_version_patch = osx_version_data[2]
    else:
        osx_version_patch = 0
```

pllim:
I dunno why Jenkins has a ❌ . I cannot access the log.

mdlpstsci:
The problem is accessing 'ssb.stsci.edu' as there is a certificate issue.  This has been a problem for many folks and there is a ticket out for ITSD to address the issue.

wscript
mdlpstsci: Both solutions look fine though I prefer the JH solution as the variable has a value by default.

mdlpstsci (approved):

pllim:
Feel free to ignore my suggestion then. Thanks, everyone!

PR #580, opened 2023-04-07
Updated JenkinsfileRT

mdlpstsci:
Updated Python version, removed deprecated channel, added a new nodetype of "linux".  Deleted Jenkinsfile as it does not seem to be used.


JenkinsfileRT
```diff
@@ -49,7 +51,11 @@ bc.failedFailureThresh = 6
 // Upload artifacts
 bc.test_configs = [data_config]
 
+bc1 = utils.copy(bc)
+bc1.name = "linux"
```
pllim, line 55: Does "linux" use something that is not RHEL?

PR #582, opened 2023-04-26
WFC3/UVIS: Removed deprecated tests and fixed bug in ASN tests

mdlpstsci:
Removed long since deprecated (skipped) tests which were not useful or
were duplicates.  Fixed UVIS and IR ASN tests so all the intended files are compared on output instead of just the last file.  The outputs list was previously overwritten instead of appended.


tests/wfc3/test_ir_03asn.py
```diff
@@ -22,7 +22,7 @@ def test_ir_03asn(self):
         flist = ['ib1f23qmq', 'ib1f23qlq', 'ib1f23qjq', 'ib1f23qhq']
 
         for rn in flist:
-            outputs = [(rn + '_flt.fits', rn + '_flt_ref.fits'),
+            outputs += [(rn + '_flt.fits', rn + '_flt_ref.fits'),
```
raswaters, line 25: Define outputs before the loop. Same below.

raswaters (approved):

PR #583, opened 2023-04-28
Set certain WFC3/UVIS tests to XFAIL

mdlpstsci:
Set certain WFC3/UVIS tests to XFAIL as the truth files on Artifactory have been updated for the new software, but this software is NOT being delivered as part of the build (HSTDP-2023.1.0) to operations.

The issue is there are many packages which are part of the "operations build".  In order for the nightly regression tests to pass, when new software is checked in, the corresponding Truth files on Artifactory must be updated.  However, if the software is not to be included in the next build to be delivered, the PyTests for this package will fail as there is only one source of Input and Truth files.  

This means there must be manual gating of the packages which contribute to the build which is not realistic.

### DIFF
--- a/tests/wfc3/test_uvis_01asn.py
+++ b/tests/wfc3/test_uvis_01asn.py
@@ -4,6 +4,7 @@ import pytest
 from ..helpers import BaseWFC3
 
 
+@pytest.mark.xfail
 class TestUVIS01ASN(BaseWFC3):
     """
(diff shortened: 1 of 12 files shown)


pllim:
As long as there is a follow-up action item to remove xfail in the future?

mdlpstsci:
Indeed, the XFAIL will be removed once the final HSTDP build is released.

Issue #558, opened 2021-11-09
Add support to calwf3 for Full Well Saturation Maps (SATUFILE)


mackjenn:
Similar to changes recently implemented for calacs (acsccd), the WFC3 branch would like to update calwf3 (wf3ccd) to flag Full-well saturation (currently part of DQICORR) using an image rather than a single scalar value. 

For reference see the hstcal issue and associated pull request:
            https://github.com/spacetelescope/hstcal/issues/453
            https://github.com/spacetelescope/hstcal/pull/539

The CCDTAB uses a single value of full-well saturation for data quality flagging for ALL amplifiers, whereas studies have shown that the full-well spatially varies across each amplifier region. The new reference file SATUFILE is 2D full-well saturation image and will be used to detect and flag pixels versus using the simple scalar value defined in the CCDTAB.

This check would happen at the end of BIASCORR, where the SATUFILE reference file will be in units of electrons and will have calibrated image dimensions after overscan trimming has occurred. The WFC3 Branch will work with CRDS/ReDCaT on implementing the new reference file.

The goal is to implement initial functionality by the end of FY2021 Q2 (March 2021) if possible, but the timeline is flexible depending on the availability of @mdlpstsci .

mdlpstsci:
Sent note to Jennifer and Sylvia regarding a timeframe for their creation of the new saturation image.

mackjenn:
A test version of the UVIS satufile for calwf3 development may be found here:
/user/mack/uvis_satufile/wfc3_uvis_sat.fits

mdlpstsci:
@mackjenn Thanks.  I got the test version of the satufile.  Are there binned WFC3 images?

mdlpstsci:
@mackjenn Ooops.  I finally read your email regarding eventually making files for binned images.  I have my answer!

mackjenn:
@mdlpstsci 
Yes, UVIS has binned modes, and we'll need to make extra saturation reffiles and add some keywords to the file we just gave you.  For reference, here's an example of a 1x1, 2x2, and 3x3 binned FLSHFILE in /grp/hst/cdbs/iref/ and corresponding keywords for the two CCDs.  Note the binning is done on-orbit before readout.

Hopefully you can use the same calwf3 logic for the  binned SATUFILE as for the binned FLSHFILE

[...]

Issue #584, opened 2023-05-15
SATUFILE - investigate level of work for 'backward compatibility'


mackjenn:
Note: this request is related to PRs [#563](https://github.com/spacetelescope/hstcal/pull/563), [#588](https://github.com/spacetelescope/hstcal/issues/558), [HSTSDP-1922](https://jira.stsci.edu/browse/HSTSDP-1922) and [HLA-711](https://jira.stsci.edu/browse/HLA-711)

A.) **Backward compatibility** - ability to run new software on older data without the SATUFILE keyword
   Solution 1:  Users manually add the new keyword to run new software
   Solution 2:  Michele modify calwf3 to build back in original logic which uses the CCDTAB if the keyword is missing

From Jennifer:  _Let's evaluate the estimated level of work required before making any changes, especially if Solution 1 is deemed acceptable for users._ 

From Michele (May 9th email): 
"I will first confirm with the datasets Isabel gave to me that the same answer is achieved both the OLD and the NEW way.  Then I have to put back all of the old code (carefully), as well as the little bits and pieces of associated information I cleaned up, as well as add new code to accommodate what is needed for essentially choosing which algorithm to use. Isabel will need to re-test and re-approve."

 B.) **Forward compatibility** - ability to run old software on new data from MAST  
     No work required from Michele.  Isabel will check that the older calwf3 will simply ignore the new keyword 

@mdlpstsci @starivera @astromariarosa

mdlpstsci:
Time Estimates:

1. Confirm the effectively same result is achieved using the _OLD_ and the _NEW_ methods, and discuss the results with the team.  If the results are not the same [byte-to-byte or within a tolerance?] (1 day)

2. Put back the old code and test with no new control logic. (3 days)

3. Add control logic to use the _OLD_ method if the SATUFILE keyword is missing, _**OR**_ add an option to CALWF3 with allows the user to choose the _OLD_ or the _NEW_ method and test.  If a new option is added, any sub-component APIs (e.g., wf3ccd) will also have to be modified to accommodate the new option.  New options require both the C components, as well as the Python wrappers, to be modified.  The new option may be more user-friendly, but the amount of work does expand! (TBD depending what WFC3 really wants.) (2 days for the simple solution, 5 days for the new co
4. mmand line option choice)

5. Integrated testing (2-3 days)

6. Testing by INS (? days)

@mackjenn @starivera @astromariarosa

mdlpstsci:
[...]

PR #585, opened 2023-08-07
Implement new (via image) and maintain old (scalar) saturation image flagging

mdlpstsci:
Completed implementation to use an image to detect and flag full-well saturation versus a simple scalar.  The new routine, dofwsat.c, is similar to what has been done for calacs. The WFC3 implementation is more complicated in that there are serial virtual overscan columns, as well as binned images, to accommodate.  The detection/flagging occurs after blev and bias corrections while the output is still in counts.  If the SATUFILE keyword is missing from the FITS header, or the keyword does not have a valid filename as a value, the code will revert to using the original method of flagging full-well saturation.  The flagging will be done in doDQI and use a single value as the saturation threshold.

### DIFF
--- a/pkg/wfc3/calwf3/lib/dodqi.c
+++ b/pkg/wfc3/calwf3/lib/dodqi.c
@@ -190,13 +190,13 @@ SingleGroup *x    io: image to be calibrated; DQ array written to in-place
 	if (wf3->dqicorr != PERFORM && wf3->dqicorr != DUMMY)
 	    return (status);
 
-    /* Issue a message so it is clear that saturation flagging is 
-       being done here in DODQI using scalar *IF* wf3->scalar_satflag is True.
-    */
-    if (wf3->scalar_satflag == True) {
-        sprintf (MsgText, "Full-well saturation flagging being applied during doDQI using a single threshold value.");
-        trlmessage (MsgText);
-    }
+	/* Issue a message so it is clear that saturation flagging is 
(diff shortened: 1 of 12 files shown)


pkg/wfc3/calwf3/lib/dodqi.c
```diff
@@ -182,7 +190,16 @@ SingleGroup *x    io: image to be calibrated; DQ array written to in-place
 	if (wf3->dqicorr != PERFORM && wf3->dqicorr != DUMMY)
 	    return (status);
 
+    /* Issue a message so it is clear that saturation flagging is
```
raswaters, line 193: Indent lines 193-199

pkg/wfc3/calwf3/lib/dodqi.c
```diff
@@ -194,6 +211,14 @@ SingleGroup *x    io: image to be calibrated; DQ array written to in-place
 					sum_dq = DQPix (x->dq.data, i, j) | ATODSAT;
 					DQSetPix (x->dq.data, i, j, sum_dq); /* atod sat */
 				}
+
+				if (wf3->scalar_satflag == True) {
```
raswaters, line 215: Are lines 215-2212 to be executed after lines 210-213, or instead of?

mdlpstsci: Lines 215-221 are executed after Lines 210-213.

raswaters (approved):

mdlpstsci:
Resolves Issue #558 #584

PR #598, opened 2024-01-12
HLA-1178:  Garbled text reported in trailer file causes failure in Cloud processinge

mdlpstsci:
Addressed Jira Ticket: HLA-1178

Fixed the bug where the structure associated with the SPOTTAB variables was not initialized.  Also, fixed the issue where the PEDIGREE and DESCRIP keywords were deliberately being read from the wrong HDU.  Only in the case of tabular reference files were the mentioned keywords being read from the wrong HDU.  Fixed so the keywords are now being read from the primary HDU and properly logged to the trailer file.


stscirij (approved):
Looks good to me

raswaters (approved):

PR #603, opened 2024-02-22
MNT: Bump checkout version

pllim:
Looks like your workflow using outdated checkout.

Ehem, @jhunkeler , where is the matrix for OSX ARM?

* https://github.com/actions/runner-images/issues/9254


jhunkeler:
Good catch... Thanks.
> Ehem, @jhunkeler , where is the matrix for OSX ARM?

To be fair the M1 runners weren't available when I wrote the config :grin:

jhunkeler (approved):

PR #608, opened 2024-02-22
Update LICENSE year

pllim:
Pretty sure hstcal existed before 2018, so I guess you are only keeping the current year, and not a range?

p.s. I mainly wanna see if the labeler works but you might as well merge this.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2024, Association of Universities for Research in Astronomy (AURA)
+Copyright (c) 1997-2024, Association of Universities for Research in Astronomy (AURA)
 All rights reserved.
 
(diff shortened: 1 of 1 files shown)


jhunkeler:
If we want to do a range then there are two starting dates to pick from. :grin: 

The first commit on this repository:

```
commit f6638bbcb7605d72e17f2696305c42bb35a44e98
Author: hack <hack at stsci.edu>
Date:   Thu Mar 11 18:20:58 2010 +0000
```

Or the genesis of `hst_calib.cl` in `stsdas`...

```
commit ec9d555c5fd92591a3d72d21877486345f9a91fe
Author: sdas <sdas@fe389314-cf27-0410-b35b-8c050e845b92>
Date:   Thu Jul 31 18:31:24 1997 +0000
```

mdlpstsci (approved):
Correction done to dates of package origin.

PR #610, opened 2024-03-27
Remove test script from hstio

jhunkeler:
- This removes the test script from `hstio/test` to avoid pytest picking it up when `pytest` is executed without arguments.
- The original C code and tiny FITS input file are retained


mdlpstsci (approved):
buh bye

PR #612, opened 2024-04-16
WIP: HLA-1036: An 'MJD#' value is added to the PHOTMODE keyword in the PHDU

mdlpstsci:
Add the 'MJD#' value to the PHOTMODE keyword in order to enable a time-dependent photometric correction for the IR.  By adding the 'MJD#' value to the PHOTMODE keyword string value, the algorithm now must search the new IMPHTTAB file for a row whose first column, OBSMODE, matches the string. This supplements the change done previously for the UVIS in 2020.


stscirij (approved):
Looks good - same as the corresponding code for ACS

PR #613, opened 2024-05-17
Update INSTALL.md

tmonroeST:
Had to fix typo to get cmake to recognize local install directory.  Verified on Linux install.


tmonroeST:
Sean Lockwood suggested this fix.

PR #615, opened 2024-06-05
Update the JenkinsfileRT 

mdlpstsci:
Update the JenkinsfileRT for Python 3.11 and ensure some regression tests run on the MacOS by adding the _--bigdata_ option.


mdlpstsci:
We are moving away from the use of Jenkins, so I am just closing this PR.

PR #616, opened 2024-06-11
Move test dependency installation to test_cmds list

jhunkeler:
This change avoids the following error produced by the delivery system when `hstcal` is tested _before_ any Python packages have been installed.

```
FileNotFoundError: [Errno 2] No such file or directory: 'pytest'
script returned exit code 1
```


mdlpstsci (changes requested):
Looks great, but I have one question.  Should Python be bumped to 3.11 or does it matter in this case?

mdlpstsci (approved):

jhunkeler:
I don't think it matters in this case. For RT runs Python is supposed to be set to the lowest-supported version. The delivery system installs python (not using this file) and executes everything in the `test_cmds` list.

PR #620, opened 2024-08-05
Restore acscteforwardmodel.e hardlink to acscte.e

jhunkeler:
The link wrapper for `acscteforwardmodel.e` was lost when I ported hstcal from WAF to CMake.

The wrapper is used by the following code:
https://github.com/spacetelescope/hstcal/blob/main/pkg/acs/lib/acscte/domaincte.c#L155-L157

My apologies, @mdlpstsci!


mdlpstsci (approved):
@jhunkeler  Looks good.  Thank you for the quick fix. 

Also, I forgot about this executable too (!) - no worries.

PR #622, opened 2024-09-27
MNT: Use hash for Action workflow versions and update if needed

pllim:
As recommended by https://scientific-python.org/specs/spec-0008/#pin-github-actions-release-workflows-to-their-full-release-commit-shas , this PR changes your Actions workflow version pins to hashes, and updates to latest release hashes (at the time of writing) if needed.

This is an automated update made by the `batchpr` tool :robot: - feel free to close if it doesn't look good! You can report issues to @pllim.

[:ghost:](https://github.com/pllim/playpen/issues/60)

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -19,7 +19,7 @@ jobs:
         gcc: [12]
     runs-on: ${{ matrix.os }}
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@d632683dd7b4114ad314bca15554477dd762a938  # v4.2.0
 
     - name: Linux dependencies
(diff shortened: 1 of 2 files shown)


zacharyburnett (approved):
LGTM
 
@mdlpstsci to explain more thoroughly; these hashes are the commits of the most recent tagged releases for these actions. Making them hashes both freezes them to a specific working version, and ensures that a malicious actor couldn't potentially take control of one of these actions' repos and re-tag a tag to point to different code (unlikely, but still technically possible if we leave them pointing to, say `@v4`). Overall, a good change for long-term stability, if we want these workflows to Just Work™️ with minimal maintenance

PR #623, opened 2024-10-04
HSTCAL: Removed a WFC3 regression test

mdlpstsci:
Removed a WFC3 regression test which processes an enormous number of input files.  This test is currently marked as skipped and does not illustrate anything which is not already covered by other tests.


stscirij (approved):
LGTM

Issue #631, opened 2025-03-17
TST: macos-12 job failed


pllim:
Example log: https://github.com/spacetelescope/hstcal/actions/runs/13773057638

pllim:
Two problems:

1. Replace macos-12 with macos-13.
2. But even after that, it still fails. Turns out Ubuntu is pulling in cfitsio 4.3.1 (passed) but brew pulling in cfitsio 4.6.0 (failed). OSX complains about unable to find `<fitsio.h>`.

I cannot figure out how to get brew to install older version of cfitsio (is that even possible)? Even if we can, should we look into why it is failing with cfitsio 4.6.0? It is quite a big jump from 4.3 to 4.6, so I think that is a bigger project that would require more extensive INS testing and so on.

pllim:
https://heasarc.gsfc.nasa.gov/fitsio/

* Mar 2025: CFITSIO Version 4.6.0 -- Released with vulnerability fixes along with other bug fixes and enhancements. (see [what's new](https://heasarc.gsfc.nasa.gov/FTP/software/fitsio/c/docs/changes.txt))
* Aug 2024: CFITSIO Version 4.5.0 -- Conversion of configure/build for automake. Bug fixes and enhancements.
* Jun 2024: CFITSIO Version 4.4.1 -- License update only, no code changes.
* Feb 2024: CFITSIO Version 4.4.0 -- released with bug fixes, enhancements and a reorganization to include additional FITS utility programs.

jhunkeler:
This is a bug on cfitsio's end.

Neither the path to the headers or the libraries are present in the installed `cfitsio.pc` file, so `pkg-config` returns a malformed argument list:

```
$ pkg-config --cflags cfitsio
-I
$ pkg-config --libs cfitsio
-L -lcfitsio
```

The expected output is:

```
$ pkg-config --cflags cfitsio
-I/path/to/include
$ pkg-config --libs cfitsio
-L/path/to/lib -lcfitsio
```

The configuration generated by 4.6.0's build system is:
```
prefix=/path/to
exec_prefix=/path/to
libdir=
includedir=

Name: cfitsio
Description: FITS File Subroutine Library
URL: https://heasarc.gsfc.nasa.gov/fitsio/
Version: 4.6.0
Libs: -L${libdir} -lcfitsio
Libs.private:  -lz -lcurl -lm
Cflags: -I${includedir}
```

[...]

PR #632, opened 2025-03-17
TST: Cancel duplicate workflow runs

pllim:
This is so commit spam won't result in long CI queue.

CI failure unrelated: https://github.com/spacetelescope/hstcal/issues/631


pllim:
This should be really not controversial, so I YOLO this one.

PR #633, opened 2025-03-18
TST: Replace macos-12 with macos-13

pllim:
Fix #631 .

PR #634, opened 2025-03-18
TST: Replace macos-12 with macos-13 and disable fail-fast

pllim:
Non-controversial changes from #633 while investigating #631


pllim:
macos-12 is dead,  so I am gonna YOLO this one too.

Issue #636, opened 2025-04-10
DeprecationWarning: Use shutil.which instead of find_executable


pllim:
Seen in RT test log:

```
tests/helpers.py:42: DeprecationWarning: Use shutil.which instead of find_executable
    cal = find_executable(calxxx)
```

https://github.com/spacetelescope/hstcal/blob/8337430f06743406f13d44a91ef124e76462025d/tests/helpers.py#L4

PR #637, opened 2025-04-10
TST: Replace find_executable with shutil.which

pllim:
Fix https://github.com/spacetelescope/hstcal/issues/636

RT https://github.com/spacetelescope/RegressionTests/actions/runs/14386641621


jhunkeler (approved):
Sure. This is safe.
`shutil.which` was added in Python 3.3.

pllim:
Thanks for the review!

Issue #346, opened 2018-08-28
Standardize command line args for underlying CALACS executables


nmiles2718:
When reviewing changes implemented in `acstools` PR [#46](https://github.com/spacetelescope/acstools/pull/46), I came across an issue when testing some updated functionality that allows for saving of intermediate products in our stand-alone destriping routine, [acs_destripe_plus](https://acstools.readthedocs.io/en/stable/acs_destripe_plus.html). The logic implemented in the updates to `acs_destripe_plus` to allow for the saving of intermediate files is correct (see [L506](https://github.com/spacetelescope/acstools/blob/ee2c124d56e7bc2501a6056d4378f1d16eca2c78/acstools/acs_destripe_plus.py#L506)), the issue is that the preceding to call `acs2d` ([L501](https://github.com/spacetelescope/acstools/blob/ee2c124d56e7bc2501a6056d4378f1d16eca2c78/acstools/acs_destripe_plus.py#L501)) is deleting the reference files. 

As it stands, only the `calacs.e` executable has the command line argument `-s` defined and it is this argument that tells `calacs.e` not to delete the intermediate data products (blv_tmp.fits, blc_tmp.fits). 
```console
(astroconda3) nmiles@nathan:~$acs2d.e 
syntax:  acs2d [--help] [-t] [-v] [-q] [--version] [--gitinfo] input output
(astroconda3) nmiles@nathan:~$acsccd.e
syntax:  acsccd [--help] [-t] [-v] [-q] [--version] [--gitinfo] input output
(astroconda3) nmiles@nathan:~$acscte.e
syntax:  acscte.e [--help] [-t] [-v] [-q] [--version] [--gitinfo] [-1|--nthreads <N>] [--ctegen <1|2>] [--pctetab <path>] input [output]
(astroconda3) nmiles@nathan:~$acssum.e
syntax:  acssum [--help] [-t] [-v] [-q] [--version] [--gitinfo] input [output]
(astroconda3) nmiles@nathan:~$acsrej.e
Syntax error: acsrej.e input output [-t] [-v]
                    [-shadcorr] [-crmask] [-readnoise_only] 
                    [-table <filename>] [-scale #]
                    [-init (med|min)] [-sky (none|mode)]
                    [-sigmas #] [-radius #] [-thresh #]
                    [-pdq #]

input             comma-delimited string (e.g., filename or filename1,filename2,filename3)
output            string
-t                print timestamps
-v                turn on verbose mode
[...]

PR #638, opened 2025-04-10
Fix acsrej argument parser

jhunkeler:
- `input` and `output` are truly positional arguments now
- Adds support for `--version`, `--gitinfo`, and `--help` (although undocumented in the usage statement)
- Handle basic errors more gracefully (tell the user what happened)

```
# Exit non-zero on no arguments
$ ./acsrej.e ; echo $?
Not enough arguments
<usage statement>

2
```

```
# Exit zero on user-requested usage/help statement
$ ./acsrej.e --help ; echo $?
<usage statement>

0
```

```
# We have arguments, but not positional ones (`input` and `output` are required)
# Exit non-zero
$ ./acsrej.e -v ; echo $?
Missing positional argument: input
<usage_statement>

2
```

```
# We have an input, but no output
# Exit non-zero
$ ./acsrej.e -v INPUT_FILE ; echo $?
Missing positional argument: output
<usage statement>

2
```

xref: #346

### DIFF
--- a/pkg/acs/lib/acsrej/rej_command.c
+++ b/pkg/acs/lib/acsrej/rej_command.c
@@ -295,12 +295,16 @@ static int getArgS (char **argv, int argc, int *ctoken, short *value) {
 static int getArgT (char **argv, int argc, int *ctoken, char *value) {
     extern int status;
 
-    if (*ctoken <= (argc-2)) {
-        strcpy (value, argv[++(*ctoken)]);
-        (*ctoken)++;
-        return (status = ACS_OK);
-    } else
-        return (syntax_error (argv[*ctoken]));
+    char *arg = NULL;
+    if ((arg = next_argument(argc, argv, *ctoken)) == NULL) {
+        char msg[255] = {0};
+        snprintf(msg, sizeof(msg) - 1, "%s requires a valid string argument, got '%s'", argv[(*ctoken)], arg);
+        return syntax_error(msg);
+    }
+
+    strcpy(value, arg);
+    (*ctoken)++;
+    return (status = ACS_OK);
 }
 
 /* ------------------------------------------------------------------*/
--- a/pkg/acs/lib/acsrej/rej_command.c
+++ b/pkg/acs/lib/acsrej/rej_command.c
@@ -13,6 +13,7 @@
 
 void rej_reset (clpar *, int []);
 static void usage(char *);
+static int is_valid_filename(char *);
 static int syntax_error (char *);
 static int getArgS (char **, int, int *, short *);
 static int getArgR (char **, int, int *, float *);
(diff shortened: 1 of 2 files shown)


pllim:
https://github.com/spacetelescope/RegressionTests/actions/runs/14390906030

jhunkeler:
@mdlpstsci 
Is there any way we can switch to one of the following well known argument parsers? Both libraries employ the same argument parsing conventions under the hood. The only real caveat is that `-long-option-here` with a single dash prefix should be abandoned in favor of short (`-s`) and long (`--long`) options.

* `getopt` - https://www.gnu.org/software/libc/manual/html_node/Getopt.html
  * Specifically `man 3 getopt_long` 
  * More robust than writing a parser from scratch in every program
  * Easy to implement
  * "Portable" because it exists everywhere, but the implementation might differ slightly between operating systems
* `argp` - https://www.gnu.org/software/libc/manual/html_node/Argp.html
  * More feature rich than `getopt`
  * Automatic usage output
  * Needs to be installed as a dependency on Mac OS: `brew install argp-standalone`

If compatibility with the current argument scheme must be retained, for better or worse, both libraries provide a way to enable "long only" mode that treats `-h`, `-help`, and `--help` all the same way.

From `man 3 getopt_long_only`:
```
getopt_long_only()  is like getopt_long(), but '-' as well as "--" can indicate a long option.
If an option that starts with '-' (not "--") doesn't match a long option, but does match a
short option, it is parsed as a short option instead.
```

From [argp_flags](https://www.gnu.org/software/libc/manual/html_node/Argp-Flags.html#index-ARGP_005fLONG_005fONLY]):
```
ARGP_LONG_ONLY

Use the GNU getopt ‘long-only’ rules for parsing arguments. This allows long-options
to be recognized with only a single ‘-’ (i.e., ‘-help’). This results in a less useful 
interface, and its use is discouraged as it conflicts with the way most GNU programs
work as well as the GNU coding standards.
```

mdlpstsci:
@jhunkeler @pllim 

I personally do not mind changing the parser in general.  I do have a few caveats.  

1. This is now Pey-Lian's "call" as she is now the Jedi master of ACS.
2. I have some concern about your note for **getopt** regarding **differ slightly** - _"Portable" because it exists everywhere, but the implementation might differ slightly between operating systems_.  If this is handled easily, then it does not sound like a showstopper.
3. We have to be careful what the executables return, or more specifically, we have to check with DSB and company with regard to some executables (e.g., calxxx.e), because they may have expectations on the return values.  Since the reentrant executables (non-calxxx.e executables) are not invoked by DSB, then it is probably fine.

Have I gracefully kicked the can to someone else's court?

pllim:
acsrej parser is directly called here also: https://github.com/spacetelescope/acstools/blob/master/acstools/acsrej.py

So whatever change needs to be compatible with that. Anything else that is different, we have to summarize it in a user friendly way and present to ACS team.

Thanks!

jhunkeler:
Continued by #639

mdlpstsci:
@jhunkeler @pllim 
FYI - This is a follow-up to my previous comment.

From DSB (Rob Swaters). "It (the DSB code) checks for the exit code. If non-zero, it concludes calxxx.e has failed."

pllim:
Closing without merge, as discussed "offline." Thanks!

PR #639, opened 2025-04-15
Prototype argp implementation for acsrej.e

jhunkeler:


pllim:
OSX doesn't like this?

```
fatal error: argp.h: No such file or directory
    6 | # include <argp.h>
      |           ^~~~~~~~
compilation terminated.
make[2]: *** [pkg/acs/lib/CMakeFiles/acs.dir/acsrej/rej_command.c.o] Error 1
make[1]: *** [pkg/acs/lib/CMakeFiles/acs.dir/all] Error 2
make: *** [all] Error 2
```

pkg/acs/lib/acsrej/rej_command.c
```diff
@@ -21,238 +23,389 @@ static int getArgT (char **, int, int *, char *);
...
+   04 Apr 25  -  Replace argument parser with argp
 */
 
-int rej_command (int argc, char **argv, char **input, char *output,
-                clpar *par, int newpar[])
-{
+extern int status;
+
+// Arguments that exit early
+#define OPT_GITINFO 1000
```
pllim, line 32: Where did you get the values for these? Are these new codes? Should they be in a header file?

jhunkeler:
I know. I'm not done :sweat_smile:

pllim:
Oh oops sorry, I see that it is not in draft and got pinged to review.

pkg/acs/lib/acsrej/rej_command.c
jhunkeler: The field where they are used can be an ascii character (short option) or an arbitrary value that argp uses to map a long option back to "something". The "something" is the number.

pllim:
ACSREJ isn't one of the focus given budget, so I propose we close this without merge. What do you think?

pllim:
Sorry, this PR got stale and has conflicts now.

PR #640, opened 2025-04-17
Replaces most sprintf with snprintf

jhunkeler:
Born from #639 - As I was replacing the custom argument parsers I uncovered several conditions that lead straight to segfaults due to buffer overflows. Most often they cropped up when the trailer file functions (`trlerror`, `trlwarn`) were attempting to report messages larger than the line buffer. :sweat_smile:

This also adjusts the size of several buffers to accommodate the expected input, otherwise they would likely be truncated at run-time.

### DIFF
--- a/hstio/hstio.c
+++ b/hstio/hstio.c
@@ -427,11 +427,12 @@ static void ioerr(HSTIOError e, IODescPtr x_, int status) {
         IODesc *x;
         char cfitsio_errmsg[81];
         x = (IODesc *)x_;
-        sprintf(&error_msg[strlen(error_msg)],
+        snprintf(&error_msg[strlen(error_msg)],
+                sizeof(error_msg) - strlen(error_msg),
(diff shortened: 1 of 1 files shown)


pllim:
Is this turning into another reunion? 😉

jhunkeler:
Maybe :grimacing:

mdlpstsci:
Let us break this stuff into separate tickets by instrument next time (hope there is no next time)!

mdlpstsci (dismissed):

mdlpstsci:
No.  It is OK.  I have been through the major agony already.  After a while my eyes were goo-goo, so I could have missed something.

jhunkeler:
Okay I won't do it :sweat_smile: ... There are still several `sprintf`s to take care of. 

>  Having a buff or msg is a break from MsgText.

Is it fair to assume all of these one-off buffers should be using `MsgText`? I thought about changing them, but didn't in case there was a good reason for them to exist

jhunkeler:
Superseded by #644

PR #643, opened 2025-04-21
Initial commit of docker-based smoke test

jhunkeler:
* Can it be compiled successfully using system libraries?
* Do the binaries "run"?

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,96 @@
+name: Red Hat
+on:
+  pull_request:
+    paths:
+      - '**.c'
+      - '**.h'
+      - '**/CMakeLists.txt'
+      - '**/cmake'
+  push:
+    branches: [ main ]
+jobs:
+  centos_stream:
+    name:  ${{ matrix.dist }}-${{ matrix.compiler}}-${{ matrix.build_type }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        dist:
+          - stream9
+          - stream10
+        build_type:
+          - Debug
+          - Release
+        compiler:
+          - gcc
+          - clang
+    container: quay.io/centos/centos:${{ matrix.dist }}
+    steps:
+      - name: Install development tools
+        run: |
+          dnf install -y epel-release
+          /usr/bin/crb enable
+          dnf install -y \
+              clang \
+              gcc \
+              g++ \
+              gfortran \
+              cmake \
+              make \
+              cfitsio-devel \
+              file \
+              git
+
+      - name: checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-tags: true
+
+      - name: Fetch tags from upstream repository
+        run: |
+          git fetch --tags https://github.com/spacetelescope/hstcal
+
+      - name: build
+        run: |
+          mkdir build
+          cd build
+          cmake .. \
+              -DCMAKE_INSTALL_PREFIX=/tmp/hstcal \
+              -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+              -DCMAKE_C_COMPILER=${{ matrix.compiler }} \
+              -DCMAKE_C_FLAGS="-Wno-incompatible-pointer-types"
+          make -j$(nproc)
+          make install
+
+      - name: File Types
+        run: |
+          find /tmp/hstcal -type f | sort | xargs file
+
+      - name: Check common CLI
+        run: |
+          option=(
+            --help
+            --version
+            --gitinfo
+          )
+          for x in $(find /tmp/hstcal -name '*.e'); do
+            echo "TEST $x:"
+            for opt in "${option[@]}"; do
+              .ci/bin/check_exec "$opt" -- "$x" || true
+            done
+            echo
+          done
+
+      - name: Dump internal version
+        run: |
+          cd build
+          cat << EOF > check_version.c
+          #include <stdio.h>
+          #include "hstcalversion.h"
+          int main(int argc, char *argv[]) {
+              printGitInfo();
+              return 0;
+          }
+          EOF
+          
+          gcc -o check_version -I../include -I. ../lib/hstcalversion.c check_version.c
+          ./check_version


.github/workflows/docker.yml
```diff
@@ -0,0 +1,96 @@
...
+
+      - name: checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-tags: true
+
+      - name: Fetch tags from upstream repository
+        run: |
+          git config --global --add safe.directory $(pwd) || true
+          git fetch --tags https://github.com/spacetelescope/hstcal
```
pllim, line 46: Didn't you already ask checkout to fetch tags above?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,96 @@
...
+              return 0;
+          }
+          EOF
+          
+          echo "Compiling:"
+          cat -n check_version.c
+          gcc -o check_version -I../include -I. ../lib/hstcalversion.c check_version.c
+          echo
+          echo "Executing:"
+          ./check_version
```
pllim, line 81: Should this just be shipped officially with the package? I feel like generic check version could be useful beyond a smoke test.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,96 @@
...
+on:
+  pull_request:
+  push:
+    branches: [ main ]
+jobs:
+  centos_stream:
+    name:  ${{ matrix.dist }}-${{ matrix.compiler}}-${{ matrix.build_type }}
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
```
pllim, line 11: This annoyingly generates a lot of combos but I guess does not hurt. How do we know when to update this matrix?

pllim:
What is the goal of this Docker workflow? Is this what DMS will be using?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,96 @@
...
+              g++ \
+              gfortran \
+              cmake \
+              make \
+              cfitsio-devel \
+              file \
+              git
+
+      - name: checkout
+        uses: actions/checkout@v4
```
pllim, line 39:
```suggestion
        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
```

.github/workflows/docker.yml
jhunkeler: When this isn't present the version data shows up as `unknown`

.github/workflows/docker.yml
jhunkeler: You caught me. The purpose of this is to generate a bunch of annoying combos. :sweat_smile: 
I'm not opposed to reducing the matrix to just `Release` mode with `gcc`

jhunkeler:
> What is the goal of this Docker workflow? Is this what DMS will be using?

[HLA-1478](https://jira.stsci.edu/browse/HLA-1478)

Yeah, the end game is to build for whatever DMS uses, but also check distros we don't have (Ubuntu, Debian, etc.) to ensure the program can be compiled using only what's provided by the base OS, and its package manager.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,96 @@
...
+              cmake \
+              make \
+              cfitsio-devel \
+              file \
+              git
+
+      - name: checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
+        with:
+          fetch-tags: true
```
pllim, line 41:
```suggestion
          fetch-depth: 0
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,96 @@
...
+
+      - name: Fetch tags from upstream repository
+        run: |
+          git config --global --add safe.directory $(pwd) || true
+          git fetch --tags https://github.com/spacetelescope/hstcal
```
pllim, line 46:
```suggestion
```

pllim (commented):
What about now?

.github/workflows/docker.yml
pllim: Since it is HLA ticket, I'll defer to @mdlpstsci but she might not respond until next week or two.

jhunkeler:
> What about now?

image

pllim:
Though... if we just ship check_version as a proper function with the package, this is a non-issue? 🤔

pllim:
Why don't we ever have a `hstcal.e`? Like `hstcal.e --version` would do what check_version here does, and `--help` would list the actual `calXXX.e` options. And that's it.

jhunkeler:
User forks don't usually have the upstream tags in them. That's why I was trying to force it to grab them from the upstream repo. Looking back at the log though, it only partially worked. The commit hash was shown instead of `unknown` but the tag field was only showing the hash, not the nearest tag.

pllim:
That is very weird because we never have this problem with pytest header reporting version from PRs at astropy.

pllim:
Example PR from fork showing correct version -- https://github.com/astropy/astropy/actions/runs/14560602516/job/40843389234?pr=18023

pllim:
Anyways, sorry that didn't work. Feel free to drop those commits.

jhunkeler:
I don't know. Maybe whatever `checkout` does mangles the HEAD so much that `git describe --long` won't work without intervention (i.e. something `setuptools_scm` knows how to repair).

jhunkeler:
https://github.com/spacetelescope/hstcal/pull/643/commits/32557edd2621010e5ad389c0eb3e2bf88e80a5a9 fixed it. 

The cmake+bash script appears to believe its operating on an unsafe path.

pllim:
Thanks! Tests look like tests. LGTM but I'll let Michele do the approval when she returns.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,142 @@
...
+          - ubuntu
+          - debian
+          - centos
+        build_type:
+          #- Debug
+          - Release
+        compiler:
+          - gcc
+          #- clang
+        include:
```
pllim, line 27: Hmm. I thought I would see a bazillion combos on the CI run but there are only 3 from this workflow. Is that expected?

jhunkeler: Yeah, I noticed this too but had to step out. I'm trying to figure out what's going on now.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,142 @@
...
+        run: |
+          mkdir build
+          cd build
+          cmake .. \
+              -DCMAKE_INSTALL_PREFIX=/tmp/hstcal \
+              -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+              -DCMAKE_C_COMPILER=${{ matrix.compiler }} \
+              -DCMAKE_C_FLAGS="-Wno-incompatible-pointer-types"
+          make -j$(nproc)
+          make install
```
pllim, line 102: Here

.github/workflows/docker.yml
```diff
@@ -0,0 +1,142 @@
...
+          )
+          option=(
+            --help
+            --version
+            --gitinfo
+          )
+          for x in "${program[@]}"; do
+            echo "TEST $x:"
+            for opt in "${option[@]}"; do
+              .ci/bin/check_exec "$opt" -- "$x"
```
pllim, line 139: and here

pllim (commented):
If I understand correctly, only these 2 might cause the run to fail, and the rest are only informational?

.github/workflows/docker.yml
jhunkeler: Yes

.github/workflows/docker.yml
jhunkeler: and yes

jhunkeler:
I want to keep this as simple as possible to avoid ballooning the PR wait time. If it compiles, links, and the binaries don't segfault instantly I think that's good enough for a smoke test. The only way to truly know is to run the tests, but we don't have any that can be run in the wild (yet?).

jhunkeler:
Note to self:
Remove `-Wno-incompatible-pointer-types` after #645 is merged

pllim:
Still LGTM but I'll let @mdlpstsci to also review since this is HLA ticket. Thanks!

pllim (approved):
HLA or not, I am just going to bite the bullet and get this in. I want this for #658. Thanks, all!

mdlpstsci:
No worries.  Looks fine for me - the parts I understand anyway.

PR #647, opened 2025-05-03
Error handler revamp

### LINKED ISSUE #187: Refactor error handling
Currently, errors are handled using the global variable ``status`` - using a global variable for such a thing is not a wise idea.

This does not seem to be true for ``calstis`` - it seems to use local ``status`` variables which functions then assign to and correctly return, i.e.

```c
int status = OK;
if ((status = function_returning_its_own_status()))
   return status;
```
as apposed to:
```c
extern int status;
if (function_returning_its_own_status())
   return status;
```

A function should 'know' what it is returning and it should be the responsibility of only the caller to handle that returned error, i.e. an error should not leak and persist beyond the caller's scope - other than when correctly returned by the caller itself so on up the call stack. Note, in this case, the function does "know" what it is returning, it is returning the error status returned to it by a specific child call.

### Temp Notes:

I'm not yet sure of what IMPHTAB uses.

This is definitely an issue for both ``calacs`` & ``calwf3`` especially ``calacs`` has it has no global declaration of ``status`` - it must be using code from ``calwf3``. 

WIP: add reasons why this is not a good idea, why the current implementation is a mismatch of global use and poor return-and-call-stack-unwinding. Note orphaned errors.

Current initial run [issuexxx-refactor-error-handling](https://github.com/jamienoss/hstcal/tree/issuexxx-refactor-error-handling)

jhunkeler:
All instrument packages provide their own error reporting routine (`pkg/[inst]/whicherror.c:WhichError`) however the code is effectively the same wherever it appears. The only exception is that when `status` was to set to `ACS_OK`, `WFC3_OK`, or `STIS_OK` (all three are defined as `0`) the function returns early.

This PR breaks the error reporting function into two:
1. `hstcal_error_state_populate` - responsible for initializing a `hstcal_error_state` structure
2. `hstcal_error_state_show` - responsible for displaying error information, similar to `WhichError`.

And introduces two macros:
1. `REPORT_ERROR_STATE()` - Effectively replaces `WhichError`, and does not require any arguments
2. `WhichError()` - Is a wrapper. The argument normally accepted by the old `WhichError` is ignored, but allows the existing hstcal to work as-is.

The `hstcal_error_state` structure just keeps all of HSTCAL's errors in one convenient place.

No additional memory is allocated by these functions, so they can be called from anywhere without fear of accidentally introducing leaks. To keep things clean and simple the backtrace feature writes directly to the `stderr` descriptor to avoid calling `malloc` if/when we're already out of RAM.

Example error (with `-DENABLE_BACKTRACE=ON`):

```
CALBEG*** ACSCCD -- Version 10.4.0 (07-May-2024) ***
Begin    03-May-2025 18:31:03 EDT
Input    j8cd02tyq_raw.fits
Output   j8cd02tyq_blv_tmp.fits
Warning    Output file `j8cd02tyq_blv_tmp.fits' already exists.
ERROR:    Couldn't process CCD data
ERROR:    CALACS processing NOT completed for j8cd02011_asn.fits
ERROR:    Invalid temporary file. (status = 1021)
DEBUG:    Location: /home/example/hstcal/pkg/acs/src/mainacs.c:238
DEBUG:    Backtrace:
/home/example/hstcal/cmake-build-debug/lib/libhstcalib.so(hstcal_error_state_populate+0xa3) [0x7ddb5c635987]
/home/example/hstcal/cmake-build-debug/lib/libhstcalib.so(hstcal_error_state_show+0x39) [0x7ddb5c635e8d]
/home/example/hstcal/cmake-build-debug/pkg/acs/calacs.e(main+0x990) [0x5760d5ad3c4b]
/usr/lib/libc.so.6(+0x27488) [0x7ddb5c035488]
/usr/lib/libc.so.6(__libc_start_main+0x8c) [0x7ddb5c03554c]
/home/example/hstcal/cmake-build-debug/pkg/acs/calacs.e(_start+0x25) [0x5760d5ad31c5]
```

Example error (with `-DENABLE_BACKTRACE=OFF` (default)):
```
CALBEG*** ACSCCD -- Version 10.4.0 (07-May-2024) ***
Begin    03-May-2025 18:31:03 EDT
Input    j8cd02tyq_raw.fits
Output   j8cd02tyq_blv_tmp.fits
Warning    Output file `j8cd02tyq_blv_tmp.fits' already exists.
ERROR:    Couldn't process CCD data
ERROR:    CALACS processing NOT completed for j8cd02011_asn.fits
ERROR:    Invalid temporary file. (status = 1021)
DEBUG:    hstcal_error_state_populate called in main() at /home/example/hstcal/pkg/acs/src/mainacs.c:238
```

And in ***mainacs:238*** we see the following:

```c
int main(int argc, char *argv[]) {
    // ...
            /* Added 19 Mar 1999 - provides interpretation of error for user */
            WhichError (status); // <<< line 238
            freeOnExit(&ptrReg);
            exit (ERROR_RETURN);
```

This `DEBUG` message is only visible when `NDEBUG` is not defined, so this follows the same behavioral pattern as all the `assert` statements in the code.

Magic numbers have been mapped to the appropriate  `#define`. For any `status` values that don't map back to a define, I created `UNKNOWN_MAGIC_CODE_nnnn` defines, and placeholder messages for each. These need to be investigated to see what errors actually need to be emitted.

Fix #187


include/hstcalerr.h
```diff
@@ -45,8 +79,16 @@
 #define COPY_NOT_POSSIBLE      1025
 
 #define INVALID_VALUE          1111
-#define GENERIC_ERROR_CODE     1111 //used in STIS
+#define GENERIC_ERROR_CODE     1112 //used in STIS
```
pllim, line 82: I don't know about changing this now... but then again I don't think anyone memorized the error codes by number even after 35 years, so... 🤷‍♀️

lib/hstcalerr.c
pllim, line 1: Since this is new file, might as well convert the tabs to 4 spaces?

include/hstcalerr.h
jhunkeler: Yeah... Hey, at least the status and the messages are unique now.

lib/hstcalerr.c
jhunkeler: Good point. Done.

I turned off the tab-indent highlight in my editor because it was making me lose my mind

include/hstcalerr.h
pllim: Does DMS actually check the actual int value of individual err status? I am just a tiny bit worried this is going to break a status check but maybe not. 🎲

pllim: Since this is STIS, let's ask @stscirij

stscirij: From what I can tell, there are no references to INVALID_VALUE.  The only reference to 1111 is here, in a comment:
`lib/cs1/getkeyinfo1.c:	Change status values 1111 and 1003 to GENERIC_ERROR_CODE.`
There are many places where GENERIC_ERROR_CODE is returned, and many comments stating that a particular error code was replaced by GENERIC_ERROR_CODE (probably a dozen or so different numbers in the 1000-1100 range).  I don't see any references to the value 1112 in the stis code or headers, so I think it should be OK to remap GENERIC_ERROR_CODE to 1112.

pllim: Thanks, Robert! Is it slightly safer than to remap INVALID_VALUE instead?

stscirij: I don't think it matters - unless one of the other instruments uses 1111 or INVALID_VALUE, we don't need it, so the line setting INVALID_VALUE to 1111 could just go and we could keep GENERIC_ERROR_CODE at 1111.  But changing GENERIC_ERROR_CODE to 1112 should also work and allow use of INVALID_VALUE if needed.

jhunkeler: > Does DMS actually check the actual int value of individual err status? I am just a tiny bit worried this is going to break a status check but maybe not. 🎲

I don't think DMS parses the trailer file for status information. As for exit codes, grep says we're only using the following:
- `0`
- `1`
- `ERROR_RETURN` (2)
- `OUT_OF_MEMORY` (111)

There is one instance of `exit(status)` in `cs12.c`, but it only leads to a function that sets `status` to one of the values above.

lib/hstcalerr.c
```diff
@@ -0,0 +1,172 @@
...
+            e.runtime.message = "No chip found.";
+            break;
+        case REF_TOO_SMALL:
+            e.runtime.message = "Reference image is binned more than science image.";
+            break;
+        case INTERNAL_ERROR:
+            e.runtime.message = "Internal error.";
+            break;
+        case INVALID_EXPTIME:
+            e.runtime.message = "Invalid exposure time.";
```
jhunkeler, line 102: Oh, I forgot I made this one up because the message didn't exist. Is this fine, or should be something else?

pllim: LGTM thanks!

pkg/acs/lib/acs2d/do2d.c
```diff
@@ -411,7 +411,7 @@ int Do2D (ACSInfo *acs2d, int extver) {
 	putSingleGroup (acs2d->output, extver, &x, option);
 	if (hstio_err()) {
 		trlerror("Couldn't write imset %d.", extver);
-		return (status = 1001);
+		return (status = INVALID_EXPTIME);
```
mdlpstsci, line 414: EXPSCORR indicates each _individual_ exposure from a CR-SPLIT, repeated sub-exposures, or DITHER observation should generate an FLT/FLC file.  In other words the "exps" is misleading in that it does not indicate _exposure time_, but rather individual exposure calibration (to FLT/FLC).  A failure here indicates "WRITE_FAILED".  Or at least this is close to what is happening, but it is still not clear why.

pllim:
Lots of conflicts now. If you are still interested in pursuing this, please rebase. Otherwise, please close the PR without merge. Thanks!

pllim:
https://github.com/spacetelescope/RegressionTests/actions/runs/32860223545

pllim:
Sorry, man. There are conflicts again.

PR #651, opened 2025-05-06
DOC: Update conda/mamba source install instructions

pllim:
Updated the instructions based on own experience. Turns out I didn't need to touch `PKG_CONFIG_PATH` using micromamba on Linux (unsure about OSX so I left it there). There is also no need to bloat up `$HOME` when you can write to the env dir.

I debated with myself on whether to update the conda instruction to pin Python or not. I always pin it, but I guess that is not really relevant here.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -100,19 +100,19 @@ export LDFLAGS="-Wl,-rpath=$CONDA_PREFIX/lib"
     cmake .. -DCMAKE_INSTALL_PREFIX=$HOME/hstcal
     ```
 
-3. Build
+2. Build
 
     ```
     make
     ```
 
-4. Install
+3. Install
 
     ```
     make install
     ```
 
-5. Add HSTCAL to `$PATH`
+4. Add HSTCAL to `$PATH`
 
     ```bash
     export PATH=$HOME/hstcal/bin:$PATH
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -76,7 +76,6 @@ dnf install cmake cfitsio-devel gcc gcc-gfortran pkgconf-pkg-config
 ```
 conda create -n hstcal -c conda-forge cmake compilers cfitsio pkgconfig python
 conda activate hstcal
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -76,7 +76,6 @@ dnf install cmake cfitsio-devel gcc gcc-gfortran pkgconf-pkg-config
 ```
 conda create -n hstcal -c conda-forge cmake compilers cfitsio pkgconfig python
 conda activate hstcal
-export PKG_CONFIG_PATH="$CONDA_PREFIX/lib/pkgconfig:$PKG_CONFIG_PATH"
```
mdlpstsci, line 79: On my Mac I have often had to explicitly set this path.   I tried to test  this just now by making a new hstcal conda environment, but my upgraded Mac is missing some software so I am unable to execute the "cmake".  Argh.  Please just leave this in the recipe for now in case users have to do this.

mdlpstsci (changes requested):
Aside from adding back the setting for PKG_CONFIG_PATH, theses changes look fine.

INSTALL.md
pllim: This section  is for Linux only. I did not touch `PKG_CONFIG_PATH` instruction for OSX.

INSTALL.md
```diff
@@ -87,6 +86,17 @@ export LDFLAGS="-Wl,-rpath=$CONDA_PREFIX/lib"
     ```
     mkdir _build
     cd _build
+    ```
+
+   If you use Conda/Mamba, you can install it to the env directly:
+
+    ```
+    cmake .. -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX
+    ```
```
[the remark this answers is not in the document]
pllim: Good point. Added!

mdlpstsci (approved):
Looks fine to me now!

pllim:
Thanks for the review!

PR #653, opened 2025-05-08
Common timestamp

jhunkeler:
Moves `TimeStamp` into the core library.

Minor improvements:

- The `rootname` _could_ be `NULL` so now it's only reported when there is something valid
- Safely pads `-` characters to the end of the `message` string


mdlpstsci (dismissed):
Looks good.

pllim:
There is conflict now. Please rebase.

pllim:
RT: https://github.com/spacetelescope/RegressionTests/actions/runs/15474012151

lib/str_util.c
```diff
@@ -1,10 +1,30 @@
-#include <stdbool.h>
 #include <assert.h>
 #include <string.h>
 #include <fitsio.h>
 
 #include "str_util.h"
 
+/**
+ * Fills a buffer with `ch` up to `max_ch` times.
```
pllim, line 8: Why do I see Sphinx doc syntax here? Probably unnecessary?

jhunkeler: 1) Habit
2) It's readable
3) And secretly makes things like this happen

Screenshot_20250605_145044

PR #655, opened 2025-05-08
FLSHCORR: Scale for linear decrease over time

### LINKED ISSUE #627: Scale flash reference file to observation date in CALACS flash correction
In the event of HST budget cuts, we may end up with very little support for ACS/WFC observations in FY26. In that case, we will not take data for post-flash reference files. The flash rate in those files has been decreasing linearly over time, so we'd like to add in a linear scaling factor to the flash correction step in CALACS to appropriately scale the last available post-flash reference file's flash rate to the observation date. 

We will update this issue when we have more details on the linear scaling factor and observation date after which to apply this change.

Note that if ACS/WFC support continues into FY26, we would not implement this change to flash correction. So, development of these changes can proceed, but until we know more about the fate of ACS, we should wait to merge it into CALACS.

pllim:
FLSHCORR: Scale for linear decrease over time if the reference image is the last ever to be delivered.

But we do not want to merge this until we know for sure, though I coded it up such that it would still run without the scaling if those new keywords are not found in the reference image primary header.

Fix https://github.com/spacetelescope/hstcal/issues/627

RT https://github.com/spacetelescope/RegressionTests/actions/runs/14917237097

### TODO

- [x] Run CI and RT, and fix any failures.
- [ ] Wait for ACS Team sign off to merge or not. Rerun CI and RT.
- [ ] Add new test case with prototype reference image to regression test.

### DIFF
--- a/pkg/acs/Dates
+++ b/pkg/acs/Dates
@@ -1,3 +1,6 @@
+ 08-May-2025   CALACS 10.5.0 FLSHCORR now applies scaling to account for linear decrease of
+                             flash rate with time, because no more new reference image will
+                             be delivered going forward.
  07-May-2024   CALACS 10.4.0 Implementation of the "Parallel and Serial CTE correction". This
                              version of the CTE correction (Generation 3) is based upon the
                              algorithm of Generation 2, but includes the additional correction
--- a/pkg/acs/History
+++ b/pkg/acs/History
@@ -1,3 +1,8 @@
+### 08-May-2025 - PLL - Version 10.5.0
+    FLSHCORR now applies scaling to account for linear decrease of
+    flash rate with time, because no more new reference image will
+    be delivered going forward.
+
 ### 07-May-2024 - MDD - Version 10.4.0
     Implementation of the "Parallel and Serial CTE correction". This
     version of the CTE correction (Generation 3) is based upon the
--- a/pkg/acs/Updates
+++ b/pkg/acs/Updates
@@ -1,3 +1,10 @@
+Update for version 10.5.0 - 08-May-2025 (PLL)
+    pkg/acs/Dates
+    pkg/acs/History
+    pkg/acs/Updates
+    pkg/acs/include/acsversion.h
+    pkg/acs/lib/acs2d/doflash.c
+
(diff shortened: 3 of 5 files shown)


pllim:
As of 2025-05-08 implementation, if you give it the existing FLSHFILE from CRDS without these new keywords, it would skip the scaling (the following output is from `-v` option, so you call `calacs.e blah_raw.fits -v`). It would report the keyword is missing but then keep going without the scaling. It would also stop checking at the first missing keyword (3 total):

```
FLSHCORR PERFORM
Trying to open jref$39h2109kj_fls.fits...
Read in Primary header from jref$39h2109kj_fls.fits...
ERROR:    Keyword = `SLOPE'.
Performing post-flash subtraction on chip 1 in imset 2
Linear decrease over time scaling for post-flash: SKIPPED
Image has an offset of 0,0
Mean of post-flash image (MEANFLSH) = 17.0107
FLSHCORR COMPLETE
```

And if you give it the new FLSHFILE, it would be like this instead:

```
FLSHCORR PERFORM
Trying to open test_fls.fits...
Read in Primary header from test_fls.fits...
Performing post-flash subtraction on chip 1 in imset 2
Linear decrease over time scaling for post-flash: 1.026984e+00
Image has an offset of 0,0
Mean of post-flash image (MEANFLSH) = 12.2674
FLSHCORR COMPLETE
```

pllim:
p.s. The scaling factor is greater than 1 probably because I just grabbed a random image from Artifactory to test and it is very old.

mdlpstsci (dismissed):
Code changes look fine.

pllim:
Update 2025-05-21: Jenna gave me a couple of sample inputs and expected outputs. I ran ACS2D from this branch and compared my outputs to hers. There are differences and I am waiting to hear back if they are expected or not.

pkg/acs/Dates
pllim, line 1: https://github.com/spacetelescope/hstcal/pull/685 is likely going in first, so my plan is then to update this PR to append to the "CALACS 10.5.0" entries, given both sets of changes would go into the same delivery anyway, but update the dates here when we are sure we want to merge this one.

PR #656, opened 2025-05-08
MNT: Delete unused release_tools directory

pllim:
I don't think we use this directory anymore?

RT: https://github.com/spacetelescope/RegressionTests/actions/runs/14914921871


jhunkeler:
This hasn't been used since the ureka days.

mdlpstsci (approved):
This has aged to become cruft.

pllim:
Thanks for quick reviews!

PR #650, opened 2025-05-05
DOC: Remove unused ReadTheDocs

pllim:
I stumbled upon RTD setup that would never work in the first place and has no knowledge that this package ever registered RTD space, so having these files around is just very confusing.

I also transferred some contents from that `index.rst` file I am deleting to package README instead.

PR #661, opened 2025-05-20
DOC: Update README

pllim:
Update README with contents from old RTD doc. This is #650 but without gutting RTD setup.


pllim:
Since Michele already verbally approved the text of the README from #650 and this PR does not touch RTD anymore, I am just going to merge without further review. If you catch anything else, I will be happy to open follow-up PR. Thanks, all!

PR #660, opened 2025-05-20
Replace conflicting struct ACSInfo with ACSInfoRef where applicable

jhunkeler:
Continuation of #659, for ACS

PR #662, opened 2025-05-20
Rename ACSInfo to CALACSInfo

jhunkeler:
Another way to solve #660


pllim (approved):
Seems pretty clean, so LGTM as along as RT is green. Thanks!

PR #664, opened 2025-06-05
hstio: replace 'int' with intended type 'long'

jhunkeler:
```
warning: overflow in conversion from ‘long int’ to ‘int’ changes value from ‘-9223372036854775808’ to ‘0’ [-Woverflow]
  105 |      const int minLong = LONG_MIN;
      |                          ^~~~~~~~
warning: overflow in conversion from ‘long int’ to ‘int’ changes value from ‘9223372036854775807’ to ‘-1’ [-Woverflow]
  106 |      const int maxLong = LONG_MAX;
```

### DIFF
--- a/hstio/numeric.c
+++ b/hstio/numeric.c
@@ -102,8 +102,8 @@ extern "C" {
      const int minLong = INT_MIN;
      const int maxLong = INT_MAX;
 # else
-     const int minLong = LONG_MIN;
-     const int maxLong = LONG_MAX;
+     const long minLong = LONG_MIN;
+     const long maxLong = LONG_MAX;
 # endif
 const float minFloat = FLT_MIN;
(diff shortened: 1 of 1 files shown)


pllim (commented):
So `minLong` and `maxLong` could be either `int` or `long` depending on `SIZEOF_INT`. What is `SIZEOF_INT` on the supported systems? In what situation would it go the other way?

pllim:
How did this not trip any of the CI checks?

mdlpstsci (approved):
Ack.

pllim:
Since approved, I will merge when RT and CI pass unless you beat me to it. Thanks, all!

pllim:
Merged. Also `long-story`👏 👏

PR #667, opened 2025-07-02
Build system fixups

jhunkeler:
* Uses CMake to discover Git tags and set the project version. (`dist` and `git_version` scripts have been removed)
* `VERSION`, `BRANCH`, and `COMMIT` defines have been prefixed with `HSTCAL_VERSION` for easier identification
* Fixes `cpack`'s usage of `PROJECT_VERSION`
* The generated `DISTINFO` file is now three lines instead of one


pllim:
Thanks. At a higher level, what problem is this PR fixing? Is there something not working as-is somewhere?

jhunkeler:
The high-level problem is my original implementation wasn't totally portable. This makes it so we only require `cmake` and `git`. Previously if the user's `bash` shell happened to be a symlink to another shell (i.e. `dash` on ubuntu is a good example) there's always a chance it could fail or act differently in subtle ways.

pllim:
Not sure how to test this except just make a release and see what happens lol. Ping me again when you take this out of draft. At a glance, LGTM (pending RT run).

.github/workflows/docker.yml
```diff
@@ -112,7 +112,7 @@ jobs:
       - name: Dump --gitinfo header
         run: |
           cd build
-          cat -n version.h
+          cat -n config.h
```
pllim, line 115: Why rename this?

pllim: Oh nvm, I see `config.h.in` diff below.

pllim (approved):

Issue #669, opened 2025-07-03
Why are we using raw assert?


pllim:
https://github.com/search?q=repo%3Aspacetelescope%2Fhstcal%20assert&type=code

Example:

https://github.com/spacetelescope/hstcal/blob/85c4cf2b6ac7cf94d03bd53c0b27370b31996b27/pkg/acs/lib/acsccd/blev_funcs_postsm4.c#L61-L64

Shouldn't we log the proper error code and exit gracefully like this?

https://github.com/spacetelescope/hstcal/blob/85c4cf2b6ac7cf94d03bd53c0b27370b31996b27/pkg/acs/lib/acsccd/blev_funcs_postsm4.c#L84

We don't want the pipeline to panic and abort, right?

Ref: https://en.cppreference.com/w/cpp/error/assert.html

mdlpstsci:
@pllim 
You are quite correct in that a raw _assert_ should **not** be used in any pipeline code!  Any bug/failure/exception should be handled with an informative message and a graceful exit.

pllim:
The fact the pipeline never panics here (I think we would have heard from DMS if it did) probably means this check is unnecessary. We do high level checks further up in `doxxx.c`, no? And some logic is set by flags that are out of our control. So basically by the time you get here, you get what you ask for (i.e., garbage in, garbage out). So here are 2 options:

1. Remove the checks entirely but leave the comment for future with implicit assumptions.
2. Change the checks to proper `trlerror` stuff, though this might duplicate checks already made elsewhere higher up the chain.

Do you have a preference, @mdlpstsci ?

mdlpstsci:
@pllim 
In truth these checks definitely seem redundant.  If the processing has gotten this far, the checks seem entirely unnecessary.  I would consider just removing them entirely.

PR #670, opened 2025-07-08
MNT: Do not use assert in non-test code

pllim:
Fix https://github.com/spacetelescope/hstcal/issues/669


pllim:
RT: https://github.com/spacetelescope/RegressionTests/actions/runs/16149843206

I don't think Docker failure is related.

mdlpstsci (dismissed):
I do not have a problem with these changes.

hstio/hstio.c
```diff
@@ -1074,7 +1071,10 @@ void copyOffsetSingleGroup(SingleGroup * output, const SingleGroup * input,
     if (!output || !input)
         return;
     //WARNING - assumes row major storage
-    assert(output->sci.data.storageOrder == ROWMAJOR && output->sci.data.storageOrder == ROWMAJOR);
+    if (output->sci.data.storageOrder != ROWMAJOR ||
+            input->sci.data.storageOrder != ROWMAJOR) {
+        return;
```
jhunkeler, line 1076: The `assert` was grave enough end the program `.storageOrder` wasn't `ROWMAJOR`. We may want to consider setting `status` before returning.

lib/hstcal_memory.c
```diff
@@ -16,12 +15,11 @@ void initPtrRegister(PtrRegister * reg)
     reg->cursor = 0; //points to last ptr NOT next slot
     reg->length = PTR_REGISTER_LENGTH_INC+1; //+1 to special case reg->ptrs[0] for 'this' pointer only
     reg->ptrs = malloc(reg->length*sizeof(*reg->ptrs));
-    assert(reg->ptrs);
     reg->freeFunctions = malloc(reg->length*sizeof(*reg->freeFunctions));
     if (!reg->freeFunctions)
     {
         free(reg->ptrs);
-        assert(0);
+        return;
```
pllim, line 22: @jhunkeler , in regards to https://github.com/spacetelescope/hstcal/pull/670/files#r2195829235 here, there is no status set at all anywhere in the file and I don't feel like introducing it now. Is a plain `return;` here okay or should I do something else? Hope you can advise.

Same comment for all similar occurrences in this file.

hstio/hstio.c
pllim: OK I set the status on L1076. But everywhere else in this file, the function declare local `int status = 0;` and then either return the status or pass it by address into something else. Can you please double check that my addition here is okay? Thanks!

pllim:
RT again: https://github.com/spacetelescope/RegressionTests/actions/runs/16183962473

hstio/hstio.c
mdlpstsci: I refreshed my memory a bit.  It looks like the _assert_ business started when JN implemented the CTE Version 2 - **argh**.  **Even more argh** when I did not recode the asserts when I updated to CTE Version 3.

It looks like the status will bubble up the chain and finally deliver the status (doing this quickly so may not be exactly this chain):
dopcte-gen3.c
           copyOffsetSingleGroup
         extractAmp
       doPCTEGen3
docte.c
        DoCTE
acscte.c
        ACScte
calacs.c
        ProcessACSCCD
     CalAcsRun

Please just force (hardcode for the test) the status to failure, process an image, and see what pops to make sure all is well.

Issue #668, opened 2025-07-02
RFC: Move bias shift code to its own file


pllim:
@mdlpstsci said

https://github.com/spacetelescope/hstcal/blob/85c4cf2b6ac7cf94d03bd53c0b27370b31996b27/pkg/acs/lib/acsccd/doccd.c#L66

and

https://github.com/spacetelescope/hstcal/blob/85c4cf2b6ac7cf94d03bd53c0b27370b31996b27/pkg/acs/lib/acsccd/doccd.c#L367

should be moved out to its own file, so `doccd.c` can be clean like this

https://github.com/spacetelescope/hstcal/blob/85c4cf2b6ac7cf94d03bd53c0b27370b31996b27/pkg/acs/lib/acsccd/doccd.c#L745-L749

pllim:
@mdlpstsci , before I can proceed, I need to understand the current layout, so hope you can advise after you return from the holiday weekend:

* Why is DARKTIME computation in `doccd.c`? I thought DARKCORR happens in ACS2D step, not ACSCCD. I ask because it also uses the `subApertures` stuff you wanted me to move out of the file.
* As for the bias shift code you want me to move, is it cleaner if we move it to one of the existing C files instead of creating a new one? Which one? As it is, there is a bunch already and things are getting confusing.
* If you insist on a new C file, any suggestion on what to name it?

```
$ ls -1 pkg/acs/lib/acsccd/
acsccd.c
blevdrift.c
blevfit.c
blev_funcs_postsm4.c
doatod.c
dobias.c
doblev.c
doccd.c
dofwsat.c
dosink.c
findblev.c
findover.c
getacsflag.c
getccdsw.c
```

Thanks!

Ref: https://hst-docs.stsci.edu/acsdhb/chapter-3-acs-calibration-pipeline/3-3-structure-of-calacs

mdlpstsci:
@pllim 
In the History (Version 10.3.3) I see "Code written for CALACS 10.2.2 was moved from ACS2D to ACSCCD so every BLV_TMP file has the correct DARKTIME keyword value."

I see the header comment in doccd.c:
24-May-2021 MDD: Read the post-flashed and unflashed columns from the CCDTAB
     reference file to use for the offset to the DARKTIME FITS keyword value
     under appropropriate date and supported subarray criteria.  The DARKTIME
     keyword value is now the default scaling factor for the superdarks, with
     the offset being an additive correction to DARKTIME under appropriate
[...]

Issue #329, opened 2018-08-01
Extend the bias shift corretion algorithm of CALACS to handle the WFC 512 and 1K subarrays


mdlpstsci:
This is a place-holder issue to extend the bias shift correction in the BLEVCORR step of the ACSCCD algorithm of CALACS to handle the WFC 512 and 1K subarrays.  Currently, the algorithm depends upon the data having both physical and virtual overscan.  The virtual overscan is not present in the 512 and 1K subarray data so the team needs to define an alternative value to be used for the algorithm.

jryon:
Can someone confirm if the bias shift correction is not currently enabled for 512 and 1K subarray data? The ACS DHB says it is not currently enabled, but the fact that #312 is completed makes me second guess it all.

mdlpstsci:
The only data supported is the following:

"WFC1A-2K", "WFC1B-2K", "WFC2C-2K", "WFC2D-2K",
"WFC1-IRAMPQ", "WFC1-MRAMPQ", "WFC2-MRAMPQ", "WFC2-ORAMPQ",
"WFC1-POL0UV", "WFC1-POL0V", "WFC2-POL0UV", "WFC2-POL0V",
"WFC1-POL60UV", "WFC1-POL60V", "WFC2-POL60UV", "WFC2-POL60V",
"WFC1-POL120UV", "WFC1-POL120V", "WFC2-POL120UV", "WFC2-POL120V"

pllim:
@mdlpstsci , is that answer from the "keyword rules" file you sent to me and Robert this morning? Thanks.

mdlpstsci:
@jryon @pllim 
The response is from the doccd.c source code.  Near the top of the routine is an array of subApertures.  There is a large comment above the variable with some description.  Also, at Line 390 there is more description.

This is the routine that I want to re-arrange which we can discuss at another time.

pllim:
Ah, I see. Thanks!

https://github.com/spacetelescope/hstcal/blob/1cecb633c5f39b3faf7b653167ae4c82e44ec092/pkg/acs/lib/acsccd/doccd.c#L66-L70

https://github.com/spacetelescope/hstcal/blob/1cecb633c5f39b3faf7b653167ae4c82e44ec092/pkg/acs/lib/acsccd/doccd.c#L387-L427

jryon:
[...]

PR #674, opened 2025-07-10
RFC: Move bias shift code out of doccd.c (try 2)

pllim:
Fix https://github.com/spacetelescope/hstcal/issues/668 . I don't think it needs a change log because if done properly, user should see no difference. 🤞 

RT: https://github.com/spacetelescope/RegressionTests/actions/runs/16435883013/job/46445714822

This is https://github.com/spacetelescope/hstcal/pull/673 with branch name that won't crash regtest Sonar scan.

### TODO

- [x] Move shared variables out.
- [x] Move bias shift to a new file called `dobiasshift.c`. Update cmake manifest(s).
- [x] Make sure CI and regression tests are clean.
- ???
- Profit!!!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,14 +10,12 @@ _CPack_Packages/
 Testing
 Makefile
 cmake_install.cmake
+cmake-build-debug/
 install_manifest.txt
 compile_commands.json
 CTestTestfile.cmake
-.lock-waf*
-.waf*
 bin.*
 *build
-build
 build.*
 Makefile
 __pycache__
@@ -26,3 +24,4 @@ __pycache__
 *.a
 *.so
 hstcal-*.sh
+.idea/
--- a/.gitignore
+++ b/.gitignore
@@ -10,7 +10,7 @@ _CPack_Packages/
 Testing
 Makefile
 cmake_install.cmake
-cmake-build-debug/
+cmake-build-*
 install_manifest.txt
 compile_commands.json
 CTestTestfile.cmake
--- a/pkg/acs/lib/acsccd/dobiasshift.c
+++ b/pkg/acs/lib/acsccd/dobiasshift.c
@@ -18,6 +18,9 @@
    this data, the dimensions of the image also have to be checked to determine
    if there are both physical and virtual overscan sections so the bias shift
    correction can be applied.
+
+   23-Jul-2025 PLL: Created isValidBiasShiftSubArrWithVirt() in order to
+       move code out of doccd.c to simplify the complexity of doccd.c
 */
 int isValidBiasShiftSubArrWithVirtOscn(int is_subarray, char *aperture, int virt_oscn_value) {
     if (is_subarray != YES)
--- a/pkg/acs/lib/acsccd/dobiasshift.c
+++ b/pkg/acs/lib/acsccd/dobiasshift.c
@@ -1,0 +1,68 @@
+#include <string.h>
+#include "acs.h"
+#include "acsinfo.h"
+
+
+/* If this is subarray data, determine if it is supported for the bias shift
+   correction.  Compare the aperture name against the array of supported
+   apertures.  Use the aperture name and not the date as the discriminant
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -10,14 +10,12 @@ _CPack_Packages/
 Testing
 Makefile
 cmake_install.cmake
+cmake-build-debug/
```
jhunkeler, line 13: I suggest making this `cmake-build-*` because the suffix changes with the release mode and toolchain (when you have more than one defined)

pkg/acs/lib/acsccd/doccd.c
```diff
@@ -437,37 +364,15 @@ int DoCCD (ACSInfo *acs_info) {
...
-                       freeOnExit (&ptrReg);
-                       return (status);
-                    }
-
-                    trlmessage("     mean of bias levels subtracted was %.6g electrons.", meanblev);
-
-                 } else {
-                       trlmessage("Default bias level from CCDTAB was subtracted.");
+                 if (!done) {
+                     trlmessage("Default bias level from CCDTAB was subtracted.");
```
mdlpstsci, line 372: I am a little fuzzy here.  Why did the **_if_** statement change?

pllim: I moved the MEANBLEV calculation to `doblev.c`. I don't see why it has to be repeated 3x in `doccd.c`. However, the "Default bias level from CCDTAB was subtracted." message logic applies outside of `doblev.c` and there is a comment that `done` controls whether the value is from CCDTAB or not. Does this explain my changes?

pkg/acs/lib/acsccd/doccd.c
```diff
@@ -593,36 +500,15 @@ int DoCCD (ACSInfo *acs_info) {
...
-
-                          if (PutKeyFlt(&x[i].sci.hdr, "MEANBLEV", meanblev,
-                                         "mean of bias levels subtracted in electrons")) {
-                             freeOnExit (&ptrReg);
-                             return (status);
-                          }
-
-                          trlmessage("     mean of bias levels subtracted was %.6g electrons.", meanblev);
-                       } else {
+                       if (!done) {
```
mdlpstsci, line 507: Same comment here as for Line 371 regarding change of **_if_** statement.

mdlpstsci (changes requested):
Changes thus far look make the code much more succinct (_doccd.c_)!  Although the changes should not impact any output, and therefore, Dates/History/Updates have not been updated, please add name/date/two_line_summary at the top of _doccd.c_ regarding the cleanup.  I would also add the same to the top of the new _dobiasshift.c_ and _dodarktime.c_.

I do have one comment, and it is **only a comment** as CALACS is now your domain.  When I had intended to move the bias shift code out of _doccd.c_, I was thinking of moving (in your new version) Lines 318-560 to a new file in order to have _doccd.c_ be more streamlined.  Comment?

pkg/acs/lib/acsccd/dobiasshift.c
pllim, line 1: I will add more stuff to this file in #329 , I think.

pllim:
@mdlpstsci 

> I was thinking of moving (in your new version) Lines 318-560 to a new file

Not exactly those lines but I moved a bulk of the logic out. Does it look good enough now? RT is green but I will do local testing just in case after you are okay with the changes.

> please add name/date/two_line_summary at the top of doccd.c regarding the cleanup. I would also add the same to the top of the new dobiasshift.c and dodarktime.c.

Sure, I can do that to all touched C files after we are sure the changes are good. Otherwise, I risk of moving something back out but accidentally leaving the comment in, causing confusion for Future Self ™️ .

pllim:
I ran regtests locally for the version on `main` and the version from this PR. I compared the trailer files of WFC tests and the logs were identical; therefore hopefully this means moving the code this way did not change any functionality of the pipeline.

pllim:
I also added the requested history blurbs.

@mdlpstsci , can you please re-review? Thanks!

PR #677, opened 2025-07-23
TST: Delete empty conftest and unnecessary init files

pllim:
I don't think we need these.

RT: https://github.com/spacetelescope/RegressionTests/actions/runs/16474314149


pllim:
OK nvm, relative import...

PR #687, opened 2025-09-19
Revert "Bias shift correction for WFC 512 and 1K subarrays"

pllim:
Reverts spacetelescope/hstcal#685 upon request from ACS Team because they found weird things that need further investigation.


pllim:
RT is at https://github.com/spacetelescope/RegressionTests/actions/runs/17866644626/attempts/1#summary-50810521690 and 2 failures are because of the reversion here. I can update the "truths" after merge if they look okay with ACS Team.

PR #697, opened 2026-01-09
Add missing define for _GNU_SOURCE

jhunkeler:
Fix the build for people using compilers that don't already have `_GNU_SOURCE` defined.


pllim (approved):
Thanks! cc @starivera

PR #699, opened 2026-02-03
Wavecorr e1 implementation

stscirij:
Implement improved wavecal calibration for CCD data taken at the E1 or E2 aperture locations.  The part of the long slit wavecal spectrum that matches the location of the source is used to calculate the SHIFTA1 value.

### DIFF
--- a/pkg/stis/lib/cs4/spatialshift.c
+++ b/pkg/stis/lib/cs4/spatialshift.c
@@ -140,69 +140,69 @@ double *shift       o: the shift, in pixels
 	    /* nwl is the full width of the image, not just ilast-ifirst+1,
 	       for convenience in indexing.
 	    */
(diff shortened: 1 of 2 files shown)


pllim:
@stscirij can you please rebase to pick up WFC3 changes so you won't see WFC3 failures in your run? Thanks.

stscirij:
@jlothringer @matimet @dstapleton22 Can one of you approve this PR please?  Thank you!

stscirij:
> Do we need to add new test(s)?

Yes, the reason the STIS runs still pass is that neither of the two datasets exercise the new functionality.  It's probably worth adding a test, but maybe for the next build, I wouldn't want to hold this build up for adding a test that's guaranteed to pass this build

Issue #700, opened 2026-02-05
Enable serial CTE correction on WFC subarrays


jryon:
We did not enable serial CTE correction on WFC subarrays. One reason was Michele found that rotating non-square arrays was hard to implement in C: "It turns out that rotating rectangular matrices (M x N where M != N) is complex to do and involves a trade-off between memory and speed.  I have been researching this, and it looks like I should use some library function (e.g., linear algebra package) to do the work.  Square matrices are sooooo easy in comparison." In a later email she mentions having added some code to do the rotation that she would then remove, but I don't know if it was complete, or committed at any point.

Another reason had to do with expectations for what CALACS should do with subarrays, given that the keyword rules currently set PCTECORR = OMIT for subarrays. The reason for that is because they cannot have bias-striping correction run on them in the way CALACS does for full-frames. That is why acs_destripe_plus exists. However, when PCTECORR is set to PERFORM for a supported subarray, then parallel CTE correction can be run on it by acs_destripe_plus.

So, I think if we can figure out how to rotate subarrays to perform the serial CTE correction within CALACS, then we should extend it to all supported subarrays.

We also added a check to skip serial CTE correction for all subarrays. The discussion on that in the old issue starts here: https://github.com/spacetelescope/hstcal/issues/577#issuecomment-2275811763  

xref

* https://github.com/spacetelescope/hstcal/pull/609

pllim:
Fixed in #709

PR #707, opened 2026-02-19
TST: Add tests for ACSCTE rotate amp functions

pllim:
This is a pre-cursor for #700 . Does not affect functionality.

`cmake .. -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DTESTS=ON`

or

`cmake .. -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DENABLE_ASAN=ON -DTESTS=ON`

## TODO 

- [x] Generalize test code to support different cases.
- [x] Implement a square test and see if that pass.
- [x] If square works but not rectangle, mark it as xfail for now (`WILL_FAIL`, see https://cmake.org/cmake/help/latest/manual/cmake-properties.7.html#test-properties)
- [x] Run RT: https://github.com/spacetelescope/RegressionTests/actions/runs/22317755490

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -57,6 +57,13 @@ GCC may be supplemented by Clang under the following conditions:
     cmake .. -DENABLE_OPENMP=OFF
     ```
 
+### C unit tests
+
+- If you want to be able to run unit tests written in C.
+
+    ```
+    cmake .. -DTESTS=ON
+    ```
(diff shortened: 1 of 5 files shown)


pllim:
OK I think this is good for now. The next PR will refactor the rotation stuff.

PR #709, opened 2026-03-16
ACSCTE: Serial CTE correction for subarrays

pllim:
Follow-up of #707 . Towards #700

`cmake .. -DCMAKE_INSTALL_PREFIX=$CONDA_PREFIX -DTESTS=ON`

### TODO

- [x] Test transpose for rectangle
- [x] ~Test flips for rectangle~ (flip is not the problem and tested indirectly)
- [x] Figure out why `rectangle_test_case_derot` is behaving not as expected initially
    - RT: https://github.com/spacetelescope/RegressionTests/actions/runs/23168884162 (passed)
-  Option A:
    - [x] See if transpose fix is enough to expand subarray support with the dataset Jenna provided 
    - RT: https://github.com/spacetelescope/RegressionTests/actions/runs/23175450071
    - [x] Fix serial logic for subarrays; something isn't right. Awaiting good data from Jenna for comparison.
    - RT: https://github.com/spacetelescope/RegressionTests/actions/runs/23266217079
    - Rebased RT: https://github.com/spacetelescope/RegressionTests/actions/runs/25767151533
- ~Option B:~
    - ~Write new function to flip amp data only once for use of both serial and parallel (originally I was very tempted to do this but now I feel like this cost does not justify the lack of benefit, maybe we defer to future if not doing this is somehow a bottleneck in processing but I don't think so because adding serial only adds a second to run time somehow)~
    - ~Retire any unused rotation functions after new function is in effect~
- [x] Make sure changes do not affect existing pipeline functionality besides affected subarrays

### DIFF
--- a/pkg/acs/lib/acscte/dopcte-gen3.c
+++ b/pkg/acs/lib/acscte/dopcte-gen3.c
@@ -126,6 +126,8 @@ int doPCTEGen3 (ACSInfo *acs, CTEParamsFast * ctePars, SingleGroup * chipImage, 
             freeOnExit(&ptrReg);
             return (status);
         }
+        nColumns = ampImage.sci.data.nx;
+        nRows = ampImage.sci.data.ny;
     }
 
     if ((status = alignAmp(&ampImage, ampID)))
@@ -297,6 +299,10 @@ int doPCTEGen3 (ACSInfo *acs, CTEParamsFast * ctePars, SingleGroup * chipImage, 
             freeOnExit(&ptrReg);
             return (status);
         }
+        /* UNCOMMENT IF NEEDED IN THE FUTURE.
+        nColumns = ampImage.sci.data.nx;
+        nRows = ampImage.sci.data.ny;
+        */
     }
 
     if ((status = insertAmp(chipImage, &ampImage, ampID, ctePars)))
--- a/pkg/acs/lib/acscte/dopcte-gen3.c
+++ b/pkg/acs/lib/acscte/dopcte-gen3.c
@@ -411,7 +411,6 @@ int rotateAmpData_acscte(FloatTwoDArray * amp, const unsigned ampID)
        Rotate the amp to put the serial trails in the same orientation
        as the parallel trails would be. A rotation requires a transpose
        and then a flip. Always transpose the data first.
-
     */
     transpose(amp);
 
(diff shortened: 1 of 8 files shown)


pkg/acs/lib/acscte/dopcte-gen3.c
```diff
@@ -461,19 +460,36 @@ int derotateAmpData_acscte(FloatTwoDArray * amp, const unsigned ampID)
...
+void transpose(FloatTwoDArray * amp)
 {
-    const unsigned nRows = amp->ny;
-    const unsigned nColumns = amp->nx;
-    float temp;
+    FloatTwoDArray orig_amp;
+    const int nRows = amp->ny, nColumns = amp->nx;
+    int i, j;
+
+    initFloatData(&orig_amp);
```
pllim, line 474: This generalizes the transpose to support rectangle at the expense of more memory and computation. Is there a smarter way to do this, @jhunkeler or @stscirij ?

pkg/acs/lib/acscte/dopcte-gen3.c
```diff
@@ -127,6 +126,8 @@ int doPCTEGen3 (ACSInfo *acs, CTEParamsFast * ctePars, SingleGroup * chipImage,
             freeOnExit(&ptrReg);
             return (status);
         }
+        nColumns = ampImage.sci.data.nx;
```
pllim, line 129: I hope this is a safe assumption. BLEVCORR would have already run by now, so hopefully not any weird binning or compression we have to worry about?

jryon:
@pllim processed subarray data of different dimensions and amplifiers to test the new implementation of serial CTE correction for subarrays ('new' data). There were two datasets: 1) internal darks using 512 subarrays with all four amps, and 2) external images of stellar fields and a single standard star using 1K and 2K subarrays with only amp B. The same raws were separately processed with `acstools.acs_destripe_plus` with CALACS v10.40, which only applies parallel CTE correction to subarrays ('current' data). 

Difference images (new - current) of dataset 1 were created and visual inspection confirmed that charge was moved in the correct direction given the subarray readout mode, and the magnitude of the corrected trails from hot/warm pixels and CRs seemed reasonable. The same was true of difference images created of dataset 2, for trails from stars. Horizontal stripes in the difference images are expected because bias striping is corrected in the current data, but not in the new data.

Example for dataset 2 (jfld01pbq):
image

For dataset 2, aperture photometry of the stars was performed, and magnitude differences between the new data (with serial) and the current data (without serial) were compared as a function of number of X transfers. The magnitude differences "increase" (stars get brighter) as the number of X transfers increases, as expected for serial CTE correction. The maximum mag difference, furthest from the amplifier, is around -0.005 mag, which is similar to the findings in ACS ISR 2024-07. The data with an individual standard star show expected magnitude differences for the number of X transfers experienced by the star.

Example stellar magnitude differences as a function of X transfers (jfld01pbq):
image

To summarize, these differences are in line with expectations for serial CTE correction, so I'd say this implementation is successful. Once this has been merged, we will need to confirm it is working as expected in `acstools.acs_destripe_plus`

pllim:
Merging and updating regression test output. Thanks

PR #711, opened 2026-04-02
Bump CALACS to 10.4.1 and add change logs

pllim:
This follows up on https://github.com/spacetelescope/hstcal/pull/687 that kept some fixes from #685 instead of full reversion. Changes are that not visible to users (e.g., fixing broken link in code comment, etc) are excluded from change logs.

RT: https://github.com/spacetelescope/RegressionTests/actions/runs/23919800504


pllim:
debian-testing is being difficult with its `/usr/bin/git`. I am pretty sure it is unrelated, so we can ignore the Docker Test failures for now.

s-goldman (approved):

PR #720, opened 2026-08-19
BUG: IMA no crash when ASN is NONE

### LINKED ISSUE #719: BUG: IMA reprocessing with ASN table NONE crashes calwf3
WFC3 Team has a notebook that tells people how to manually reprocess IR data. It goes something like this:

1. Run calwf3 on RAW to make IMA and FLT.
2. Throw away the FLT.
3. Fiddle with IMA *CORR flags.
4. Pass the IMA as input to calwf3 now to make final FLT.

Problem: Step 4 crashes if IMA has ASN table set to NONE.

Desired outcome: calwf3 run to completion and make FLT. ASN table value stay as NONE.

Out of scope: PARTIAL ASN reprocessing case.

pllim:
Fix #719

RT: https://github.com/spacetelescope/RegressionTests/actions/runs/32797847497

### TODO

- [x] Add regression test.
- [x] Add change log.
- [x] Wait for XCTE and crosstalk PRs to go in first.

### DIFF
--- a/pkg/wfc3/lib/calwf3/wf3table.c
+++ b/pkg/wfc3/lib/calwf3/wf3table.c
@@ -1,5 +1,5 @@
 /* These routines provide the interface to the WF3 association tables.
- **	
+ **
  **	Howard Bushouse, 2000 Aug 23
  **		Initial version.
  **		[Adapted from acstable.c by W. Hack]
@@ -25,7 +25,7 @@
  **		Modified GetAsnTable to turn off CRCORR/RPTCORR processing if
  **		there aren't any sub-products with > 1 member. (PR 66366)
  **  M. Sosey 2015
- **      Updates for CTE calibration code, see #1193 
+ **      Updates for CTE calibration code, see #1193
  */
 
 # include <ctype.h>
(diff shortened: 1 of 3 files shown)


pkg/wfc3/lib/calwf3/wf3table.c
```diff
@@ -1354,6 +1350,10 @@ int GetAsnName (char *filename, char *asn_name) {
         return (status = KEYWORD_MISSING);
     }
 
+    if (strncmp(asn_name, "NONE", 4) == 0) {
+        strncpy(asn_name, filename, strlen(filename));
```
pllim, line 1354: Real fix is here.

starivera (approved):
Tested the fix using the HST WFC3/IR up-the-ramp fitting notebook and verified that it enables calwf3 to process an IMA FITS file with the primary header keyword ASN_TAB set to NONE to completion as a single dataset, without requiring a manual update of the ASN_TAB keyword. I compared the resulting FITS products against those produced by the current calwf3 after manually setting ASN_TAB and found only the expected differences in the ASN_TAB header value. Trailer file content is also as expected.

Issue #35, opened 2016-11-29
Address the many compilation warnings


jamienoss:
When compiling hstcal (at least with clang) there exists a superfluous number of warnings that cloud the build output and make it difficult and tedious to locate those possibly introduced during development.

Making this issue the parent issue to the various sub-issues, addressing the warnings on an individual basis.

pllim:
Did you ever install that `gcc` with @jhunkeler ? I only see 2 warnings compiling with `--debug` with `gcc`.

sosey:
hstcal is currently being built with gcc for both mac and linux and distributed as such to enable the openmp calls

jamienoss:
@pllim Yes, we did.
@sosey I don't see how this stops the cause of these warnings from existing within the code?

sosey:
@jamienoss It doesn't stop the cause of the warnings existing in the code. I'm reminding everyone that joe implemented the gcc build with 2016.3 to accommodate wfc3 and we are not currently building or distributing a clang build. If you want to fix the clang build errors I'm fine with that, just know that building wfc3 on clang won't produce a fully functional executable.

jamienoss:
@sosey copy that.

PR #724, opened 2026-08-25
MNT: Exorcise compilation warnings in acs2d

pllim:
A small step towards #35 

RT: https://github.com/spacetelescope/RegressionTests/actions/runs/32889977628

The compilation log still has over 2k lines after this fix. 😭

### DIFF
--- a/pkg/acs/lib/acs2d/dospot.c
+++ b/pkg/acs/lib/acs2d/dospot.c
@@ -6,37 +6,33 @@
 # define    PIX(v,i,j,nx)   v[(i) + (j) * (nx)]
 
 int readSpotImage(char *spotname, SingleGroup *inspot, SingleGroupLine *spotline) {
-    
+
     extern int status;
-    int dimx;
-	
+    const int dimx = inspot->sci.data.nx;
+
     initSingleGroup (inspot);
 	/* Open the input image. */
 	getSingleGroup (spotname, 1, inspot);
 	if (hstio_err())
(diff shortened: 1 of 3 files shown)


pkg/acs/lib/acs2d/dospot.c
```diff
@@ -94,47 +89,39 @@ float binterp (float x, float y, int nxpix, int nypix, SingleGroup *img) {
...
+    }
 
-    value = tx * ty * Pix(img->sci.data,nx,ny) + sx * ty * hold21 + 
+    value = tx * ty * Pix(img->sci.data,nx,ny) + sx * ty * hold21 +
             sy * tx * hold12 + sx * sy * hold22 ;
 
     return (value);
 }
-
-float ninterp (float x, float y, int nxpix, int nypix, SingleGroup *img) {
```
pllim, line 134: Compiler was complaining about unused arguments but this whole function is unused, so just delete the whole thing.

jhunkeler:
```
$ grep -Hnr ninterp
pkg/acs/lib/acs2d/dospot.c:134:float ninterp (float x, float y, int nxpix, int nypix, SingleGroup *img) {
grep: cmake-build-debug/pkg/acs/lib/libhstacs.so: binary file matches
grep: cmake-build-debug/pkg/acs/lib/CMakeFiles/acs.dir/acs2d/dospot.c.o: binary file matches
```

Agreed, this isn't used at all

PR #728, opened 2026-08-29
MNT: Exorcise compilation warnings in acsrej, acssum, and calacs

pllim:
Sequel of https://github.com/spacetelescope/hstcal/pull/726

A small step towards https://github.com/spacetelescope/hstcal/issues/35

RT: https://github.com/spacetelescope/RegressionTests/actions/runs/33228159433

### DIFF
--- a/pkg/acs/lib/calacs/acstable.c
+++ b/pkg/acs/lib/calacs/acstable.c
@@ -514,7 +514,7 @@ int GetAsnTable (AsnInfo *asn) {
              return (status = ELEMENT_NOT_FOUND);
          }
          /* Convert to lowercase for use as a file name */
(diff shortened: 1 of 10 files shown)


pkg/acs/lib/calacs/acstable.c
```diff
@@ -531,7 +531,7 @@ int GetAsnTable (AsnInfo *asn) {
             user to correct the value.      3-Mar-99 WJH
         */
          lowcase (exp[row].type, exp[row].mtype);
-         for (i = 0; i < strlen(exp[row].type); i++) {
+         for (i = 0; i < (int) strlen(exp[row].type); i++) {
```
pllim, line 534: The more correct way is to use size_t counter but that will cause more diff. Just easier this way.

PR #729, opened 2026-08-31
hstio:make_iodesc - Fix buffer overflow

jhunkeler:
On Ubuntu, `gcc`'s default flags include `-D_FORTIFY_SOURCE=2` and `-fstack-protector-strong`. When `hstcal` is compiled with `-O3` optimization `make_iodesc` throws a buffer overflow error.

This PR generates the `[EXTNAME,EXT]` (or `[0]`) file suffix ahead of time so that we always allocate enough space to load the path into the result.

### DIFF
--- a/hstio/hstio.c
+++ b/hstio/hstio.c
@@ -2030,9 +2030,8 @@ static void detect_iraferr(void) {
 ** constructed filename is returned.
 */
 static char *make_iodesc(IODesc **x, char *fname, char *ename, int ever) {
-        int i, n, flen;
-        char *tmp;
-        IODesc *iodesc;
+        int i, n, fname_len;
+    IODesc *iodesc;
         char xname[9];
 
         iodesc = (IODesc *)calloc(1,sizeof(IODesc));
@@ -2051,7 +2050,8 @@ static char *make_iodesc(IODesc **x, char *fname, char *ename, int ever) {
         iodesc->type = 0;
         if (fname == 0) fname = "";
         if (ename == 0) ename = "";
-        iodesc->filename = (char *)calloc(((flen = strlen(fname)) + 1), sizeof(char));
+        fname_len = strlen(fname) + 1;
+        iodesc->filename = (char *)calloc(fname_len, sizeof(char));
         if (iodesc->filename == NULL) {
             free(iodesc);
             error(NOMEM,"Allocating I/O descriptor");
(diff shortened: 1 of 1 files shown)


hstio/hstio.c
```diff
@@ -2030,9 +2030,8 @@ static void detect_iraferr(void) {
 ** constructed filename is returned.
 */
 static char *make_iodesc(IODesc **x, char *fname, char *ename, int ever) {
-        int i, n, flen;
-        char *tmp;
-        IODesc *iodesc;
+        int i, n, fname_len;
+    IODesc *iodesc;
```
pllim: Nitpick: What's up with weird indentation?

hstio/hstio.c
```diff
@@ -2077,17 +2077,27 @@ static char *make_iodesc(IODesc **x, char *fname, char *ename, int ever) {
...
-        const size_t flen_new = flen + 80;
-        tmp = (char *)calloc(flen_new,sizeof(char));
-        if (tmp == NULL) { error(NOMEM,"Allocating I/O descriptor"); return NULL; }
-        strcpy(tmp,fname);
-        if (ever == 0 || ename == 0 || ename[0] == '\0' || ename[0] == ' ')
-            strcat(tmp,"[0]");
-        else
-            snprintf(&tmp[flen], flen_new, "[%s,%d]",xname,ever);
+        const bool have_xname = strlen(xname) != 0;
+        const bool need_sep = ever == 0 || ename == 0 || ename[0] == '\0' || ename[0] == ' ';
```
pllim: Please explain human logic for this check as code comment, so we know how to modify in the future if needed.

hstio/hstio.c
```diff
@@ -2077,17 +2077,27 @@ static char *make_iodesc(IODesc **x, char *fname, char *ename, int ever) {
...
+        char fname_suffix[80] = {0};
```
pllim: Should we put this magic number "80" as a `#define` in https://github.com/jhunkeler/hstcal/blob/main/include/hstcal.h ?

jhunkeler: I don't know, because I don't believe there is a fixed length. Whoever made it 80 before me just wanted the string to fit.

`[` + `EXTNAME` + `EXTVER` + `]`

The maximum allowable length of EXTNAME is 8. EXTVER is converted from an integer to a string. So 8 plus however many characters fit INT_MAX, plus 2 for the brackets, plus 1 for the terminator.

pllim:
RT: https://github.com/spacetelescope/RegressionTests/actions/runs/33788231601

pllim (approved):
If nothing else, it is easier to read. Both CI and RT passed. Thanks!

PR #736, opened 2026-09-15
Unleash pre-commit for clang-format

pllim:
Unleash pre-commit check but use the existing clang-format (instead of copying from jwst).

Right now, it only checks files in `ctests` as a proof-of-concept. We would want to eventually apply this to all the code but we can do that slowly in batches.


pllim:
pre-commit.ci autofix

pllim:
I am going to merge to unblock #737 . Since this only going to affect two test files, we can easily adjust the rules in follow-up PR if you care, but I don't think anyone does until I unleash it into STIS.

PR #735, opened 2026-09-14
MNT: Exorcise compilation warnings in calstis

pllim:
Sequel of https://github.com/spacetelescope/hstcal/pull/734

A "small" step towards https://github.com/spacetelescope/hstcal/issues/35

RT: https://github.com/spacetelescope/RegressionTests/actions/runs/34998296477

### Blocked by

* #740
- [ ] Exorcise `(int)` from `strlen`, if any

PR #737, opened 2026-09-15
MNT: Exorcise compilation warnings in ctests

pllim:
Sequel of https://github.com/spacetelescope/hstcal/pull/735

A small step towards https://github.com/spacetelescope/hstcal/issues/35

As this only touches unit tests that run in regular CI, no need to run RT.


pllim:
Ooops I guess I should just merge #736

PR #742, opened 2026-09-17
HSTIO: "Warnings"

jhunkeler:
... but more like "everything."

The work got out of hand to say the least. 

- Reformatted the code
- Guarded if statements
- Changed the scope of a lot of index variables
- Removed redundant parentheses
- Removed unnecessary scopes, i.e. `{ int i; { // do something with i }}`
- CFITSIO expects the `long` type, so we give it `long`s 
- Fixes a several oddities related to string handling
- Optimizes `error_msg` a little by writing the message in one shot with `snprintf` instead of using `strcat`
- Align the short parameter names in functions with the long names in the prototypes.  i.e. `ever` becomes `extver`, and so on
- Sets read-only variables and parameters to `const` when possible. `typedef`s to `void *` makes this impossible in some areas of the code.
- Removes a few prototypes declared in the C code that are already exposed by `hstio.h` (why are there so many of these one-off prototypes anyway?)


pkg/acs/lib/acscte/dopcte-gen3.c
```diff
@@ -538,8 +537,7 @@ static int insertAmp(SingleGroup * image, const SingleGroup * amp, const unsigne
     unsigned rowSkipLength = image->sci.data.nx;
     unsigned offset = ctePars->columnOffset;
 
-    copyOffsetSingleGroup(image, amp, nRows, nColumns, offset, 0, rowSkipLength, nColumns);
-    return status;
+    return copyOffsetSingleGroup(image, amp, nRows, nColumns, offset, 0, rowSkipLength, nColumns);
```
pllim, line 540: Changes like this might have unintended side effects. `status` is extern int and may be changed globally. This is changing actual status return behavior.

include/hstio.h
```diff
@@ -145,23 +145,23 @@ enum StorageOrder
 
 typedef struct {
         short *buffer;          /* the pointer to the beg. of the buffer */
-        int buffer_size;        /* the size of the full 2-d array in the */
-        int tot_nx;             /* buffer.                               */
-        int tot_ny;             /*     buffer_size = tot_nx*tot_ny       */
-        int nx;                 /* The size of the current "view" of the */
-        int ny;                 /* full 2-d array.                       */
+        long buffer_size;        /* the size of the full 2-d array in the */
```
pllim, line 148: This is changing the whole struct data types. Not sure if this is gonna fly. Why is this needed?

pkg/acs/lib/acscte/dopcte-gen3.c
jhunkeler: `copyOffsetSingleGroup` returns a known `status` (`ALLOCATION_ERROR` or `ERROR_RETURN`)
`extractAmp` sets the `status` equal to a known value, or returns the status returned by copyOffsetSingleGroup. The return value of `extractAmp` sets `status` at the top-level so the `status` is still preserved up the chain

jhunkeler: Design wise its not clear why one must set `status` and return the `status` if its global in the first place.

```c
extern int status;

// definition
int func2(void) {
    if (broken) {
        return (status = ERROR_HERE);
    }
    return (status = SOMETHING_OK); // aka zero
}

// the caller does
void func1(void) {
    if ((status = func2())) {
        return status
    }
    return (status = SOMETHING_OK); // aka zero
}

int main(int argc, char *argv[]) {
    if (status = func1()) {
        check_status();
        exit(status);
    }
    return 0;
}
```

If the global variable must exist and must be used should it be simplified to something like this?

```c
extern int status;

// definition
int func2(void) {
    char *val = calloc(HUGE_ARRAY_HERE, sizeof(char));
    if (!val) {
        status = ALLOCATION_ERROR;
        return 1;
    }
    // ...
    return 0;
}

// the caller does
int func1(void) {
    if (func2()) {
        return 1;
    }
    // ...
    return 0;
}

int main(int argc, char *argv[]) {
    if (func1()) {
        goto fatal;
    }
   // ...
    fatal:
    exit(check_status()); // where check_status() reports the global status
}
```

PR #744, opened 2026-09-18
CMake: projects define their required language(s)

jhunkeler:
* HSTCAL is a C project with a splash of Fortran
* Bugfix: When C++ was not available on the system, the configuration stage would fail

In the end all this does is let someone build HSTCAL on systems where C++ might not be installed (i.e. docker)


pllim:
But who is insane enough to build hstcal on a docker without C++? 😆 

RT: https://github.com/spacetelescope/RegressionTests/actions/runs/35387931504

jhunkeler:
Me 😁 
I ran into this with stasis when I added a new library project and forgot to add `C`. The delivery workflows started failing because c++ wasn't available even though it wasn't needed.

pllim (approved):
RT is green. Let's merge so stasis can be... fluid? 😏

